PR #28, opened 2020-07-13
Allow configuration of Rails queue adapter with `:good_job`

bensheldon:
Closes #28


aried3r:
Hey! We're looking into different ActiveJob backends right now. I was wondering, is this something that could go upstream into Rails itself with the others?

https://github.com/rails/rails/tree/main/activejob/lib/active_job/queue_adapters

aried3r:
Oh, I wasn't aware. Thank you!

PR #55, opened 2020-07-24
Fix migration script in readme

thedanbob:
`add_index` within `create_table` fails because the good_jobs table doesn't exist yet


bensheldon:
@thedanbob nice catch 😓 Thank you 🙌

PR #56, opened 2020-07-24
Add migration generator

thedanbob:
Addresses #6.

Thanks for this project by the way, for background jobs I usually reach for sidekiq but I'm trying this out in an application where sidekiq would be major overkill.


bensheldon:
@thedanbob this looks great! Thank you 🙌 

I will merge this in and create a new Issue to add a test--it'll get hairy because it touches the filesystem and I don't want that to block the straightforward improvement.

thedanbob:
You're welcome! :tada:

In case it's helpful, I lifted some of this from delayed_job and maybe what they do for testing could be useful for this project too: https://github.com/collectiveidea/delayed_job_active_record/blob/master/spec/helper.rb#L55-L72

Doesn't directly apply to #57 but it would keep the schema for other tests in sync with the migration template.

PR #61, opened 2020-07-25
Update the setup documentation to use correct bin setup command

jm96441n:
## Feature
_(Short summary of changes)_
Small update to the readme so advise running `bin/setup` as opposed to `bin/setup_test`. If it's preferred we could keep the documentation the same and add the `bin/setup_test` script to handle just setting up the test env.

### Checklist
- [x] All existing tests passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ To run tests:
 $ git clone git@github.com:bensheldon/good_job.git
 
 # Set up the local environment
-$ bin/setup_test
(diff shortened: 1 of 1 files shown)


bensheldon:
Nice catch! Thank you. 

I remember I originally named it because I didn't want it to be confused with the Rails script. But looks like I changed the name at and forgot to update the docs.

PR #94, opened 2020-08-25
Add test for `rails g good_job:install`

### LINKED ISSUE #57: Add test for `rails g good_job:install`
There should be a test that the generator, at the very least, does not blow up when being run.

Deferred because it potentially involves filesystem changes. 

- Validating the stdout output of `bin/rails g good_job:install --pretend --force` looks like the right shape would be good enough.
- Otherwise, would need to potentially overwrite the existing migration file in `spec/dummy` and then restore it for a deeper inspection of the generated files.

Follow-up to #56

arku:
Fixes #57

Reference: https://github.com/foundation/foundation-rails/blob/4dfe9b12e8cf3a1aa04b257ff64c782832efc6a0/spec/features/generator_spec.rb

### DIFF
--- a/spec/lib/generators/good_job/install_generator_spec.rb
+++ b/spec/lib/generators/good_job/install_generator_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+require 'generators/good_job/install_generator'
+
+describe GoodJob::InstallGenerator, type: :generator do
+  after { remove_example_app }
+
+  it 'creates a migration for good_jobs table' do
+    setup_example_app
+
+    run_in_example_app 'rails g good_job:install'
+
+    expect(Dir.glob("#{example_app_path}/db/migrate/[0-9]*_create_good_jobs.rb")).not_to be_empty
+  end
+end
--- a/spec/support/example_app_helper.rb
+++ b/spec/support/example_app_helper.rb
@@ -1,0 +1,35 @@
+require 'fileutils'
+
+module ExampleAppHelper
+  def setup_example_app
+    root_path = example_app_path.join('..')
+
+    FileUtils.cd(root_path) do
+      `rails new #{app_name} -d postgresql --skip-action-mailer --skip-action-mailbox --skip-action-cable --skip-sprockets --skip-listen --skip-javascript --skip-turbolinks --skip-system-test --skip-test-unit --skip-bootsnap --skip-spring`
+    end
+
+    File.open("#{example_app_path}/Gemfile", 'a') do |f|
+      f.puts "gem 'good_job'"
+    end
+  end
+
+  def run_in_example_app(*args)
+    FileUtils.cd(example_app_path) do
+      system(*args) || raise("Command #{args} failed")
+    end
+  end
+
+  def remove_example_app
+    FileUtils.rm_rf(example_app_path)
+  end
+
+  def example_app_path
+    Rails.root.join('..', app_name)
+  end
+
+  def app_name
+    'example_app'
+  end
+end
+
+RSpec.configure { |c| c.include ExampleAppHelper, type: :generator }


spec/support/good_job_test_helper.rb
```diff
@@ -0,0 +1,33 @@
+require 'fileutils'
+
+module GoodJobTestHelper
```
bensheldon, line 3: This class name is a little generic. It's specifically for setting up and tearing down a full rails app. Unfortunately there is already a `test_app`, so maybe `ExampleAppHelper` ("example" in this case meaning a RSpec example).

spec/support/good_job_test_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+  def create_dummy_app
+    FileUtils.cd(tmp_path) do
+      `rails new dummy --skip-action-mailer --skip-action-mailbox --skip-action-cable --skip-sprockets --skip-listen --skip-javascript --skip-turbolinks --skip-system-test --skip-test-unit --skip-bootsnap --skip-spring`
```
bensheldon: I'd like to avoid using "dummy" because of inclusive language. `example_app` instead?

```suggestion
      `rails new example_app --skip-action-mailer --skip-action-mailbox --skip-action-cable --skip-sprockets --skip-listen --skip-javascript --skip-turbolinks --skip-system-test --skip-test-unit --skip-bootsnap --skip-spring`
```

spec/lib/generators/good_job/install_generator_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe GoodJob::InstallGenerator, type: :generator do
+  it 'creates a migration for good_jobs table' do
+    create_dummy_app
+
+    install_good_job
+
+    expect(Dir.glob("#{dummy_app_path}/db/migrate/[0-9]*_create_good_jobs.rb")).not_to be_empty
+
+    remove_dummy_app
```
bensheldon, line 12: If this is moved into an `after { remove_dummy_app }` I think it will cleanup even if the assertion fails.

spec/support/good_job_test_helper.rb
```diff
@@ -0,0 +1,33 @@
+require 'fileutils'
+
+module GoodJobTestHelper
+  def create_dummy_app
+    FileUtils.cd(tmp_path) do
+      `rails new dummy -d postgresql --skip-action-mailer --skip-action-mailbox --skip-action-cable --skip-sprockets --skip-listen --skip-javascript --skip-turbolinks --skip-system-test --skip-test-unit --skip-bootsnap --skip-spring`
```
bensheldon, line 6: I'd like to avoid using `dummy` because of inclusive language. What about `example_app`?

spec/lib/generators/good_job/install_generator_spec.rb
arku: Good point.

spec/support/example_app_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def remove_example_app
+    FileUtils.rm_rf(example_app_path)
+  end
+
+  def install_good_job
+    File.open("#{example_app_path}/Gemfile", 'a') do |f|
+      f.puts "gem 'good_job'"
+    end
```
bensheldon, line 17: I think this should be moved into the `create_example_app` because I think it will probably always be used for set up.

spec/support/example_app_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    FileUtils.cd(example_app_path) do
+      `rails g good_job:install`
+    end
```
bensheldon, line 21: I think this could be turned into a generic function like:
```suggestion
    def example_app_cmd(*args)
      FileUtils.cd(example_app_path) do
        system(*args) || abort("\n== Command #{args} failed ==")
      end
    end
```

spec/support/example_app_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def example_app_path
+    File.join(tmp_path, app_name)
+  end
+
+  def tmp_path
+    @_tmp_path ||= File.join(File.dirname(__FILE__), '..')
+  end
```
bensheldon, line 30: I don't like that this method gets mixed into the example, but I realize it's necessary to both have the root temp path and the path that has the rails app. I wonder if this method could be done away with because the only time the root temp path is needed is in the initial `rails new` command, which could be turned into `example_app_path.join('..')` I think.

spec/lib/generators/good_job/install_generator_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+require 'generators/good_job/install_generator'
+
+describe GoodJob::InstallGenerator, type: :generator do
+  after { remove_example_app }
+
+  it 'creates a migration for good_jobs table' do
+    create_example_app
+
+    install_good_job
```
bensheldon, line 10: I think this line should be expanded inside of the Example. E.g. if you follow my other suggestions:

```
example_app_cmd "rails g good_job:install"
```

spec/support/example_app_helper.rb
arku: Sure. What should we call the method as it's no longer just creating an app? 🤔

bensheldon: Good question. I don't think "setup_example_app" collides with anything.

spec/support/example_app_helper.rb
arku: We also need it to generate the `example_app_path` [here](https://github.com/bensheldon/good_job/pull/94/files#diff-cfad4f51e9f5a80e6676b6c3600f0de0R25) 🙂

spec/lib/generators/good_job/install_generator_spec.rb
arku: How about naming the method `run_in_example_app` or something better?

spec/support/example_app_helper.rb
bensheldon: I'm going to resolve this and make a broader suggestion :-) Almost there!

spec/support/example_app_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def app_name
+    'example_app'
+  end
+end
+
+RSpec.configure { |c| c.include ExampleAppHelper, type: :generator }
```
bensheldon, line 37:
```ruby
  def setup_example_app
    root_path = example_app_path.join('..')
    FileUtils.cd(root_path) do
      `rails new #{app_name} -d postgresql --skip-action-mailer --skip-action-mailbox --skip-action-cable --skip-sprockets --skip-listen --skip-javascript --skip-turbolinks --skip-system-test --skip-test-unit --skip-bootsnap --skip-spring`
    end

    File.open(example_app_path.join("Gemfile"), 'a') do |f|
      f.puts "gem 'good_job'"
    end
  end

  #...

  def example_app_path
    Rails.root.join('..', app_name)
  end

  def app_name
    'example_app'
  end
end

RSpec.configure { |c| c.include ExampleAppHelper, type: :generator }
```

spec/lib/generators/good_job/install_generator_spec.rb
bensheldon: 👍

spec/support/example_app_helper.rb
arku: I made a few minor changes but yeah, we got rid of `tmp_path` 😄

PR #97, opened 2020-08-25
Remove executor/reloader for less interlocking

### LINKED ISSUE #95: Freezes puma on code change
When running GoodJob async in a puma process, if I save a file on the filesystem and then make a new puma request then the whole process locks up irrecoverably, and I have to `kill -9` the process. I'm using Rails 6.0.3.2, puma 4.3.5 and good_job 1.2.1, in the default configuration, i.e with zeitwerk and listen, on macOS.

sj26:
I believe this will fix a bunch of dependency interlocking issues. The only thing good_job needs from the executor and/or reloader is activerecord connection management, and considering how tightly intertwined with activerecord it is, it's very reasonable to own the connection management in the same way [actioncable does](https://github.com/rails/rails/blob/master/actioncable/lib/action_cable/subscription_adapter/postgresql.rb).

ActiveJob takes care of the interface to user rails code which requires reloading, and makes sure the executor wraps user code in the right places:
https://github.com/rails/rails/blob/acaa546026471354ae1ff15e96c6bf5d39b0e34a/activejob/lib/active_job/railtie.rb#L39-L47

The only reason Sidekiq etc have to get involved with the executor is because they execute user code outside of ActiveJob.

Fixes #95

### DIFF
--- a/lib/good_job/notifier.rb
+++ b/lib/good_job/notifier.rb
@@ -70,36 +70,37 @@ module GoodJob # :nodoc:
     def listen
       future = Concurrent::Future.new(args: [@recipients, @pool, @listening], executor: @pool) do |recipients, pool, listening|
         Rails.application.reloader.wrap do
-          conn = ActiveRecord::Base.connection.raw_connection
-          ActiveSupport::Notifications.instrument("notifier_listen.good_job") do
-            conn.async_exec "LISTEN #{CHANNEL}"
-          end
-
-          begin
-            ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
-              while pool.running?
-                listening.make_true
-                conn.wait_for_notify(WAIT_INTERVAL) do |channel, _pid, payload|
-                  listening.make_false
-                  next unless channel == CHANNEL
-
-                  ActiveSupport::Notifications.instrument("notifier_notified.good_job", { payload: payload })
-                  parsed_payload = JSON.parse(payload, symbolize_names: true)
-                  recipients.each do |recipient|
-                    target, method_name = recipient.is_a?(Array) ? recipient : [recipient, :call]
-                    target.send(method_name, parsed_payload)
-                  end
-                end
-                listening.make_false
-              end
+          with_listen_connection do |conn|
+            ActiveSupport::Notifications.instrument("notifier_listen.good_job") do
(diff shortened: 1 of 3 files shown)


lib/good_job/scheduler.rb
```diff
@@ -137,11 +137,7 @@ def create_thread(state = nil)
         return false unless @performer.next?(state)
       end
 
-      future = Concurrent::Future.new(args: [@performer], executor: @pool) do |performer|
-        output = nil
-        Rails.application.executor.wrap { output = performer.next }
-        output
-      end
+      future = Concurrent::Future.new(executor: @pool, &@performer.method(:next))
```
bensheldon, line 140: This is very clean, but I think I may regret it in the future because it's easy to overlook _this is where the work gets done_.

But I'm happy to roll with it for now and just make a mental note that if it causes confusion in the future, to go to the more expansive version.

And just while I'm noodling:
- I think it would be possible to cache the proc so it doesn't have to be recreated everytime this method is called
- I wonder if the performer should alias `call` to `next` so it is directly callable (I'm imagining there was a reason why that wasn't done originally but I'm not remembering it... though it was likely that I wanted explicitness)

lib/good_job/notifier.rb
```diff
@@ -112,5 +111,16 @@ def listen
     def listen_observer(_time, _result, _thread_error)
       listen unless shutdown?
     end
+
+    def with_listen_connection
+      ar_conn = ActiveRecord::Base.connection_pool.checkout.tap do |conn|
+        ActiveRecord::Base.connection_pool.remove(conn)
```
bensheldon, line 117: Not a blocker, but I am curious: this will have the effect of using one _additional_ database connection than the pool size?

sj26: Yes, it will, good spot. It should probably be documented. But I don't think this is unreasonable — the total number of ideal connections will be one for the notifier, and then one for each worker thread.

sj26: Oh, I don't think they do — I can't find it. I'd expect it to be here:

https://edgeguides.rubyonrails.org/action_cable_overview.html#postgresql-adapter

sj26: https://github.com/rails/rails/pull/40107

sj26:
🥳

PR #99, opened 2020-08-26
Return to using executor.wrap around Scheduler execution task

bensheldon:
@sj26 I'm seeing some behavior indicating that #97 is losing advisory locks. Would you be able to take a look? 

It looks similar to what's described here: https://github.com/rails/rails/issues/26956#issuecomment-258059971

> Off the top of my head, the most potentially-surprising outcome I'd anticipate is that `executor.wrap` inside `with_connection` would cause the connection to be returned early

...at least, that's my hypothesis given that the Advisory Locks are connection-based, implying that the connection taking the advisory lock is a different connection than the one trying to release it (the `WARNING:  you don't own a lock of type ExclusiveLock`).

```bash
$ bin/rspec spec/integration/scheduler_spec.rb

Randomized with seed 533
.WARNING:  you don't own a lock of type ExclusiveLock
WARNING:  you don't own a lock of type ExclusiveLock
WARNING:  you don't own a lock of type ExclusiveLock
WARNING:  you don't own a lock of type ExclusiveLock
F

Failures:

  1) Schedule Integration when there are a large number of jobs pops items off of the queue and runs them
     Failure/Error: expect(RUN_JOBS.size).to eq number_of_jobs

       expected: 500
            got: 501

       (compared using ==)
     # ./spec/integration/scheduler_spec.rb:80:in `block (3 levels) in <top (required)>'
     # ./spec/support/reset_rails_queue_adapter.rb:4:in `block (2 levels) in <top (required)>'
     # ./spec/support/database_cleaner.rb:13:in `block (3 levels) in <top (required)>'
     # ./spec/support/database_cleaner.rb:12:in `block (2 levels) in <top (required)>'

Finished in 7.81 seconds (files took 3.96 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/integration/scheduler_spec.rb:52 # Schedule Integration when there are a large number of jobs pops items off of the queue and runs them

Randomized with seed 533

```

### DIFF
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -56,20 +56,12 @@ module GoodJob
       result = nil
       error = nil
 
-      connection_pool.with_connection do
-        unfinished.priority_ordered.only_scheduled.limit(1).with_advisory_lock do |good_jobs|
-          good_job = good_jobs.first
-          break unless good_job
+      unfinished.priority_ordered.only_scheduled.limit(1).with_advisory_lock do |good_jobs|
+        good_job = good_jobs.first
+        # TODO: Determine why some records are fetched without an advisory lock at all
+        break unless good_job&.owns_advisory_lock?
 
-          # TODO: Determine why some records are fetched without an advisory lock at all
-          if good_job.advisory_locked? && !good_job.owns_advisory_lock?
-            puts "LOCKED BUT DOES NOT OWN LOCK"
-          elsif !good_job.advisory_locked?
-            puts "TOTALLY UNLOCKED"
-          end
-
-          result, error = good_job.perform
-        end
+        result, error = good_job.perform
       end
 
       [good_job, result, error] if good_job
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -61,6 +61,13 @@ module GoodJob
           good_job = good_jobs.first
           break unless good_job
 
+          # TODO: Determine why some records are fetched without an advisory lock at all
+          if good_job.advisory_locked? && !good_job.owns_advisory_lock?
+            puts "LOCKED BUT DOES NOT OWN LOCK"
+          elsif !good_job.advisory_locked?
+            puts "TOTALLY UNLOCKED"
+          end
+
           result, error = good_job.perform
         end
       end
--- a/lib/good_job/scheduler.rb
+++ b/lib/good_job/scheduler.rb
@@ -138,7 +138,11 @@ module GoodJob # :nodoc:
         return false unless @performer.next?(state)
       end
 
-      future = Concurrent::Future.new(executor: @pool, &@performer.method(:next))
+      future = Concurrent::Future.new(args: [@performer], executor: @pool) do |performer|
+        output = nil
+        Rails.application.executor.wrap { output = performer.next }
+        output
+      end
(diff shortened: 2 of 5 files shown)


sj26:
Ah, hmm, perhaps the executor is checking-in the connection too early, now. Hmm. Perhaps we need to restore the executor around the perform block only, before it takes the advisory lock.

I'll take a look.

bensheldon:
@sj26 I found something interesting/unfortunate: 

- It seems like using `with_connection` doesn't appropriately retain the same connection leading to a situation where the advisory lock is owned by a different connection than the one that tries to unlock it.
- Even when I go back to the old `executor.wrap` there is, occasionally, a record that is returned from the query that isn't advisory locked at all. This seems to happen consistently when the scheduler is running hot (1000 jobs), so fyi there may be some intersecting issues here.

For the latter issue of the query not reliably returning an advisory locked record, I can patch an additional guard clause. It's not ideal with an extra database query, but it's _good_ until I can dig more into that.

I do feel confident that returning the `executor.wrap` addresses the former issue of the connection changing.

I'm very fascinated by all of this and appreciate your efforts 🙌

sj26:
Sorry, I ran out of time to look at this yesterday. I think what you're proposing here is correct.

ActiveRecord installs query cache handling into the executor:

https://github.com/rails/rails/blob/7101489293ece071013417aeed2da3aa1b0927c9/activerecord/lib/active_record/railtie.rb#L244

It releases connections when the executor completes, regardless of whether the connection was already held before the executor starts:

https://github.com/rails/rails/blob/7101489293ece071013417aeed2da3aa1b0927c9/activerecord/lib/active_record/query_cache.rb#L38-L46

So by the time the ActiveJob has returned we have lost the connection which took the advisory lock.

Double-wrapping the executor is harmless, the second wrap will simply yield:

https://github.com/rails/rails/blob/7101489293ece071013417aeed2da3aa1b0927c9/activesupport/lib/active_support/execution_wrapper.rb#L84

I would push the executor wrap into the perform method rather than putting it in the future for clarity about what is being wrapped a why, but this should fix the issue. :+1:

This shouldn't reintroduce the deadlock because these perform methods only run for the duration of a job. The issue was the notifier being run in an executor which never returns. As long as we don't re-wrap the notifier I think we should be good. 🙌

bensheldon:
@sj26 I appreciate all of your help. Especially digging into what's happening here and all the improvements 🎉 

> I would push the executor wrap into the perform method rather than putting it in the future for clarity about what is being wrapped a why,

I have it in the Scheduler because it felt closer to the responsibility of the thread to set up a proper execution environment, rather than the Performer or Job-level code to be aware that it was running inside of a thread. That said, I'll admit it's perhaps un-YAGNI thinking. 

I think it's becoming clear that the querying/locking implementation may need to change and that might drive my thinking and and the code further.

sj26:
Nice, I tried this out locally with my async puma setup and code reload worked without deadlocks. :+1:

gadimbaylisahil:
`break unless good_job&.owns_advisory_lock?` 

This change broke my implementation in an application with
multi-tenancy enabled with Postgres Schemas. Using https://github.com/rails-on-services/apartment gem. 

I will try to also look into why the advisory lock isn't owned. Or at least will try to create an example application with same setup.

Issue #103, opened 2020-08-27
requiring more dependencies in then needed


thilo:
good_job currently requires rails which pulls in more frameworks than needed.

bensheldon:
@thilo thank you for making the PR 😎 I just merged it and will cut a patch release momentarily.

 I’m curious about your application context; is it Rails API? That can help me be aware of what kinds of minimal Rails Apps folks are running and be mindful of dependencies in the future.

PR #104, opened 2020-08-27
stop depending on all rails libs

thilo:
only railties, activerecord, activejob
fixing [#103]


bensheldon:
Thank you! This is very helpful! 🎉

PR #112, opened 2020-08-28
Add first cut at `Performer` docs

Mr0grog:
I took a few minutes to go through `Performer` and attempt to document it after we finished talking this afternoon. Feel free to use or toss!

**Note: this is targeted at the `code_docs` branch for #111.**

### DIFF
--- a/lib/good_job/performer.rb
+++ b/lib/good_job/performer.rb
@@ -1,7 +1,29 @@
 module GoodJob
+  #
+  # Performer queries the database for jobs and performs them on behalf of a
+  # {Scheduler}. It mainly functions as glue between a {Scheduler} and the jobs
+  # it should be executing.
+  #
   class Performer
+    # @!attribute [r] name
+    # @return [String]
+    #   a meaningful name to identify the performer in logs and for debugging.
+    #   This is usually set to the list of queues the performer will query,
+    #   e.g. +"-transactional_messages,batch_processing"+.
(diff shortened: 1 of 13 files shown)


lib/good_job/performer.rb
```diff
@@ -1,18 +1,46 @@
 module GoodJob
+  #
+  # Performer queries the database for jobs and performs them on behalf of a
```
bensheldon, line 3: This is the applied purpose for the Perfomer. Though the performer is intended to be generic/ignorant of the nature of the task. 

The practical reason for the performer is to enforce a callable that does not rely on scoped/closures variables because they might not be available when executed in a different thread. Or alternatively, it describes an interface that is compatible with the Scheduler.

lib/good_job/performer.rb
```diff
@@ -1,18 +1,46 @@
...
       @name = name
       @filter = filter
     end
 
+    # Find and perform any eligible jobs.
     def next
       @target.public_send(@method_name)
     end
 
+    # Tests whether this performer should be used in GoodJob's current state by
```
bensheldon, line 39: "State" is the generic. Practically, the state is a LISTEN/NOTIFY message that is passed down from the Notifier to the Scheduler, and the Scheduler then is able to ask the Performer "does this message relate to you?" and allow the Performer to say "no, don't bother waking up thread".

Issue #110, opened 2020-08-28
Add environment variable to mirror `cleanup_preserved_jobs --before-seconds-ago=SECONDS`


bensheldon:
The CLI should have an environment variable option for all parameters. `cleanup_preserved_jobs --before-seconds-ago=SECONDS` is missing an environment variable option.

Add `GOOD_JOB_CLEANUP_PRESERVED_JOBS_BEFORE_SECONDS_AGO` as an environment variable that defers to the parameter.

gadimbaylisahil:
I have finished with the PR including documentation change, please review :pray:

PR #114, opened 2020-08-31
Allow env variable config for cleanups

gadimbaylisahil:
Refers to https://github.com/bensheldon/good_job/issues/110

I was thinking to add tests to CLI specs as well but think configuration is sufficient.

### DIFF
--- a/lib/good_job/configuration.rb
+++ b/lib/good_job/configuration.rb
@@ -51,5 +51,13 @@ module GoodJob
         1
       ).to_i
     end
+
+    def cleanup_preserved_jobs_before_seconds_ago
+      (
+        options[:before_seconds_ago] ||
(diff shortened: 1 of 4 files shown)


lib/good_job/configuration.rb
```diff
@@ -51,5 +51,13 @@ def poll_interval
         1
       ).to_i
     end
+
+    def before_seconds_ago
```
bensheldon, line 55: I think this method should be named "cleanup_preserved_jobs_before_seconds_ago". It's a little unsymmetrical because the command line option isn't the whole phrase, but the longer phrase seems like the appropriate message to ask of the configuration object.

gadimbaylisahil: Yes, that came to my mind at first, but decided not to do as I wasn't going to change the public configuration option GoodJob asks from user.

Changes now

Issue #117, opened 2020-08-31
Jobs are not processed in multi schema setup. Apartment + GoodJob ( post 1.1.2 )


gadimbaylisahil:
Hi, I have an application with multi tenant architecture using PostgreSQL Schemas with the help of [Apartment gem](https://github.com/rails-on-services/apartment). 

Prior to 1.1.2(without Listen/Notify), Good Job processed this setup just fine. 

After 1.1.2, queried jobs does not own advisory_lock. job.owns_advisory_lock returns `false`. 

I have limited knowledge in this area but tried to debug it for a few days without luck. However, I see that there is TODO
comment in Job.rb which is interesting. 

```
# TODO: Determine why some records are fetched without an advisory lock at all
```

Can you please provide info on what is happening here or are you still in dark clouds as well regarding this?

I have created an example application [here](https://github.com/gadimbaylisahil/apartment-with-good-job-test) with exact same setup.

Any direction is appreciated, I can debug more :+1:

bensheldon:
@gadimbaylisahil Yikes! 😰 Thanks for flagging this! I've been curious about multi-database setups. And thank you for all your contributions too 🙏 

To point you in the directions I'm thinking:

- The underlying SQL query that does the advisory locking has a problem. I added the guard clause you saw because I was seeing occasionally, when under heavy load, jobs would be fetched by the query that were not locked. It was fairly rare which is why I felt ok adding a guard clause and flagging it for later: https://github.com/bensheldon/good_job/blob/fe2d5cbf81ccc665e8a3df012b803cd6ba04f344/lib/good_job/job.rb#L59-L62
- Database connections are being lost/swapped/leaked. I spent a lot of time with that in #99
- GoodJob doesn't currently work with the Apartment Gem and the guard clause has made that obvious. 

I'll dig into this too. Thank you for the example application.

bensheldon:
Just to keep you in the loop with my debugging. I'm seeing a SQL error when running this:

```bash
irb(main):008:0> GoodJob::Job.all.advisory_locked
    GoodJob::Job Load (2.1ms) SELECT "public"."good_jobs".* FROM "public"."good_jobs" LEFT JOIN pg_locks ON pg_locks.locktype = 'advisory'
  AND pg_locks.objsubid = 1
  AND pg_locks.classid = ('x'||substr(md5('public.good_jobs' || "public.good_jobs"."id"::text), 1, 16))::bit(32)::int
[...]

PR #119, opened 2020-09-01
Better table name detection for Job queries

gadimbaylisahil:
https://github.com/bensheldon/good_job/issues/117

### DIFF
--- a/lib/good_job/lockable.rb
+++ b/lib/good_job/lockable.rb
@@ -25,8 +25,8 @@ module GoodJob
         join_sql = <<~SQL
           LEFT JOIN pg_locks ON pg_locks.locktype = 'advisory'
             AND pg_locks.objsubid = 1
-            AND pg_locks.classid = ('x'||substr(md5(:table_name || "#{table_name}"."#{primary_key}"::text), 1, 16))::bit(32)::int
-            AND pg_locks.objid = (('x'||substr(md5(:table_name || "#{table_name}"."#{primary_key}"::text), 1, 16))::bit(64) << 32)::bit(32)::int
+            AND pg_locks.classid = ('x'||substr(md5(:table_name || #{quoted_table_name}.#{quoted_primary_key}::text), 1, 16))::bit(32)::int
+            AND pg_locks.objid = (('x'||substr(md5(:table_name || #{quoted_table_name}.#{quoted_primary_key}::text), 1, 16))::bit(64) << 32)::bit(32)::int
         SQL
 
         joins(sanitize_sql_for_conditions([join_sql, { table_name: table_name }]))


gadimbaylisahil:
@bensheldon There are two commits here. First one just removes the .public from the table name and is a hacky way to display the working version. 

The second commit tries to include the table_name in 

```ruby
      scope :joins_advisory_locks, (lambda do
        join_sql = <<~SQL
          LEFT JOIN pg_locks ON pg_locks.locktype = 'advisory'
            AND pg_locks.objsubid = 1
            AND pg_locks.classid = ('x'||substr(md5(:table_name || :table_with_primary_key_quoted::text), 1, 16))::bit(32)::int
            AND pg_locks.objid = (('x'||substr(md5(:table_name || :table_with_primary_key_quoted::text), 1, 16))::bit(64) << 32)::bit(32)::int
        SQL

        table_with_primary_key_quoted = [quoted_table_name, quoted_primary_key].join('.')

        joins(sanitize_sql([join_sql, { table_name: quoted_table_name, table_with_primary_key_quoted: table_with_primary_key_quoted }]))
      end)
```

with `table_with_primary_key_quoted = [quoted_table_name, quoted_primary_key].join('.')`. However this produces different SQL string. 

Prior does as "good_jobs"."id" but the latter is '"public"."good_jobs"."id"'::text. This results in empty collection. 

If I were to put the "public"."good_jobs"."id" manually without the single quotes that sanitization produces, all works fine. 

I assume this is due to md5 generation which includes single quotes in the latter version which then doesn't matchup with other quries/locks(not sure).

lib/good_job/lockable.rb
```diff
@@ -13,7 +13,7 @@ module Lockable
 
         query = cte_table.project(cte_table[:id])
                   .with(composed_cte)
-                  .where(Arel.sql(sanitize_sql_for_conditions(["pg_try_advisory_lock(('x'||substr(md5(:table_name || \"#{cte_table.name}\".\"#{primary_key}\"::text), 1, 16))::bit(64)::bigint)", { table_name: table_name }])))
+                  .where(Arel.sql(sanitize_sql_for_conditions(["pg_try_advisory_lock(('x'||substr(md5(:table_name || \"#{cte_table.name}\".\"#{primary_key}\"::text), 1, 16))::bit(64)::bigint)", { table_name: quoted_table_name }])))
```
bensheldon, line 16: I think you have the logic reversed on these. 

The parameterized value does not need to be quoted because it is intended to be inserted into the query as a parameterized string. 

The interpolated table/columns are the ones that need to be quoted. (Sorry, I'm on my phone so I can't write out code easily). 

Eg

```ruby
.where(Arel.sql(sanitize_sql_for_conditions(["pg_try_advisory_lock(('x'||substr(md5(:table_name || #{quoted_table_name(cte_table.name)}.#{quoted_column_name(primary_key)}::text), 1, 16))::bit(64)::bigint)", { table_name: table_name }])))
```

gadimbaylisahil: I reverted this change as it was unrelated. But than you for mentioning the quotation thing. I removed extra quotations now. The only change I made is in joins_advisory_lock scope which adds quoted_table_name and quoted_primary_key

gadimbaylisahil:
Closes #117

bensheldon:
@gadimbaylisahil I just released GoodJob v1.2.4 with this change.

Issue #87, opened 2020-08-21
Use Zeitwerk for auto-loading


bensheldon:
Example: https://thurlow.io/ruby/2019/11/17/add-code-auto-loading-to-a-ruby-project-with-zeitwerk.html

gadimbaylisahil:
@bensheldon Ready to review

bensheldon:
@gadimbaylisahil thank you! 🎉

PR #123, opened 2020-09-01
Zeitwerk Loader Implementation

gadimbaylisahil:
https://github.com/bensheldon/good_job/issues/87

### DIFF
--- a/good_job.gemspec
+++ b/good_job.gemspec
@@ -43,6 +43,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pg", ">= 1.0.0"
   spec.add_dependency "railties", ">= 5.1.0"
   spec.add_dependency "thor", ">= 0.14.1"
+  spec.add_dependency "zeitwerk", ">= 2.0"
 
   spec.add_development_dependency "appraisal"
   spec.add_development_dependency "database_cleaner"
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -1,18 +1,18 @@
 require "rails"
-require 'good_job/railtie'
 
-require 'good_job/configuration'
-require 'good_job/log_subscriber'
-require 'good_job/lockable'
-require 'good_job/job'
-require 'good_job/scheduler'
-require 'good_job/multi_scheduler'
-require 'good_job/adapter'
-require 'good_job/performer'
-require 'good_job/current_execution'
-require 'good_job/notifier'
+require "active_job"
+require "active_job/queue_adapters"
 
-require 'active_job/queue_adapters/good_job_adapter'
+require "zeitwerk"
+
+loader = Zeitwerk::Loader.for_gem
+loader.inflector.inflect(
+  'cli' => "CLI"
+)
+loader.push_dir(File.join(__dir__, ["generators"]))
+loader.setup
+
+require "good_job/railtie"
 
 # GoodJob is a multithreaded, Postgres-based, ActiveJob backend for Ruby on Rails.
 #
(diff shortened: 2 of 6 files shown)


gadimbaylisahil:
Logger currently has issues. Not sure why yet, will investigate later. 

GoodJob.logger doesn't change, due to initializer not being triggered I suppose.

spec/lib/good_job/cli_spec.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 require 'rails_helper'
-require 'good_job/cli'
```
gadimbaylisahil, line 3: Not needed

good_job.gemspec
```diff
@@ -43,6 +43,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pg", ">= 1.0.0"
   spec.add_dependency "railties", ">= 5.1.0"
   spec.add_dependency "thor", ">= 0.14.1"
+  spec.add_dependency "zeitwerk", ">= 2.2.2"
```
gadimbaylisahil, line 46: Not sure about the version range to depend on

lib/good_job.rb
```diff
@@ -1,19 +1,17 @@
...
-require 'good_job/lockable'
-require 'good_job/job'
-require 'good_job/scheduler'
-require 'good_job/multi_scheduler'
-require 'good_job/adapter'
-require 'good_job/performer'
-require 'good_job/current_execution'
-require 'good_job/notifier'
+require "active_job"
+require "active_job/queue_adapters"
```
gadimbaylisahil, line 4: https://github.com/fxn/zeitwerk#reopening-third-party-namespaces

lib/good_job.rb
```diff
@@ -1,19 +1,17 @@
...
+require "good_job/railtie"
```
gadimbaylisahil, line 5: When not requiring the Railtie the initializers were not triggered in railtie.rb file due to loading orders. 

To verify this assumption I did two things: 

1. Use `loader.eager_load` after `loader.setup` which worked.
2. Enable eager loading in test.rb file of spec/test_app which worked

Thus moved the railtie to good_job top requires. 

I usually require railtie on top of my main gem entry class as well. Opinions appreciated

lib/good_job.rb
```diff
@@ -1,19 +1,17 @@
...
 
-require 'active_job/queue_adapters/good_job_adapter'
+require "zeitwerk"
 
+loader = Zeitwerk::Loader.for_gem
+loader.inflector.inflect(
+  'cli' => "CLI"
+)
+loader.push_dir(File.join(__dir__, ["generators"]))
```
gadimbaylisahil, line 13: Generator is not a namespace thus adding it to main directories

gadimbaylisahil:
> Logger currently has issues. Not sure why yet, will investigate later.
> 
> GoodJob.logger doesn't change, due to initializer not being triggered I suppose.

Found the issue. See PR code comments

lib/good_job.rb
bensheldon: I ran across this before in Zeitwerk being troublesome and did some [Github searching](https://github.com/search?l=&q=zeitwerk+%22railtie%22+path%3Alib&type=Code).

I think this should happen at the bottom of the file _after_ Zeitwerk is set up and _after_ the root namespace is created. Here's an example: https://github.com/dukex/aggregated_record/blob/e8a597aaee94eb6e03ba0a26654a9c7f50a9c321/lib/aggregated_record.rb#L4-L13

good_job.gemspec
gadimbaylisahil: changed

lib/good_job.rb
gadimbaylisahil: Changed

Issue #50, opened 2020-07-23
Create Web UI Dashboard


strzibny:
Is UI similar to Sidekiq UI available or planned?

bensheldon:
Thanks for asking!

I've put a little thought into it, but nothing firmly held:

- A UI would be fun. Sidekiq's chart is very satisfying. 
- A UI would be fairly easy to build given GoodJob is Rails and Postgres. 
- 👍 : economical person's Kibana wallboard. Visually see status of latency and queue depth. Quick feedback that things are working.
- 👎 : administrative tool for deleting or re-enqueing or deeply inspecting the status or errors of jobs. 

I don't have a lot of experience building and securing Rails Engines or the like, but would welcome contributions that are inline with the thoughts above. And also welcome to start small, eg a table of worked jobs. Or a simple chart. Release it and iterate.

inopinatus:
There are differential user stories here according to scale. Many users will be satisfied with a minimal interface, for which delayed_job_web is my own reference point, but anyone coming from a large Sidekiq install will bring more sophisticated preferences.

gadimbaylisahil:
I have started testing out GoodJob for one of our production applications, if all goes well with testing, I would be glad to take on this project.

bensheldon:
Thanks @gadimbaylisahil! I hope it goes well 😓

bensheldon:
This feature is next up in the backlog. I wanted to start flushing it out. Iterative development is great, not all of these features are not expected to land together.

- Table of enqueued jobs:
  Note: GoodJob creates a new job record when a job errors and is retried. e.g. there will exist multiple GoodJob::Job IDs/records for the same ActiveJob ID. Some kind of grouping would be nice to have, but not necessary initially.

  Columns:
  - Scheduled at
  - Queue
  - Job Class
  - Arguments

- Charts.
  - Scheduled jobs by time slot
  - Executed jobs by time slot (split by whether they finished succesfully or errored)
  - Queue latency, the delta between a job is scheduled to run and when it actually runs. Example SQL:
    ```
    SELECT
        queue_name,
        date_trunc('hour', CAST(performed_at AS timestamp)) AS performed_at,
        EXTRACT(epoch FROM avg(latency)) AS avg
    FROM (
             SELECT
                 (performed_at - scheduled_at) AS latency,
                 finished_at AS finished_at,
                 performed_at AS performed_at,
                 queue_name AS queue_name
             FROM good_jobs
         ) source
[...]

PR #131, opened 2020-09-07
Spike on data dashboard; pull in full Bootstrap CSS and JS

bensheldon:
Connects to #50 

Screen Shot 2020-09-07 at 2 15 48 PM


bensheldon:
- [ ] Add `engine` directory to Gemspec

morgoth:
@bensheldon I wanted to try the engine but it doesn't load for me.
I'm using the gem from master branch, but I cannot require "good_job/engine".

I can also see that it is not in load path as stated in https://github.com/bensheldon/good_job/blob/main/good_job.gemspec#L2

```
> pp $LOAD_PATH.grep /good_job/
=> ["/home/wojtek/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0/bundler/gems/good_job-c84288185da6/lib"]
```

Do I need to do some extra setup to make it work?

bensheldon:
@morgoth thank you for taking the time to try to get it working. I am investigating.

bensheldon:
@morgoth I have a PR (#134) that I think fixes the problem. Could you try it out locally and give me feedback? Here's some preliminary documentation:

1. Explicitly require the Engine code at the top of your `config/application.rb` file, immediately after Rails is required. This is necessary because the mountable engine is an optional feature of GoodJob.

    ```ruby
    # config/application.rb
    require_relative 'boot'

    require 'rails/all'
    require 'good_job/engine' # <= Add this line
    # ...
    ```

2. Mount the engine in your `config/routes.rb file. The following will mount it at the `example.com/good_job` URL.

    ```ruby
    # config/routes.rb
    # ...
    mount GoodJob::Engine => 'good_job'
    ```

    Because jobs can potentially contain sensitive information, you should have authorization. For example, using Devise's `authenticate` helper, that might look like:

    ```ruby
    # config/routes.rb
    # ...
    authenticate :user, ->(user) { user.admin? } do
      mount GoodJob::Engine => 'good_job'
    end
    ``` 

Thank you 🙏

Issue #136, opened 2020-09-14
Preserve only failed jobs


morgoth:
Looks like currently there is no configuration option to preserve only failed jobs and delete successfully finished ones, as `preserve_job_records` will preserve all.

I think it would be very helpful to have such functionality out of the box (or by config) to be possible to easily investigate failed job errors.

Is this easy to do on my own, with currently available GoodJob API?

bensheldon:
@morgoth I think this is a great feature request and could see it implemented in GoodJob.

I looked at how easy it would be to implement outside of GoodJob; GoodJob isn't currently instrumented to make it easy :-( So there are two options:

1. Monkeypatch `GoodJob::Job#perform` by wrapping the original method and then deleting the job if the job is finished and there isn't an error.
2. Write your own rake task that deletes finished/unerrored jobs and delete them regularly. `GoodJob::Job.where(errored_at: nil).where.not(finished_at: nil).delete_all`

morgoth:
@bensheldon How would you see it implemented?

I was thinking about changing config values for `GoodJob.preserve_job_records` to be something like `:never, :always, :on_error`
and then changing the `perform` method signature to something like `def perform(preserve_record: GoodJob.preserve_job_records, reperform_on_standard_error: GoodJob.reperform_jobs_on_standard_error)`

Then I guess we would also need to set `finished_at` on such errored record so it is not retried again.

Do you have some different idea on your mind?

6temes:
Sidekiq has an interesting way to manage the failed jobs:

![image](https://user-images.githubusercontent.com/5655861/93334358-c80b8080-f85f-11ea-9e6c-8877709bc4d9.png)

"Failed jobs" means the number of times a job has failed, but maybe it eventually ran successfully after several retries.

"Dead jobs" is a queue where the jobs that have been retried failed more than N number of times go.

Usually, when you have a problem with a job, you catch it in the retries queue because they usually retry during a couple of times, but when Sidekiq gives up, you can find them in "Dead jobs."

I think that this is a really useful pattern.

So, my point is that, keeping jobs that have ran successfully can be optional, but dead jobs should never be removed.

bensheldon:
@morgoth Yes! That's very close to my initial thoughts. Some suggested adjustments:

- The Boolean signature for `GoodJob.preserve_job_records=` needs to be preserved for compatibility. I think the smallest change would be to allow `nil`, `False`, `True`, and the new one `:on_error`.
[...]

Issue #72, opened 2020-08-11
Job exceptions not properly attached to good_jobs record 


bensheldon:
When there is an exception that is rescued by `retry_on` or `discard_on`, it is not consistently available to the GoodJob adapter as an argument, and does not always appear in the `good_jobs.error` column. This can happen if there are additional `*_perform` hooks that intercept the return value.

It seems like the only reliable way to access [`retry_on`](https://github.com/rails/rails/blob/23a9e29d9b1882e4d28530fca98b8c57b3de0410/activejob/lib/active_job/exceptions.rb#L122-L126) or `discard_on` (without monkey-patching them) is to use ActiveSupport::Instrumentation to hook into the events: https://guides.rubyonrails.org/active_support_instrumentation.html#retry-stopped-active-job

Unfortunately, this means having to pass those objects up through a global (like [ActiveSupport::Current](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html)) and retrieved in the job processor.

PR #138, opened 2020-09-16
Extract "execute" method to reduce "perform" method complexity

morgoth:
While looking at a way to implement https://github.com/bensheldon/good_job/issues/136 I found perform method quite complex and was trying to simplify it a bit.

Feel free to close it if you prefer to have it in one method.

### DIFF
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -193,28 +193,16 @@ module GoodJob
       raise PreviouslyPerformedError, 'Cannot perform a job that has already been performed' if finished_at
 
       GoodJob::CurrentExecution.reset
-      result = nil
-      rescued_error = nil
-      error = nil
 
       self.performed_at = Time.current
       save! unless destroy_after
 
-      params = serialized_params.merge(
-        "provider_job_id" => id
-      )
-
-      begin
-        ActiveSupport::Notifications.instrument("perform_job.good_job", { good_job: self, process_id: GoodJob::CurrentExecution.process_id, thread_name: GoodJob::CurrentExecution.thread_name }) do
-          result = ActiveJob::Base.execute(params)
-        end
-      rescue StandardError => e
-        rescued_error = e
-      end
+      result, rescued_error = execute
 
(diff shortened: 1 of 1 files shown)


lib/good_job/job.rb
```diff
@@ -193,28 +193,16 @@ def perform(destroy_after: !GoodJob.preserve_job_records, reperform_on_standard_
...
-      end
+      result, rescued_error = execute
 
       retry_or_discard_error = GoodJob::CurrentExecution.error_on_retry ||
                                GoodJob::CurrentExecution.error_on_discard
 
+      error = nil
       if rescued_error
         error = rescued_error
       elsif result.is_a?(Exception)
```
morgoth, line 208: Do you know why/when the result can be an exception? I find it quite surprising.

bensheldon: Good question! Apologies if this turns into story time.

Briefly: when using `ActiveJob.rescue_from` (or `retry_on` / `discard_on` which are implemented with `rescue_from`), ActiveJob will return the exception that is caught/handled.

Storytime...

I'm not sure whether the above behavior is intentional or merely a coincidence. 

The behavior is unreliable though. I discovered in production that Sentry's instrumentation interrupts that behavior: https://github.com/getsentry/raven-ruby/blob/a5f101a9c7776168df1c0e2aac13d46b70bbb24e/lib/raven/integrations/rails/active_job.rb#L21-L31

...you'll notice that Sentry will call `rescue_with_handler` but does _not_ return the result. I just checked it off my todo list by submitting a PR (thanks for the inspiration 🎉 ): https://github.com/getsentry/raven-ruby/pull/1027 ...but I have to assume that other gems may also interrupt this behavior too. This is explored in #72.

bensheldon (approved):
I think this is a great refactor. 

A weakly held stylistic thought is that `result, rescued_error = execute` does visually understate _THIS IS WHERE THE WORK HAPPENS_ and I could see it being overlooked by someone visiting this file with fresh eyes. 

...but that's not really actionable feedback. I will merge this and keep it in mind 🙌

PR #140, opened 2020-09-17
Remove arguments from perform method

morgoth:
Follow up to https://github.com/bensheldon/good_job/issues/136#issuecomment-693451491

### DIFF
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -177,25 +177,17 @@ module GoodJob
     end
 
     # Execute the ActiveJob job this {Job} represents.
-    # @param destroy_after [Boolean]
-    #   Whether to destroy the {Job} record after executing it if the job did
-    #   not need to be reperformed. Defaults to the value of
-    #   {GoodJob.preserve_job_records}.
-    # @param reperform_on_standard_error [Boolean]
-    #   Whether to re-queue the job to execute again if it raised an instance
-    #   of +StandardError+. Defaults to the value of
-    #   {GoodJob.reperform_jobs_on_standard_error}.
     # @return [Array<(Object, Exception)>]
     #   An array of the return value of the job's +#perform+ method and the
     #   exception raised by the job, if any. If the job completed successfully,
     #   the second array entry (the exception) will be +nil+ and vice versa.
-    def perform(destroy_after: !GoodJob.preserve_job_records, reperform_on_standard_error: GoodJob.reperform_jobs_on_standard_error)
+    def perform
       raise PreviouslyPerformedError, 'Cannot perform a job that has already been performed' if finished_at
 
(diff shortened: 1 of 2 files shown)


morgoth:
The linter fails because code uses Ruby 2.5 syntax and Rubocop is set to lint against 2.4 (the same as in gemspec https://github.com/bensheldon/good_job/blob/main/good_job.gemspec#L48) but there are no tests for Ruby 2.4.
Should I change the code to 2.4 syntax or do you think it is fine to bump gem to use 2.5?

bensheldon:
@morgoth Oof, sorry you got bitten by that inconsistency. Let's drop support for Ruby 2.4; it looks like [Ruby 2.4 EOLed on March 31, 2020](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/). 

Could you please remove the references to 2.4, and change the Linter config to 2.5 as its own PR?

spec/lib/good_job/job_spec.rb
```diff
@@ -204,11 +204,16 @@ def perform(result_value = nil, raise_error: false)
     end
 
     it 'can preserve the job' do
-      good_job.perform(destroy_after: false)
+      current_preserve_job_records = GoodJob.preserve_job_records
+      GoodJob.preserve_job_records = true
```
bensheldon, line 208: The RSpec-y way to do this is with a [message stub](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/configuring-responses) which will be automatically cleaned up and reset at the end of the test example:  `allow(GoodJob).to receive(:preserve_job_records).and_return(true)`

lib/good_job/job.rb
```diff
@@ -215,15 +207,15 @@ def perform(destroy_after: !GoodJob.preserve_job_records, reperform_on_standard_
...
       else
         self.finished_at = Time.current
 
-        if destroy_after
-          destroy!
-        else
+        if GoodJob.preserve_job_records
           save!
+        else
+          destroy!
```
bensheldon, line 218: This reads so much better now with the positive condition coming first 🎉

bensheldon (commented):
This is fantastic! 🙌 

I left some feedback on the tests, specifically to use RSpec message stubs instead of manually setting and resetting global values. I have to apologize (and express my gratitude to you for working in them) because those tests could be more exemplary.

Once you fix the linter issue, and make the changes to the tests, this is ready to go 👍

bensheldon:
@morgoth I just merged #142 to unblock this.

morgoth:
@bensheldon updated and tests are green now. Thanks

bensheldon:
@morgoth Thank _you_! 🙌

PR #141, opened 2020-09-17
Prevent jobs hanging

morgoth:
Vide https://github.com/bensheldon/good_job/pull/140/checks?check_run_id=1127285410

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes


bensheldon (approved):
Yikes! Nice improvement. 👍

PR #142, opened 2020-09-17
Drop Ruby 2.4 support

morgoth:
+ autocorrected syntax by Rubocop


bensheldon (approved):
There it is: https://blog.bigbinary.com/2017/10/24/ruby-2.5-allows-rescue-inside-do-end-blocks.html

Thank you!

PR #145, opened 2020-09-18
Add `GoodJob.preserve_job_records = :on_unhandled_error` option to only preserve jobs that errored

### LINKED ISSUE #136: Preserve only failed jobs

morgoth:
Adds option :on_error to preserve jobs in database only on final error and delete otherwise.

This is done in 2 commits. First one just introduces the functionality, the second one changes options to :never, :always, :on_error with backward compatibility.

closes #136

### DIFF
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -209,10 +209,12 @@ module GoodJob
 
       if rescued_error && GoodJob.reperform_jobs_on_standard_error
         save!
+      elsif rescued_error && GoodJob.preserve_job_records == :on_error
+        update!(finished_at: Time.current)
       else
         self.finished_at = Time.current
 
-        if GoodJob.preserve_job_records
+        if GoodJob.preserve_job_records == true
           save!
         else
           destroy!
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -191,37 +191,25 @@ module GoodJob
 
       result, rescued_error = execute
 
-      retry_or_discard_error = GoodJob::CurrentExecution.error_on_retry ||
-                               GoodJob::CurrentExecution.error_on_discard
+      result_error = nil
+      result, result_error = nil, result if result.is_a?(Exception) # rubocop:disable Style/ParallelAssignment
 
-      error = nil
-      if rescued_error
-        error = rescued_error
-      elsif result.is_a?(Exception)
-        error = result
-        result = nil
-      elsif retry_or_discard_error
-        error = retry_or_discard_error
-      end
+      job_error = rescued_error ||
+                  result_error ||
+                  GoodJob::CurrentExecution.error_on_retry ||
+                  GoodJob::CurrentExecution.error_on_discard
 
-      error_message = "#{error.class}: #{error.message}" if error
-      self.error = error_message
+      self.error = "#{job_error.class}: #{job_error.message}" if job_error
 
       if rescued_error && GoodJob.reperform_jobs_on_standard_error
         save!
-      elsif rescued_error && GoodJob.preserve_job_records == :on_error
+      elsif GoodJob.preserve_job_records == true || (GoodJob.preserve_job_records == :on_error && rescued_error)
         update!(finished_at: Time.current)
       else
-        self.finished_at = Time.current
-
-        if GoodJob.preserve_job_records == true
-          save!
-        else
-          destroy!
-        end
+        destroy!
       end
 
-      [result, error]
+      [result, job_error]
     end
 
     private
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -189,21 +189,21 @@ module GoodJob
       self.performed_at = Time.current
       save! if GoodJob.preserve_job_records
 
-      result, rescued_error = execute
+      result, unhandled_error = execute
 
       result_error = nil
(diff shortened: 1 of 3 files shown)


morgoth:
I also have a bit of refactoring ready on https://github.com/tiramizoo/good_job/compare/preserve-on-error...tiramizoo:refactoring?expand=1
but don't want to put too much into this PR

bensheldon:
@morgoth Thank you for tackling this! I'm really excited to add the functionality. 

I'd like to retain (not deprecate) the Boolean values for preserve_job_records, and only add the one `on_error` symbol right now. I'm not ready to commit fully to symbolic values right now.

morgoth:
ok, I removed the second commit

lib/good_job/job.rb
```diff
@@ -209,10 +209,12 @@ def perform
 
       if rescued_error && GoodJob.reperform_jobs_on_standard_error
         save!
+      elsif rescued_error && GoodJob.preserve_job_records == :on_error
+        update!(finished_at: Time.current)
```
bensheldon, line 213: I think this needs to be testing against `self.error`(any error that occurred during the job), rather than just `rescued_error` (errors that bubble up to the GoodJob backend).... or I think we should rename it specifically to be "on_backend_error"/"on_unhandled_error" (naming is hard).

Assuming that `self.error` matches with your needs for this, I think the entire block might be clearer with further rewriting (apologies it was messy in the first place). Something like:

```ruby
if rescued_error && GoodJob.reperform_jobs_on_standard_error
  save!
elsif GoodJob.preserve_job_records == true || (error && GoodJob.preserve_job_records == :on_error)
  self.finished_at = Time.current
  save!
else
  destroy!
end
```

bensheldon (changes requested):
@morgoth I have some feedback, but I think we're close. 

BTW, after writing out my code suggestion in this review, I went back to your refactoring branch and my suggestion looks nearly identical to your branch :-)

bensheldon:
@morgoth re:refactoring. you drew my attention to the unfortunate fact that the local variable `error` is shadowing `GoodJob::Job#error` 😓

lib/good_job/job.rb
morgoth: I actually wanted saving only on "rescued" aka "unhandled" error, so I renamed the config to be `:on_unhandled_error`, as otherwise it would create a bit of noise to have multiple records of  the same job/error.
It would be not clear if those entries are coming from the same AJ job or different ones.
This is matching the behavior of other AJ queue adapters.

I think storing on every error could also be an option (maybe another config value?) with some nice web UI grouping, so it's clear that given group of job exception is coming from the same AJ job.

morgoth (commented):
I applied my refactoring branch already. Hope it is more clear now.

lib/good_job/job.rb
bensheldon: That is great. That makes me think (as a separate PR) that we should rename the configuration `GoodJob. reperform_jobs_on_standard_error` as `GoodJob.reperform_jobs_on_unhandled_error`. That configuration name was always unwieldy.

lib/good_job/job.rb
```diff
@@ -189,37 +189,27 @@ def perform
       self.performed_at = Time.current
       save! if GoodJob.preserve_job_records
 
-      result, rescued_error = execute
+      result, unhandled_error = execute
 
-      retry_or_discard_error = GoodJob::CurrentExecution.error_on_retry ||
-                               GoodJob::CurrentExecution.error_on_discard
+      result_error = nil
+      result, result_error = nil, result if result.is_a?(Exception) # rubocop:disable Style/ParallelAssignment
```
bensheldon, line 195: Style: I think the stylecop should be preserved and this done as 2 lines.

lib/good_job/job.rb
```diff
@@ -189,37 +189,27 @@ def perform
...
 
-      error_message = "#{error.class}: #{error.message}" if error
-      self.error = error_message
+      self.error = "#{job_error.class}: #{job_error.message}" if job_error
 
-      if rescued_error && GoodJob.reperform_jobs_on_standard_error
+      if unhandled_error && GoodJob.reperform_jobs_on_standard_error
         save!
+      elsif GoodJob.preserve_job_records == true || (GoodJob.preserve_job_records == :on_unhandled_error && unhandled_error)
+        update!(finished_at: Time.current)
```
bensheldon, line 207: Style: If a model object has been previously dirtied, `save` should be used rather than `update`.

bensheldon (approved):
👍 thank you for this feature! I will merge and then make a quick PR to address my style problems.

lib/good_job/job.rb
morgoth: @bensheldon Can I prepare a PR for renaming reperform_jobs_on_standard_error to reperform_jobs_on_unhandled_error with deprecation warning on old usage?

bensheldon: @morgoth that would be great! 

Question: what are your thoughts about an even briefer `retry_on_unhandled_error`?

PR #152, opened 2020-09-23
Extract Scheduler polling behavior to its own object

### LINKED ISSUE #128: Extract polling from scheduler into Polling object
Create a new object with a similar interface to the Notifier that performs the polling. 

This simplifies the implementation of the Scheduler by only having a single threadpool to manage, and can hopefully provide symmetry between the interfaces of 3 objects: Scheduler, Notifier, Poller.

bensheldon:
Closes #128

### DIFF
--- a/lib/good_job/poller.rb
+++ b/lib/good_job/poller.rb
@@ -1,0 +1,94 @@
+require 'concurrent/atomic/atomic_boolean'
+
+module GoodJob # :nodoc:
+  #
+  # Pollers regularly wake up execution threads to check for new work.
+  #
+  class Poller
+    # Defaults for instance of Concurrent::TimerTask.
+    # The timer controls how and when sleeping threads check for new work.
+    DEFAULT_TIMER_OPTIONS = {
+      execution_interval: Configuration::DEFAULT_POLL_INTERVAL,
+      timeout_interval: 1,
+      run_now: true,
+    }.freeze
+
+    # @!attribute [r] instances
+    #   @!scope class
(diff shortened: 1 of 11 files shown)


lib/good_job/poller.rb
```diff
@@ -0,0 +1,95 @@
...
+  # Pollers regularly wake up execution threads to check for new work.
+  #
+  class Poller
+    # Default poll interval.
+    DEFAULT_POLL_INTERVAL = 1
+
+    # Defaults for instance of Concurrent::TimerTask.
+    # The timer controls how and when sleeping threads check for new work.
+    DEFAULT_TIMER_OPTIONS = {
+      execution_interval: DEFAULT_POLL_INTERVAL,
```
bensheldon, line 14: This should be `Configuration::DEFAULT_POLL_INTERVAL`

lib/good_job/poller.rb
```diff
@@ -0,0 +1,95 @@
...
+    def self.from_configuration(configuration)
+      GoodJob::Poller.new(poll_interval: configuration.poll_interval)
+    end
+
+    # List of recipients that will receive notifications.
+    # @return [Array<#call, Array(Object, Symbol)>]
+    attr_reader :recipients
+
+    # @param recipients [Array<#call, Array(Object, Symbol)>]
+    # @param timer_options [Hash] Options to instantiate a Concurrent::TimerTask
```
bensheldon, line 34: This param needs to be updated.

PR #153, opened 2020-09-25
Add info how to setup basic auth for engine

morgoth:
I think it's useful as it's not obvious how to apply basic auth on Rails engine (I didn't find it in Rails guides nor SO).


bensheldon:
@morgoth thank you! 🙌

PR #154, opened 2020-10-01
Rename reperform_jobs_on_standard_error to retry_on_unhandled_error

morgoth:
Preserves backward compatibility.

Follow up to https://github.com/bensheldon/good_job/pull/145#discussion_r492069976

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -37,14 +37,30 @@ module GoodJob
   #   @return [Boolean]
   mattr_accessor :preserve_job_records, default: false
 
-  # @!attribute [rw] reperform_jobs_on_standard_error
+  # @!attribute [rw] retry_on_unhandled_error
   #   @!scope class
   #   Whether to re-perform a job when a type of +StandardError+ is raised to GoodJob (default: +true+).
   #   If +true+, causes jobs to be re-queued and retried if they raise an instance of +StandardError+.
   #   If +false+, jobs will be discarded or marked as finished if they raise an instance of +StandardError+.
   #   Instances of +Exception+, like +SIGINT+, will *always* be retried, regardless of this attribute's value.
   #   @return [Boolean]
-  mattr_accessor :reperform_jobs_on_standard_error, default: true
+  mattr_accessor :retry_on_unhandled_error, default: true
+
+  # @deprecated Use {GoodJob#retry_on_unhandled_error} instead.
+  def self.reperform_jobs_on_standard_error
+    ActiveSupport::Deprecation.warn(
+      "Calling 'GoodJob.reperform_jobs_on_standard_error' is deprecated. Please use 'retry_on_unhandled_error'"
+    )
+    retry_on_unhandled_error
+  end
+
+  # @deprecated Use {GoodJob#retry_on_unhandled_error=} instead.
+  def self.reperform_jobs_on_standard_error=(value)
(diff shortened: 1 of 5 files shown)


lib/good_job.rb
```diff
@@ -37,14 +37,30 @@ module GoodJob
...
   #   @!scope class
   #   Whether to re-perform a job when a type of +StandardError+ is raised to GoodJob (default: +true+).
   #   If +true+, causes jobs to be re-queued and retried if they raise an instance of +StandardError+.
   #   If +false+, jobs will be discarded or marked as finished if they raise an instance of +StandardError+.
   #   Instances of +Exception+, like +SIGINT+, will *always* be retried, regardless of this attribute's value.
   #   @return [Boolean]
-  mattr_accessor :reperform_jobs_on_standard_error, default: true
+  mattr_accessor :retry_on_unhandled_error, default: true
+
+  # Deprecated
```
morgoth, line 49: Do you know a proper way of documenting deprecations in yard?

bensheldon: `@depreacted` is the tag to use:

```suggestion
  # @deprecated Use {GoodJob#retry_on_unhandled_error} instead.
```

Screen Shot 2020-10-01 at 8 50 53 AM

morgoth: Thanks, updated.

spec/lib/good_job/job_spec.rb
```diff
@@ -265,10 +265,25 @@ def perform(result_value = nil, raise_error: false)
...
+
+          GoodJob.reperform_jobs_on_standard_error = false
+
+          expect(GoodJob.preserve_job_records).to be(false)
+          expect(GoodJob.reperform_jobs_on_standard_error).to be(false)
+          expect(ActiveSupport::Deprecation).to have_received(:warn).twice
+        ensure
+          GoodJob.preserve_job_records = current_preserve_job_records
+        end
+      end
```
bensheldon, line 281: It would be best for this to live in in `good_job_spec.rb` instead of `job_spec.rb` because these are testing methods on `GoodJob`.

Here is some suggested (potentially overly-) idiomatic RSpec to replace the `ensure` reset with an `around` block, and a compound expectation that assigning a value changes the values:

```suggestion
  describe '.reperform_jobs_on_standard_error' do
    around do |example|
      original_retry_on_unhandled_error = GoodJob.retry_on_unhandled_error
      example.run
      GoodJob.retry_on_unhandled_error = original_retry_on_unhandled_error
    end

    it 'is deprecated and replaced with .retry_on_unhandled_error' do
      allow(ActiveSupport::Deprecation).to receive(:warn)
      GoodJob.retry_on_unhandled_error = true

      expect {
        GoodJob.reperform_jobs_on_standard_error = false
      }.to(
        change(GoodJob, :retry_on_unhandled_error).from(true).to(false)
          .and change(GoodJob, :reperform_jobs_on_standard_error).from(true).to(false)
      )

      expect(ActiveSupport::Deprecation).to have_received(:warn).exactly(3).times
    end
  end
```

bensheldon (approved):
This looks great!  I made one suggestion about cleaning up the tests.

spec/lib/good_job/job_spec.rb
morgoth: done

bensheldon (approved):
Thank you 🎉

Issue #90, opened 2020-08-24
Why poll when using LISTEN/NOTIFY?


sj26:
Absolutely loving good_job — good job! 😅

I'm curious why it polls, though, when it supports LISTEN/NOTIFY?

bensheldon:
Thanks for the kind words 🙌 

LISTEN/NOTIFY works for jobs scheduled immediately (`ExampleJob.perform_later`), but isn't used for future-scheduled jobs (`ExampleJob.set(wait: 1.day).perform_later` or retry with exponential backoff ). Polling is necessary to pick up those latter jobs. LISTEN/NOTIFY allows increasing the poll interval (On my applications I set it to 10 seconds), but some applications may want greater scheduling accuracy so the default is still 1 second. Though I'm not committed to any of these constants.

Just while I'm thinking of it. Currently the polling implementation is integrated with the Scheduler object, whereas it could also be integrated with the Notifier object. A process may have multiple schedulers but only a single notifier. For some configurations having the Notifier implement polling could be more efficient, but I'm not sure.

slimdave:
If the GoodJob process wasn't running when a job was enqueued, would the polling be responsible for picking that job up when the GoodJob process starts also?

bensheldon:
> If the GoodJob process wasn't running when a job was enqueued, would the polling be responsible for picking that job up when the GoodJob process starts also?

Yes 👍 

Polling is _necessary_ for proper functioning, and LISTEN/NOTIFY is _nice-to-have_.

Though if you did want to disable polling, that's possible with `GOOD_JOB_POLL_INTERVAL=-1`. Though that would lead to jobs not being executed for all the previously stated reasons.

slimdave:
Ah hah, clever stuff. If you had multiple processes monitoring the queue, then they would all get the NOTIFY but only one would be able to grab the advisory lock and the rest would continue listening and polling?

sj26:
Interesting!

What I'd imagine is that enqueuing new jobs, immediate or scheduled, would post a simple "changed" message via NOTIFY. Then there would be a supervisor process which LISTENs, see the "changed" message, and checks to see if there's new immediate work — and kick off workers — or if the next scheduled time has changed, and then go back to sleep until the next scheduled job or LISTEN message.

bensheldon:
Having the Scheduler itself track when the next scheduled job is expected would be interesting. If you want to dive into the code you could make a proof of concept.

[...]

PR #155, opened 2020-10-03
Create a Timer object that will wake up the Scheduler at specific times

bensheldon:
The intention is to allow a Scheduler to store the times of future scheduled tasks and trigger a wake at those times. 

My goal is to allow polling to be significantly dialed back (~1 minute), if not done away with entirely and allow the `:async` execution mode to be non-noisily set as the default in the Rails development environment. (#139 and #90)

Current concerns:
- What is the appropriate number of future times to store? I'm a little worried that an unbounded queue is unsafe in production (uses a lot of memory and overlaps with the database's responsibility). But I do think it would make sense in the development environment and match Rails's own [AsyncAdapter](https://github.com/rails/rails/blob/166b63eb674b56faf4f75dfe300b14f00fb79dae/activejob/lib/active_job/queue_adapters/async_adapter.rb#L75-L82).
- Complexity and yet another `Concurent::ThreadPoolExecutor`. The `Timer` object has its own threadpool to execute `Concurrent::ScheduledTasks` that then wake up the `GoodJob::Scheduler` to fetch and perform the actual job. The `GoodJob::Timer` is currently entangled with both the `Scheduler` and the `Performer` because the Timer needs to refresh its queue of future tasks. I'm struggling with whether the `Timer` is an implementation detail of the `Scheduler`, or whether it can be non-complexly extracted like the Poller in #152.

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -82,6 +82,7 @@ module GoodJob
   def self.shutdown(wait: true)
     Notifier.instances.each { |notifier| notifier.shutdown(wait: wait) }
     Scheduler.instances.each { |scheduler| scheduler.shutdown(wait: wait) }
+    Timer.instances.each { |timer| timer.shutdown(wait: wait) }
   end
 
   # Tests whether jobs have stopped executing.
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -78,8 +78,12 @@ module GoodJob
         end
       end
 
-      executed_locally = execute_async? && @scheduler.create_thread(queue_name: good_job.queue_name)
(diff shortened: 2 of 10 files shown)


bensheldon:
Thinking about this more, this can live entirely within the `Scheduler` which will enqueue either a `Concurrent::Future` for immediate execution, or a `Concurrent::ScheduledTask` for scheduled execution. I'll keep the `queue` idea that the Scheduler manage a bounded number of ScheduledTasks, though in async mode, by default, I might make it unbounded/infinite to match Rails async adapter.

bensheldon:
Replaced by #163 to use a `Concurrent::ScheduledTask` directly.

Issue #89, opened 2020-08-24
ActiveRecord::ConnectionNotEstablished when using async mode


sj26:
I have a fairly standard new-ish rails app setup with puma, and added the following:

```ruby
# config/initializers/development.rb
Rails.application.configure do
  # ...
  config.active_job.queue_adapter = GoodJob::Adapter.new(execution_mode: :async)
  # ...
end
```

Then ran:

```
DISABLE_SPRING=1 bin/rails server
```

But got:

```
❯ DISABLE_SPRING=1 bin/rails server
=> Booting Puma
=> Rails 6.0.3.2 application starting in development 
=> Run `rails server --help` for more startup options
/usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/adapter.rb:11: warning: Passing the keyword argument as the last hash parameter is deprecated
/usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/configuration.rb:5: warning: The called method `initialize' is defined here
Exiting
Traceback (most recent call last):
...
	 7: from /Users/sj26/Projects/TheSizzle/config/environments/development.rb:34:in `block in <main>'
	 6: from /Users/sj26/Projects/TheSizzle/config/environments/development.rb:34:in `new'
	 5: from /usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/adapter.rb:23:in `initialize'
	 4: from /usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/scheduler.rb:42:in `from_configuration'
	 3: from /usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/scheduler.rb:42:in `map'
	 2: from /usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/scheduler.rb:44:in `block in from_configuration'
	 1: from /usr/local/rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/good_job-1.2.1/lib/good_job/configuration.rb:37:in `max_threads'
[...]

PR #161, opened 2020-10-19
Don't disconnect a nil activerecord connection

bensheldon:
Related to #89.

Being seen on Heroku when precompiling assets and (assumed) the database is not available.

### DIFF
--- a/lib/good_job/notifier.rb
+++ b/lib/good_job/notifier.rb
@@ -147,7 +147,7 @@ module GoodJob # :nodoc:
       pg_conn.exec("SET application_name = #{pg_conn.escape_identifier(self.class.name)}")
       yield pg_conn
     ensure
-      ar_conn.disconnect!
+      ar_conn&.disconnect!
     end
   end
 end


byellokore:
**The error changed:**

remote:        I, [2020-10-20T00:18:08.386654 #399]  INFO -- : [GoodJob] Notifier unsubscribed with UNLISTEN
remote:        E, [2020-10-20T00:18:08.387770 #399] ERROR -- : [GoodJob] Notifier errored: could not connect to server: Connection refused
remote:         Is the server running on host "127.0.0.1" and accepting
remote:         TCP/IP connections on port 5432?


However, I was checking the instructions again and I have on my **production.rb** this config:

`config.active_job.queue_adapter = GoodJob::Adapter.new(execution_mode: :async, max_threads: 2, poll_interval: 5)`

if I remove this line I don't see any errors. So, I'm not sure if I was doing something wrong or not. @bensheldon

bensheldon:
@byellokore If you're seeing this during the build stage, I think the problem here is that Heroku does not expose environment variables (like the `DATABASE_URL`) to the application during the build stage. So it's trying to connect to the database and failing.

I think a workaround is this: https://devcenter.heroku.com/articles/rails-asset-pipeline#failures-in-the-assets-precompile-task

But that's not pretty. 

My better recommendation is to use the environment variable `GOOD_JOB_EXECUTION_MODE=async` in your Heroku environment.

byellokore:
@bensheldon Hi! 

I tried the environment variables but I was no able to send new builds to Heroku. However, if I don't set any environment variable and just keep my application.rb with 

`config.active_job.queue_adapter = :good_job `

It works fine! However It working only with the default configuration.

PR #170, opened 2020-10-28
Format serialized params to ease reading

morgoth:
In addition move params to the very end as it can be huge and next columns could be missed.

image


bensheldon:
@morgoth Thank you! This is a really nice change.

PR #173, opened 2020-11-22
UI: Admin UI with filters and space efficient layout

zealot128:
Happy to see a basic admin layout! Some changes I might propose

- ``container-fluid`` instead of container - Very useful for larger tables..., table-sm to preserve whitespace for a more space efficient table
- Basic filter for: Job classes and finished/unfinished fields

Not sure where to put the filter ruby class from the code perspective.

Bildschirmfoto 2020-11-22 um 20 10 36

### DIFF
--- a/engine/app/controllers/good_job/dashboards_controller.rb
+++ b/engine/app/controllers/good_job/dashboards_controller.rb
@@ -1,8 +1,57 @@
 module GoodJob
   class DashboardsController < GoodJob::BaseController
-    def index
-      @jobs = GoodJob::Job.display_all(after_scheduled_at: params[:after_scheduled_at], after_id: params[:after_id])
+    class JobFilter
+      attr_accessor :params
+
+      def initialize(params)
+        @params = params
+      end
+
+      def last
+        @_last ||= jobs.last
+      end
+
+      def jobs
+        sql = GoodJob::Job.display_all(after_scheduled_at: params[:after_scheduled_at], after_id: params[:after_id])
                           .limit(params.fetch(:limit, 10))
+        if params[:job_class] # rubocop:disable Style/IfUnlessModifier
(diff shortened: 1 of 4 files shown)


engine/app/controllers/good_job/dashboards_controller.rb
```diff
@@ -1,8 +1,56 @@
...
+          sql = sql.where("serialized_params->>'job_class' = ?", params[:job_class])
+        end
+        if params[:state]
+          case params[:state]
+          when 'finished'
+            sql = sql.finished
+          when 'unfinished'
+            sql = sql.unfinished
+          when 'errors'
+            sql = sql.where.not(error: nil).count
```
morgoth, line 26: I believe the `count` should not be added

zealot128: oh, copy&paste error from the facets. fixed it.

bensheldon (approved):
@zealot128 Thank you for the contribution. This is a very nice improvement.

I have two suggestions, but I'm willing to merge this as is and accept the improvements as later PRs:

1. I think the filters should always be visible, even if there are no jobs. Currently they disappear if a filter is selected that results in no jobs.
2. When a filter is selected, the link button should be in an active state with `active` class

zealot128:
> @zealot128 Thank you for the contribution. This is a very nice improvement.
> 
> I have two suggestions, but I'm willing to merge this as is and accept the improvements as later PRs:
> 
>     1. I think the filters should always be visible, even if there are no jobs. Currently they disappear if a filter is selected that results in no jobs.
> 
>     2. When a filter is selected, the link button should be in an active state with `active` class

@bensheldon Good suggestions! I've updated the branch. It now looks like this:

Bildschirmfoto 2020-11-29 um 14 23 16

bensheldon:
@zealot128 Thank you so much!  I just released this in `v1.3.3`

PR #180, opened 2020-12-27
Run CI against Rails 6.1 and Ruby 3

bensheldon:


bensheldon:
Broken out into separate PRs #182 and #183.

PR #193, opened 2021-01-10
Fix JRuby version number

tedhexaflow:


bensheldon:
@tedhexaflow thank you!

Issue #195, opened 2021-01-14
Rails 6.1 & async - `queue_parser': undefined method `first' for "*":String (NoMethodError)


tedhexaflow:
Hello Ben,

There is an issue starting up the server with Rails 6.1 (both 6.1.0 & 6.1.1) for async. 
config.active_job.queue_adapter = GoodJob::Adapter.new(execution_mode: :async, max_threads: 5, poll_interval: 30)

This is not specific to JRuby. 
I've tested with ruby-2.5.7, ruby-2.7.2 and the results are the same.
It is working fine with Rails 6.0.3.4.

I've made a rails app for testing. https://github.com/tedhexaflow/good_job_start_up
I've also tested with https://github.com/bensheldon/good_job/tree/main/spec/test_app, and the results are the same.

Can you please have a look and let me know if you need more information.

Thanks.
Ted

=> Booting Puma
=> Rails 6.1.1 application starting in development 
=> Run `bin/rails server --help` for more startup options
Exiting
Traceback (most recent call last):
        72: from bin/rails:4:in `<main>'
        71: from bin/rails:4:in `require'
        70: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/railties-6.1.1/lib/rails/commands.rb:18:in `<top (required)>'
        69: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/railties-6.1.1/lib/rails/command.rb:50:in `invoke'
        68: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/railties-6.1.1/lib/rails/command/base.rb:69:in `perform'
        67: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/thor-1.0.1/lib/thor.rb:392:in `dispatch'
        66: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/thor-1.0.1/lib/thor/invocation.rb:127:in `invoke_command'
        65: from /Users/tedhexaflow/.asdf/installs/ruby/2.5.7/lib/ruby/gems/2.5.0/gems/thor-1.0.1/lib/thor/command.rb:27:in `run'
[...]

PR #196, opened 2021-01-14
Fix start-up issue with Rails 6.1 & async mode

tedhexaflow:
I'm not sure this is an acceptable fix for https://github.com/bensheldon/good_job/issues/195, but this would not raise `queue_parser': undefined method `first' for "*":String (NoMethodError) anymore.

I also notice the start-up time for async mode is significantly slower than other modes. (It took more than 5 minutes to start-up).

I'm not sure it is related to this fix or any other factor of rails 6.1.

### DIFF
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -30,7 +30,7 @@ module GoodJob
     def self.queue_parser(string)
       string = string.presence || '*'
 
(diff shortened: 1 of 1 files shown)


tedhexaflow:
closed in favor of https://github.com/bensheldon/good_job/pull/199

PR #199, opened 2021-01-16
Configure GoodJob via `Rails.application.config` instead of recommending `GoodJob::Adapter.new`

bensheldon:
This is a moderately significant change in recommended configuration. 

The problem is that calling `GoodJob::Adapter.new` within `config/application.rb` or `config/environments/*.rb` causes the Adapter to be instantiated too early in the Rails initialization process. 

This addresses #195, #89.

This also contains changes to `GoodJob::Performer`, making it less generic and renaming it to `GoodJob::JobPerformer`. The intent is to  to clearly delay references to GoodJob::Job. This is necessary because there is a circular dependency in Rails 6.1 when ActiveJob initializes GoodJob, which references ActiveRecord, which tries to initialize ActiveJob (for the destroy associations async feature).


README.md
```diff
@@ -189,9 +189,31 @@ If you are preserving job records this way, use this command regularly
...
+config.active_job.queue_adapter = :good_job
+
+# Configure options individually...
+config.good_job.execution_mode = :async
+config.good_job.max_threads = 5
+config.good_job.poll_interval = 30 # seconds
+
+# ...or all at once.
+config.good_job = {
+  execution_mode = :async,
```
nnc, line 210: @bensheldon this should probably be `execution_mode: :async`?

Thanks for Good Job, btw. It's doing a Great job so far. 💯

bensheldon: Thank you for catching this! 🙌

PR #226, opened 2021-03-09
Remove leftover text

weh:
I think this text snippet was left from the last change on README.md


bensheldon (approved):
Thanks for the clean up! 💯

PR #235, opened 2021-04-14
Add initial Sorbet linting; greatly improve YARD docs

bensheldon:
- Using `sord` gem for generating RBI file from documentation
- Not yet packaging a `good_job.rbi` file


bensheldon:
This PR is blocked by dynamically assigning a class parent: https://github.com/bensheldon/good_job/blob/505d0168fa60d8c0e163eac5ddb0e5fe8b095b6b/lib/good_job/job.rb#L5

I posted a question on StackOverflow following Sorbet's support instructions: https://stackoverflow.com/questions/67182040/how-to-hint-sorbet-with-const-get-and-dynamic-parent-class

Issue #236, opened 2021-04-15
No connection pool for 'ActiveRecord::Base' found


morgoth:
I start good job, by `bundle exec good_job start --poll-interval=1`

for first few jobs that are enqueued, I got the stacktrace:
```
[GoodJob] GoodJob error: No connection pool for 'ActiveRecord::Base' found.
"/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:1125:in `retrieve_connection'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/connection_handling.rb:327:in `retrieve_connection'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/connection_handling.rb:283:in `connection'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/relation/delegation.rb:93:in `connection'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/relation/query_methods.rb:1394:in `preprocess_order_args'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/relation/query_methods.rb:360:in `order!'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/relation/query_methods.rb:355:in `order'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/good_job-1.8.0/lib/good_job/job.rb:75:in `block in <class:Job>'",
 "/home/wojtek/.asdf/installs/ruby/2.7.3/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3.1/lib/active_record/relation.rb:411:in `instance_exec'",
[...]

PR #238, opened 2021-04-16
Allow to specify parent class for active record

morgoth:
Refs #236

POC

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -17,7 +17,13 @@ require "good_job/railtie"
 #
 # +GoodJob+ is the top-level namespace and exposes configuration attributes.
 module GoodJob
-  # TODO: explain me
+  # @!attribute [rw] active_record_parent_class
+  #   @!scope class
+  #   The ActiveRecord parent class inherited by +GoodJob::Job+ (default: +ActiveRecord::Base+).
+  #   Use this when using multiple databases or other custom ActiveRecord configuration.
+  #   @return [ActiveRecord::Base]
+  #   @example Change the base class:
+  #     GoodJob.active_record_parent_class = "CustomApplicationRecord"
   mattr_accessor :active_record_parent_class, default: "ActiveRecord::Base"
 
   # @!attribute [rw] logger
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -17,6 +17,9 @@ require "good_job/railtie"
 #
 # +GoodJob+ is the top-level namespace and exposes configuration attributes.
 module GoodJob
+  # TODO: explain me
+  mattr_accessor :active_record_parent_class, default: "ActiveRecord::Base"
+
   # @!attribute [rw] logger
   #   @!scope class
   #   The logger used by GoodJob (default: +Rails.logger+).
--- a/lib/good_job/job.rb
+++ b/lib/good_job/job.rb
@@ -2,7 +2,7 @@ module GoodJob
   #
   # Represents a request to perform an +ActiveJob+ job.
   #
-  class Job < ActiveRecord::Base
+  class Job < Object.const_get(GoodJob.active_record_parent_class)
     include Lockable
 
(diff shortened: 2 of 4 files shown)


lib/good_job/lockable.rb
```diff
@@ -143,7 +143,7 @@ def with_advisory_lock
       def supports_cte_materialization_specifiers?
         return @_supports_cte_materialization_specifiers if defined?(@_supports_cte_materialization_specifiers)
 
-        @_supports_cte_materialization_specifiers = ActiveRecord::Base.connection.postgresql_version >= 120000
+        @_supports_cte_materialization_specifiers = Job.connection.postgresql_version >= 120000
```
bensheldon, line 146: The `GoodJob::Lockable` class is intended to be a generic ActiveRecord mixin, so this hopefully can be:

```suggestion
        @_supports_cte_materialization_specifiers = self.class.connection.postgresql_version >= 120000
```

lib/good_job/lockable.rb
```diff
@@ -158,7 +158,7 @@ def advisory_lock
         WHERE pg_try_advisory_lock(('x'||substr(md5($1 || $2::text), 1, 16))::bit(64)::bigint)
       SQL
       binds = [[nil, self.class.table_name], [nil, send(self.class.primary_key)]]
-      ActiveRecord::Base.connection.exec_query(pg_or_jdbc_query(query), 'GoodJob::Lockable Advisory Lock', binds).any?
+      Job.connection.exec_query(pg_or_jdbc_query(query), 'GoodJob::Lockable Advisory Lock', binds).any?
```
bensheldon, line 161: Same here:

```suggestion
      self.class.connection.exec_query(pg_or_jdbc_query(query), 'GoodJob::Lockable Advisory Lock', binds).any?
```

lib/good_job/job.rb
```diff
@@ -2,7 +2,7 @@ module GoodJob
   #
   # Represents a request to perform an +ActiveJob+ job.
   #
-  class Job < ActiveRecord::Base
+  class Job < Object.const_get(GoodJob.active_record_parent_class)
```
bensheldon, line 5: 👍

lib/good_job.rb
```diff
@@ -17,6 +17,9 @@
 #
 # +GoodJob+ is the top-level namespace and exposes configuration attributes.
 module GoodJob
+  # TODO: explain me
```
bensheldon, line 20:
```suggestion
  # @!attribute [rw] active_record_parent_class
  #   @!scope class
  #   The ActiveRecord parent class inherited by +GoodJob::Job+ (default: +ActiveRecord::Base+).
  #   Use this when using multiple databases or other custom ActiveRecord configuration.
  #   @return [ActiveRecord::Base]
  #   @example Change the base class:
  #     GoodJob.active_record_parent_class = "CustomApplicationRecord"
```

bensheldon (commented):
This looks very reasonable 👍

lib/good_job/lockable.rb
morgoth: hmm, this doesn't work somehow

bensheldon: Whoops. This is within `class_methods` so it just needs to be: `connection.postgresql_version >= 120000`

morgoth:
@bensheldon Anything more I should add to make it merged? I was thinking about some spec, but it might be tricky to test inheritance as the job file is already loaded on spec run.

bensheldon:
@morgoth I agree about the difficulty of testing. It seems like it will either work, or fail across the board. Let's move it forward and see what we learn.

bensheldon:
I forgot that this should also be documented in the Readme. Done in  https://github.com/bensheldon/good_job/commit/9eac0bed6cd4c712d47a1b00e2d5f9224de9a7a1

bensheldon:
This has been released in `v1.9.1`

PR #251, opened 2021-05-14
Fix dashboard jobs endless pagination with timezone handling

morgoth:
Having `config.time_zone = "Europe/Berlin"` in the app, to properly handle time zone, we need to parse timestamp.
The difference before the patch:

```
irb(main):033:0> GoodJob::Job.display_all(after_scheduled_at: '2021-04-21 10:36:12 +0200').count
   (1.3ms)  SELECT COUNT(*) FROM "good_jobs" WHERE ((COALESCE(scheduled_at, created_at)) < ('2021-04-21 10:36:12 +0200'))
=> 1
irb(main):034:0> GoodJob::Job.display_all(after_scheduled_at: Time.zone.parse('2021-04-21 10:36:12 +0200')).count
   (1.2ms)  SELECT COUNT(*) FROM "good_jobs" WHERE ((COALESCE(scheduled_at, created_at)) < ('2021-04-21 08:36:12'))
=> 0
```

### DIFF
--- a/engine/app/controllers/good_job/dashboards_controller.rb
+++ b/engine/app/controllers/good_job/dashboards_controller.rb
@@ -12,7 +12,8 @@ module GoodJob
       end
 
       def jobs
-        sql = GoodJob::Job.display_all(after_scheduled_at: params[:after_scheduled_at], after_id: params[:after_id])
+        after_scheduled_at = params[:after_scheduled_at].present? ? Time.zone.parse(params[:after_scheduled_at]) : nil
+        sql = GoodJob::Job.display_all(after_scheduled_at: after_scheduled_at, after_id: params[:after_id])
                           .limit(params.fetch(:limit, 10))
         if params[:job_class] # rubocop:disable Style/IfUnlessModifier
           sql = sql.where("serialized_params->>'job_class' = ?", params[:job_class])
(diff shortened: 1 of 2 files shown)


lib/good_job/job.rb
```diff
@@ -126,6 +126,8 @@ def self.queue_parser(string)
     # @return [ActiveRecord::Relation]
     scope :display_all, (lambda do |after_scheduled_at: nil, after_id: nil|
       query = order(Arel.sql('COALESCE(scheduled_at, created_at) DESC, id DESC'))
+      after_scheduled_at = Time.zone.parse(after_scheduled_at) if after_scheduled_at.present?
```
bensheldon, line 129: This will fail if `after_scheduled_at` is a `Time` or `DateTime`. My preference would be that type coercion happens in the controller (close to where it's coming in as a string), and models operate off of higher level objects (like `Time`/`DateTime`) rather than primitives (like Strings)

bensheldon (commented):
@morgoth I'm so glad you caught this! I requested a change about where the type coercion happens. Let me know if you have thoughts about that.

bensheldon:
@morgoth Also, I apologize that there are some flakes in the Java Test Suite right now. I'm just re-running checks if they have an isolated failure and hopefully building up an understanding of where the flake is.

lib/good_job/job.rb
morgoth: updated

Issue #258, opened 2021-05-17
Update from bootstrap 4 to bootstrap 5


morgoth:
The bootstrap 5 i out and this project uses version 4.

BTW what is the motivation to use https://github.com/bensheldon/good_job/blob/main/engine/app/assets/vendor/bootstrap/bootstrap-native.js ? Is it to not use jQuery? In bootstarp 5 there is no jQuery any more so maybe it can be dropped to simplify setup and reduce dependencies?

bensheldon:
Thanks! It looks like Bootstrap 5 should have a fairly simple upgrade path. 

Yes, `bootstrap-native.js` was to avoid adding jquery. Which isn't really a big deal, but was one less thing to add.

FYI, GoodJob's frontend asset dependencies are kind've weird in how they're loaded. I didn't believe I could assume that the end-user using sprocket-based asset pipeline. But unless I'm overlooking something obvious, I don't think it's worth the trouble to make any big changes to how the assets are served.

morgoth:
> It looks like Bootstrap 5 should have a fairly simple upgrade path.

Yes, I actually wanted to do that, but wasn't sure about this bootstrap-native js.

> I didn't believe I could assume that the end-user using sprocket-based asset pipeline

Correct. In one of our project we don't have sprockets at all. It's fully managed by webpacker.

> I don't think it's worth the trouble to make any big changes to how the assets are served.

Sure, I guess even easier would be to directly link to CDN hosted files, but it's not a big difference.

I'll try to upgrade it tomorrow.

bensheldon:
> Sure, I guess even easier would be to directly link to CDN hosted files, but it's not a big difference.

It would be easier, but it's possible someone is using GoodJob in a controlled environment, which is why everything is self-hosted.

> I'll try to upgrade it tomorrow.

🤩

PR #260, opened 2021-05-18
Update Dashboard to Bootstrap 5

morgoth:
Looks like for current functionality, there is no need for js at all.
It can be added when such need will arrive.

Closes https://github.com/bensheldon/good_job/issues/258


bensheldon:
@morgoth I fixed the Gemfile issue on the main branch (#261). If you rebase against main I think this will pass.

morgoth:
@bensheldon it helped. Thanks

bensheldon (approved):
👍 thank you!

bensheldon:
@morgoth this is released in `v1.9.5`

PR #263, opened 2021-05-26
Collapse Dashboard params by default

morgoth:
Having expanded all of the params is very noisy. With this change we can also pack more jobs on page

image

### DIFF
--- a/engine/app/views/shared/_jobs_table.erb
+++ b/engine/app/views/shared/_jobs_table.erb
@@ -19,7 +19,13 @@
             <td><%= job.queue_name %></td>
             <td><%= job.scheduled_at || job.created_at %></td>
             <td><%= job.error %></td>
-            <td><pre><%= JSON.pretty_generate(job.serialized_params) %></pre></td>
+            <td>
+              <%= tag.button "Preview", type: "button", class: "btn btn-sm btn-outline-primary", role: "button",
+                data: {bs_toggle: "collapse", bs_target: "##{dom_id(job, 'params')}"},
+                aria: {expanded: false, controls: dom_id(job, "params")}
+              %>
+              <%= tag.pre JSON.pretty_generate(job.serialized_params), id: dom_id(job, "params"), class: "collapse" %>
+            </td>
           </tr>
         <% end %>
       </tbody>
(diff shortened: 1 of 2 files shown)


engine/app/views/shared/_jobs_table.erb
```diff
@@ -19,7 +19,10 @@
             <td><%= job.queue_name %></td>
             <td><%= job.scheduled_at || job.created_at %></td>
             <td><%= job.error %></td>
-            <td><pre><%= JSON.pretty_generate(job.serialized_params) %></pre></td>
+            <td>
+              <%= link_to "Preview", "#params-#{job.id}", "role" => "button", "data-bs-toggle" => "collapse", "aria-expanded" => false, "aria-controls" => "#params-#{job.id}" %>
+              <pre class="collapse" id="params-<%= job.id %>"><%= JSON.pretty_generate(job.serialized_params) %></pre>
```
bensheldon, line 24: It's more semantic/accessible to use a button rather than link for the toggle. I also suggest using the `dom_id(job, "params")` helper to construct the target.  

```suggestion
              <% collapse_target = "#{dom_id(job)}-params" %>
              <%= button_tag "Toggle Params", class: "btn btn-link", type: "button", "data-bs-toggle": "collapse", "data-bs-target": "##{collapse_target}", "aria-expanded": false, "aria-controls": "##{collapse_target}" %>
              <pre class="collapse" id="<%= collapse_target %>"><%= JSON.pretty_generate(job.serialized_params) %></pre>
```

bensheldon (approved):
I think this is great 👍  I left some HTML suggestions.

This does make me think about how to make specific jobs easier to find. I'm a big user of `CMD-F` to search within a page and collapsing content does make that more difficult. 

I'm good moving forward with this and seeing if we get feedback. Some future ideas:

- Adding a toggle to the column header that toggles them all
- A searchbar function in the header; another benefit of postgres is fulltext search (even without an index)

engine/app/views/shared/_jobs_table.erb
bensheldon: 👍

PR #265, opened 2021-05-28
Add deleting jobs from UI.

morgoth:
refs #256

image

### DIFF
--- a/engine/app/controllers/good_job/active_jobs_controller.rb
+++ b/engine/app/controllers/good_job/active_jobs_controller.rb
@@ -4,11 +4,5 @@ module GoodJob
       @jobs = GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id])
                           .order(Arel.sql("COALESCE(scheduled_at, created_at) DESC"))
     end
-
-    def destroy
-      deleted_count = GoodJob::Job.where(id: params[:id]).delete_all
-      message = deleted_count.positive? ? { notice: "Job deleted" } : { alert: "Job not deleted" }
-      redirect_to root_path, **message
-    end
   end
 end
--- a/engine/app/controllers/good_job/active_jobs_controller.rb
+++ b/engine/app/controllers/good_job/active_jobs_controller.rb
@@ -6,8 +6,9 @@ module GoodJob
     end
 
     def destroy
-      GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id]).delete_all
-      redirect_to root_path, notice: "Job deleted"
+      deleted_count = GoodJob::Job.where(id: params[:id]).delete_all
+      message = deleted_count.positive? ? { notice: "Job deleted" } : { alert: "Job not deleted" }
+      redirect_to root_path, **message
     end
   end
 end
--- a/engine/app/controllers/good_job/active_jobs_controller.rb
+++ b/engine/app/controllers/good_job/active_jobs_controller.rb
@@ -4,5 +4,10 @@ module GoodJob
       @jobs = GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id])
                           .order(Arel.sql("COALESCE(scheduled_at, created_at) DESC"))
     end
+
+    def destroy
+      GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id]).delete_all
+      redirect_to root_path, notice: "Job deleted"
+    end
   end
 end
--- a/engine/app/views/layouts/good_job/base.html.erb
+++ b/engine/app/views/layouts/good_job/base.html.erb
@@ -51,6 +51,15 @@
       </div>
     </div>
 
+    <% if notice %>
+      <div class="alert alert-success alert-dismissible fade show d-flex align-items-center offset-md-3 col-6" role="alert">
+        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-check-circle-fill flex-shrink-0 me-2" viewBox="0 0 16 16">
+          <path d="M16 8A8 8 0 1 1 0 8a8 8 0 0 1 16 0zm-3.97-3.03a.75.75 0 0 0-1.08.022L7.477 9.417 5.384 7.323a.75.75 0 0 0-1.06 1.06L6.97 11.03a.75.75 0 0 0 1.079-.02l3.992-4.99a.75.75 0 0 0-.01-1.05z" />
+        </svg>
+        <div><%= notice %></div>
+        <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
+      </div>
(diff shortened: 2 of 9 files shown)


engine/app/controllers/good_job/active_jobs_controller.rb
```diff
@@ -4,5 +4,10 @@ def show
       @jobs = GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id])
                           .order(Arel.sql("COALESCE(scheduled_at, created_at) DESC"))
     end
+
+    def destroy
+      GoodJob::Job.where("serialized_params ->> 'job_id' = ?", params[:id]).delete_all
```
morgoth, line 9: I used the same lookup as for show page, but why are we using AJ job id here? Wouldn't be easier to just use db id?

spec/system/dashboard_spec.rb
```diff
@@ -5,4 +5,17 @@
     visit '/good_job'
     expect(page).to have_content 'GoodJob 👍'
   end
+
+  # This requires storing job in database, but rest of the tests is using inline adapter
```
morgoth, line 9: When I add this to config/environments/test.rb the unit tests are failing
```
  config.active_job.queue_adapter = :good_job
  config.good_job = {execution_mode: :external}
```

engine/app/controllers/good_job/active_jobs_controller.rb
bensheldon: When an ActiveJob is retried, it generates a new GoodJob resource. That GJ resource has a new primary ID, but the ActiveJob ID is the same. In other words, a single ActiveJob job's history is represented across multiple GoodJob job entries.

I think you have the implementation right here, by deleting all of the GoodJob resources that have the same ActiveJob ID.

...but it's not pretty. And when thinking about "retry" in the future, we have to make sure that it's operating on the latest/newest/head GoodJob resource for a particular ActiveJob job.

To put it another way, `GoodJob::Job` resources are "runs".

engine/app/views/layouts/good_job/base.html.erb
```diff
@@ -51,6 +51,15 @@
       </div>
     </div>
 
+    <% if notice %>
+      <div class="alert alert-success alert-dismissible fade show d-flex align-items-center offset-md-3 col-6" role="alert">
+        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-check-circle-fill flex-shrink-0 me-2" viewBox="0 0 16 16">
+          <path d="M16 8A8 8 0 1 1 0 8a8 8 0 0 1 16 0zm-3.97-3.03a.75.75 0 0 0-1.08.022L7.477 9.417 5.384 7.323a.75.75 0 0 0-1.06 1.06L6.97 11.03a.75.75 0 0 0 1.079-.02l3.992-4.99a.75.75 0 0 0-.01-1.05z" />
+        </svg>
```
bensheldon, line 58: Can these SVGs be extracted into a shared partial?

spec/system/dashboard_spec.rb
bensheldon: I will look into that.

engine/app/controllers/good_job/active_jobs_controller.rb
bensheldon: This also makes me wonder about the safety of having the "delete" on the GJ Job, rather than on the Show page itself to represent that it's deleting _all_ the runs. But this also doesn't have to be perfect.

I think the worry I have is that someone will delete a job that has another run that is currently executing or waiting to execute.

Generative/blue sky idea: delete could be wrapped with an advisory lock on all the matching records.

bensheldon (commented):
This is fantastic! Thank you.

It does uncover some conceptual issues of how GoodJob jobs related to ActiveJob jobs, but no time like the present to figure it out 😄

engine/app/views/layouts/good_job/base.html.erb
morgoth: do you mean separate partial for each SVG?

engine/app/controllers/good_job/active_jobs_controller.rb
morgoth: Ah right. We have `GoodJob.preserve_job_records = :on_unhandled_error` so I never seen this.

So I think that for deleting we should use record id to delete only given record, as I believe this is what would be expected, by the principle of least surprise. WDYT?

spec/system/dashboard_spec.rb
bensheldon: @morgoth You can add this:

```
 ActiveJob::Base.queue_adapter = GoodJob::Adapter.new(execution_mode: :external)
 ```
 
The [adapter is reset](https://github.com/bensheldon/good_job/blob/b3dd40dd90a2cc3d8b20c9c40662f9c038b0a353/spec/support/reset_rails_queue_adapter.rb#L1-L7) after every test, so that's should be all you need:

engine/app/controllers/good_job/active_jobs_controller.rb
morgoth: I changed it, to remove only one, given job.
I believe to be more REST-ful we could change the show action to be actually the index one with applied filter.

bensheldon: Thanks, I think that's least surprise to only remove the one job 👍 

I want to suggest moving the destroy action into `good_jobs#destroy` because it is acting upon a good_job record. It could be clearer by having a `GoodJobJobsController` and an `ActiveJobJobsController` maybe?

I want to be fairly permissive at this stage because I'm learning both the functionality that people want, and it's also helpful to think through these things with you on actual PRs 😄

morgoth: That makes sense to not mix both, however what do you think about:
>  I believe to be more REST-ful we could change the show action to be actually the index one with applied filter.

This way all the actions fit into one controller.
Also currently the "show" action can display many jobs, which might be confusing.

bensheldon: In my mind, the `active_jobs#show` controller is still displaying a single ActiveJob Job, of which the details are multiple GoodJob Jobs.

I am imagining that this page would, in the future, also have some roll-up statistics (when the ActiveJob Job was initially enqueued, whether it has finished running, total number of retries, etc.).

As an end-user, the ActiveJob job is _the thing_ I imagine most users are conceptualizing and the GoodJob Job is the a "run" or "attempt". Unfortunately there isn't a single database record that models the ActiveJob Job yet for the purpose of display.

morgoth: Thanks for explanation. I moved it out to separate controller

bensheldon (approved):
👍

bensheldon:
I just released this as `v1.9.6`.

PR #268, opened 2021-06-08
Test run for PG 9.4 compatibility

gadimbaylisahil:
Documentation mentions compatibility with Postgres 9.6+. 

Only thing may be missing in Postgres 9.4 vs 9.6 seems like: https://www.postgresql.org/about/featurematrix/detail/276/

However, not sure if it's relevant here, so, trying to run the Github CI with 9.4


gadimbaylisahil:
@bensheldon Seems like all is fine with 9.4. Should we update docs?

gadimbaylisahil:
Will look into tomorrow again

bensheldon:
Nuts. One other thing I forgot to check: it looks like both PG 9.4 and 9.5 have been end-of-lifed:

https://www.postgresql.org/support/versioning/

I think it's fine to indicate that GoodJob is compatible with them currently... but I don't intend to maintain compatibility for EOLed dependencies. (It probably would be good to document this 😀)

gadimbaylisahil:
> Nuts. One other thing I forgot to check: it looks like both PG 9.4 and 9.5 have been end-of-lifed:
> 
> https://www.postgresql.org/support/versioning/
> 
> I think it's fine to indicate that GoodJob is compatible with them currently... but I don't intend to maintain compatibility for EOLed dependencies. (It probably would be good to document this grinning)

Thanks, Ben! 

Yes, it's EOF. I have no intention to bring this in to GoodJob documentation actually. Just stuck with 9.4 for some of my projects, can't update. So, was interested to try it out here in CI

bensheldon:
@gadimbaylisahil is it ok to close this PR?

Issue #206, opened 2021-01-24
Custom advisory locks to prevent certain jobs from being worked on concurrently?


reczy:
Hi Ben,

First of all, thanks so much for your hard work on this gem. I've been following the project since it showed up on HN a few months ago and recently started porting over my background jobs from Resque.

I'm submitting this issue to see if you'd be open to custom advisory locks? In my case, the purpose would be to prevent certain jobs (across multiple job classes) that work on data for a particular tenant from being worked on concurrently. In other words - perform only one job at a time for a particular account/organization, based on the job arguments.

As a workaround for now, I am grabbing a second advisory lock in around_perform that's based on a particular job argument, but this approach is obviously inefficient and fails to take advantage of your nice code already in Lockable.

Currently, you use part of the md5 hash of (1) the good_job table name and (2) the job primary key to determine the advisory lock for a particular job. Ideally, I'd have the opportunity to specify the text your Lockable module feeds to md5. In my opinion, doing it this way (rather than allowing the user to just specify the final 64-bit integer or 2 32-bit integers) seems like a decent trade-off between control and complexity. I suppose that just being able to specify (2) would be fine as well since the table_name doesn't change.

I've tried to think through the easiest way of accomplishing this. One option is to allow a job method that would return a string to be fed to md5 (if not present, then fallback to the current way of doing things). 

PSUEDO CODE:

```
# jobs/specific_job.rb
class SpecificJob < ApplicationJob
  
  # args would be job arguments
  self.good_job_lock_text(*args)
    "TENANT_LOCK_#{args.first}"
  end

  # ...
end
```

Then in GoodJob::Lockable

```
def hashable_lock_text
  <user defined text from good_job_lock_text> or <table_name + primary_key>
end
```

and elsewhere in Lockable 
```md5(#{hashable_lock_text})```

---

So anyway, I'd love to hear your thoughts on whether this is something you'd consider for this gem. I have no attachment to the specific design suggestion above - just trying to spark ideas. It would be great to be able to rely on the code already in Lockable to limit job concurrency (and avoid duplicating db locking). 

Thanks!

bensheldon:
@reczy thanks for proposing this idea! I think it's possible to extend GoodJob / Lockable, but with some tweaks.

I think it would be fairly simple to have Lockable have a configurable SQL lock string that gets MD5ed. It would need to be done entirely in SQL though, meaning it would probably end up looking like `md5(COALESCE(custom_column, id))...`.

[...]

PR #273, opened 2021-06-22
Allow Lockable to be passed custom column, key, and Postgres advisory lock/unlock function

bensheldon:
- Eventual intention is working towards locking on the ActiveJob ID rather than the GoodJob Job ID.
- Also supports work of #206.

### DIFF
--- a/lib/good_job/lockable.rb
+++ b/lib/good_job/lockable.rb
@@ -22,17 +22,25 @@ module GoodJob
     RecordAlreadyAdvisoryLockedError = Class.new(StandardError)
 
     included do
+      # Default column to be used when creating Advisory Locks
+      cattr_accessor(:advisory_lockable_column, instance_accessor: false) { primary_key }
+
+      # Default Postgres function to be used for Advisory Locks
+      cattr_accessor(:advisory_lockable_function) { "pg_try_advisory_lock" }
+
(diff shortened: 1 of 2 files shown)


reczy:
Hey @bensheldon - thoughts on using a single multi-line string in the `:advisory_lock` scope rather than a series of arel commands? My opinion is that reading the sql in its entirety makes it a bit easier to grasp what's going on here.

bensheldon:
@reczy I agree that using Arel has the regex problem of "now you have two problems". I have a few reasons to use Arel:

- I'm personally leaning into Rails and ActiveRecord, which heavily involves Arel.
- I'd like to extract the Lockable module into its own gem, and I'd like to retain some flexibility in what it can do and expect and decompose.

...that being written, I also don't want Arel to be a blocker for people contributing.

The sql query is documented in the tests: https://github.com/bensheldon/good_job/blob/e83d27448abf54195fa09eb03a872eefb39b9a03/spec/lib/good_job/lockable_spec.rb#L21-L41

...but I can better document what's happening in the code itself too. 

Also, thank you for commenting on this PR. I was a little premature in opening it (I gotta get better at "Draft" PRs)

bensheldon:
I extracted the `FOR UPDATE SKIP LOCKED` change to #274.

PR #281, opened 2021-06-24
Add concurrency extension for ActiveJob

### LINKED ISSUE #206: Custom advisory locks to prevent certain jobs from being worked on concurrently?

bensheldon:
Creates`GoodJob::ActiveJobExtensions::Concurrency` module that can be included into ActiveJob classes.

When enqueuing jobs and concurrency is reached/exceeded, jobs are not enqueued and the `MyJob.perform_later` returns `false` using native callback behavior of `throw :abort`

When performing jobs and concurrency is reached/exceeded, jobs are retried using an incremental backoff, using native ActiveJob `retry_on ... attempts: Float::INFINITY, wait: :exponentially_longer` functionality.

Closes #206.

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -1,0 +1,68 @@
+module GoodJob
+  module ActiveJobExtensions
+    module Concurrency
+      extend ActiveSupport::Concern
+
+      ConcurrencyExceededError = Class.new(StandardError)
+
+      included do
+        class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
+
+        before_enqueue do |job|
+          # Always allow jobs to be retried because the current job's execution will complete momentarily
+          next if CurrentExecution.active_job_id == job.job_id
+
+          limit = job.class.good_job_concurrency_config.fetch(:enqueue_limit, Float::INFINITY)
+          next if limit.blank? || (0...Float::INFINITY).exclude?(limit)
+
+          key = job.good_job_concurrency_key
+          next if key.blank?
+
+          GoodJob::Job.new.with_advisory_lock(key: key, function: "pg_advisory_lock") do
+            # TODO: Why is `unscoped` necessary? Nested scope is bleeding into subsequent query?
+            enqueue_concurrency = GoodJob::Job.unscoped.where(concurrency_key: key).unfinished.count
+            # The job has not yet been enqueued, so check if adding it will go over the limit
+            throw :abort if enqueue_concurrency + 1 > limit
+          end
+        end
+
+        retry_on(
+          GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+          attempts: Float::INFINITY,
+          wait: :exponentially_longer
+        )
+
+        before_perform do |job|
+          limit = job.class.good_job_concurrency_config.fetch(:perform_limit, Float::INFINITY)
+          next if limit.blank? || (0...Float::INFINITY).exclude?(limit)
+
+          key = job.good_job_concurrency_key
+          next if key.blank?
+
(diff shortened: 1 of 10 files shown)


lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -0,0 +1,41 @@
...
+        def good_job_enqueue_exclusively_with(limit: 1)
+          before_enqueue do |job|
+            next unless job.respond_to? :good_job_concurrency_key
+            next if $ERROR_INFO.is_a? GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError
+
+            GoodJob::Job.new.with_advisory_lock(key: job.good_job_concurrency_key, function: "pg_advisory_lock") do
+              # TODO: Why is `unscoped` necessary? Nested scope is bleeding into subsequent query?
+              enqueue_concurrency = GoodJob::Job.unscoped.where(concurrency_key: job.good_job_concurrency_key).unfinished.count
+              # The job has not yet been enqueued, so check if adding it will go over the limit
+              throw :abort if enqueue_concurrency + 1 > limit
```
reczy, line 22: Does this create a race condition that needs to be accounted for? Say the enqueue_concurrency limit is 1, and the actual enqueue concurrency goes from 1 to 0 between the above two lines of code.

bensheldon: I think the race condition is addresses because this is wrapped with a `pg_advisory_lock` on the key, which should function as a database-wide mutex. 

But It's also gonna introduce latency because the PG connection will wait until it can acquire an advisory lock before running the code. If done on the webserver when under heavy load and there's a database statement timeout... that could be bad. I _hope_ that jobs that are likely to need enqueue limits will be triggered from a backend process (rake task, upcoming cron feature, etc).

reczy:
@bensheldon thank you for this!

PR #294, opened 2021-07-09
Wrap and truncate error message, which can be a huge text

morgoth:
In addition I'm thinking about dedicated show page, that will display single record with all the details, as with this change it is not possible to see the full error.


bensheldon:
@morgoth The CI failure is caused by a change in Rails HEAD. I fixed it (https://github.com/bensheldon/good_job/pull/299/files#diff-ba08e71d597ff998a90c368fd87831444de2625b01a474331cff63482dc420cfR64-R70) and you should be good to rebase this.

bensheldon:
@morgoth I was able to rebase it myself 💪

Issue #298, opened 2021-07-19
Add Frozen String Literal to all files


bensheldon:
Something overlooked in the initial project set up.

- [x] Update Rubocop to remove disabling of `Style/FrozenStringLiteralComment` rule (not sure why that was initially there; probably merely copy-pasta from some other project)
- [x] Run Rubocop to add the literal to all the files

tedhexaflow:
https://github.com/bensheldon/good_job/pull/302

bensheldon:
@tedhexaflow Thank you 🙏

PR #302, opened 2021-07-20
Add Frozen String Literal to all files

tedhexaflow:
Ref: https://github.com/bensheldon/good_job/issues/298


tedhexaflow:
Hi @bensheldon,

Please note, after removing the Style/FrozenStringLiteralComment, robocop raises another cop `Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect`.

I also fixed that cop in this PR.

Issue #325, opened 2021-08-10
Race conditions in ActiveJob concurrency extension


codyrobbins:
I’ve been testing out the brand new concurrency controls for use in a production system and unfortunately in practice they don’t appear to be working correctly. There seem to be two separate issues with the controls around enqueue and perform, but they both boil down to race conditions. I’ve created a pull request that I believe addresses the issues that we are seeing in our particular case, but I’m not familiar enough with the GoodJob or ActiveJob internals to be sure that there are not other problems I’m unaware of introduced by what I’ve done here.

In case it makes a difference, I’ve been running into these particular issues with the following concurrency limits:

```
good_job_control_concurrency_with(
  perform_limit: 1,
  enqueue_limit: 2,
  key: […]
)
```

The goal here is to ensure that the job is never being performed more than once at the same time, but we can enqueue two additional jobs to act as a debounce rather than a throttle on job performance.

For enqueue, it appears as though the problem is that the advisory lock is released in the `before_enqueue` callback before the job actually gets enqueued, and this is allowing multiple jobs to be enqueued simultaneously with the advisory lock essentially being useless.

https://github.com/bensheldon/good_job/blob/4833fb75f196e6a70e0d2bc20a1f0209d9f3496d/lib/good_job/active_job_extensions/concurrency.rb#L22-L27

Immediately after the advisory lock is relinquished at the end of this block, but before control is returned to the code that goes ahead and actually does the enqueueing, another job can enter the critical section and can calculate an invalid `enqueue_concurrency` because it won’t take into account the job that has just been allowed to enqueue but has not actually done so yet.

I’m guessing there are other internal GoodJob or ActiveJob callbacks that need to run after this code but before the enqueue occurs which exacerbate this problem and make it more likely to occur, although even if there aren’t I think instructions can still be interleaved in such a way that this happens. In practice it seems to end up happening a significant proportion of the time for me and with the above concurrency configuration I was getting more than two jobs enqueued at the same time consistently.

The simple solution I came up with that seems to resolve this problem is changing the before callback to an around one. This ensures the advisory lock is not released until the enqueue operation has actually occurred and that no other jobs can enter the critical section in the meantime.

For perform, the problem we’re seeing is a huge amount of lock contention that in practice introduces heavy unnecessary delays to queue processing but in certain circumstances seems to prevent the jobs from running indefinitely.

When a job begins to be performed, its row is locked here:

[...]

PR #326, opened 2021-08-10
Fixes for race conditions in ActiveJob concurrency extension

codyrobbins:
Fixes for #325.

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -9,21 +9,21 @@ module GoodJob
       included do
         class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
 
-        before_enqueue do |job|
+        around_enqueue do |job, block|
           # Always allow jobs to be retried because the current job's execution will complete momentarily
-          next if CurrentExecution.active_job_id == job.job_id
+          next(block.call) if CurrentExecution.active_job_id == job.job_id
 
           limit = job.class.good_job_concurrency_config.fetch(:enqueue_limit, Float::INFINITY)
-          next if limit.blank? || (0...Float::INFINITY).exclude?(limit)
+          next(block.call) if limit.blank? || (0...Float::INFINITY).exclude?(limit)
 
           key = job.good_job_concurrency_key
-          next if key.blank?
+          next(block.call) if key.blank?
 
           GoodJob::Job.new.with_advisory_lock(key: key, function: "pg_advisory_lock") do
             # TODO: Why is `unscoped` necessary? Nested scope is bleeding into subsequent query?
             enqueue_concurrency = GoodJob::Job.unscoped.where(concurrency_key: key).unfinished.count
             # The job has not yet been enqueued, so check if adding it will go over the limit
-            throw :abort if enqueue_concurrency + 1 > limit
+            block.call unless enqueue_concurrency + 1 > limit
           end
         end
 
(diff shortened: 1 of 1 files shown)


lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -41,9 +41,9 @@ module Concurrency
           next if key.blank?
 
           GoodJob::Job.new.with_advisory_lock(key: key, function: "pg_advisory_lock") do
-            perform_concurrency = GoodJob::Job.unscoped.where(concurrency_key: key).advisory_locked.count
+            allowed_jobs = GoodJob::Job.unscoped.where(concurrency_key: key).advisory_locked.order(performed_at: :asc).limit(limit).collect(&:active_job_id)
```
bensheldon, line 44: This is genius! I have some functional and stylistic suggestions:

```suggestion
            allowed_active_job_ids = GoodJob::Job.unscoped.where(concurrency_key: key).advisory_locked.order(Arel.sql "COALESCE(performed_at, scheduled_at, created_at) ASC").limit(limit).pluck(&:active_job_id)
```

- `performed_at` is only present if `GoodJob.preserve_job_records` is true so I want to suggest the COALESCE to ensure that there is always a deterministic value and not null/nil: https://github.com/bensheldon/good_job/blob/9c2b9f74396881a2f22b5a38681aefabe4405e40/lib/good_job/job.rb#L259-L260
- using ActiveRecord's `pluck` for a more efficient query
- being longwinded in variable names to be clear if a variable holds a good_job_id or an active_job_id

lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -9,21 +9,21 @@ module Concurrency
       included do
         class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
 
-        before_enqueue do |job|
+        around_enqueue do |job, block|
           # Always allow jobs to be retried because the current job's execution will complete momentarily
-          next if CurrentExecution.active_job_id == job.job_id
+          block.call if CurrentExecution.active_job_id == job.job_id
```
bensheldon, line 14: For each of these `block.calls` I think they need to instead be `next(block.call)` to return early and ensure that `block.call` is not called multiple times.

bensheldon (changes requested):
This is awesome! I left some suggestions for improvement; the primary one issue is ensuring that `block.call` is not called multiple times.

lib/good_job/active_job_extensions/concurrency.rb
codyrobbins: Awesome, thank you!

I staunchly support overly descriptive variable names—and not thinking about `performed_at` being empty is one of those unknown unknowns I knew I was missing!

I didn’t go with `pluck` because I noticed there’s an accessor method for `Job#active_job_id` that falls back to `serialized_params['job_id']`. I’m guessing that might be for unsaved records, which isn’t relevant here, and so `pluck` is fine? Just want to double-check!

lib/good_job/active_job_extensions/concurrency.rb
codyrobbins: You are absolutely right and that went completely over my head. That’s a really good catch!

lib/good_job/active_job_extensions/concurrency.rb
bensheldon: There is a column for `active_job_id`... but it's not guaranteed to exist for a user if they haven't kept up with GoodJob migrations. I am comfortable leaving it this situation where the column may not exist and raise a database exception (SEMVER is a lot of work to get perfect).

But, if you wanted to go the extra mile, you could do a feature check and implement it both the more performant and less performant ways with a deprecation notice like this: https://github.com/bensheldon/good_job/blob/811fb264bd29626331caaa9609bf17151e9ca005/lib/good_job/job.rb#L210-L223

I think this could be a fairly hot code path, so the better query is important.

lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -41,9 +41,9 @@ module Concurrency
           next if key.blank?
 
           GoodJob::Job.new.with_advisory_lock(key: key, function: "pg_advisory_lock") do
-            perform_concurrency = GoodJob::Job.unscoped.where(concurrency_key: key).advisory_locked.count
+            allowed_active_job_ids = GoodJob::Job.unscoped.where(concurrency_key: key).advisory_locked.order(Arel.sql "COALESCE(performed_at, scheduled_at, created_at) ASC").limit(limit).pluck(:active_job_id)
```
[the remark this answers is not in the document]
codyrobbins: Fixed!

bensheldon (approved):
🙌  Thank you again 🎉

bensheldon:
Released in `v1.12.1`

PR #333, opened 2021-08-18
Don’t attempt to enforce concurrency limits with other queue adapters

codyrobbins:
One other minor issue I came across with the concurrency extension is that it is always being applied regardless of which queue adapter is actually handling the job. This became a problem when I got around to testing because in our particular test suite we’re using `ActiveJob::QueueAdapters::TestAdapter`. In this case jobs don’t end up being performed because there are no associated `GoodJob::Job` rows with advisory locks. This change skips enforcing the concurrency limits unless the job is being handled by the GoodJob adapter.

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -10,6 +10,9 @@ module GoodJob
         class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
 
         around_enqueue do |job, block|
+          # Don't attempt to enforce concurrency limits with other queue adapters.
+          next(block.call) unless job.class.queue_adapter.is_a?(GoodJob::Adapter)
+
           # Always allow jobs to be retried because the current job's execution will complete momentarily
           next(block.call) if CurrentExecution.active_job_id == job.job_id
 
@@ -34,6 +37,9 @@ module GoodJob
         )
 
         before_perform do |job|
+          # Don't attempt to enforce concurrency limits with other queue adapters.
+          next unless job.class.queue_adapter.is_a?(GoodJob::Adapter)
+
           limit = job.class.good_job_concurrency_config.fetch(:perform_limit, Float::INFINITY)
           next if limit.blank? || (0...Float::INFINITY).exclude?(limit)
 


codyrobbins:
Just noticed this causes a bunch of test failures so I’ll have to look into getting those passing tomorrow!

lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -19,6 +19,9 @@ module Concurrency
           key = job.good_job_concurrency_key
           next(block.call) if key.blank?
 
+          # Don't attempt to enforce concurrency limits with other queue adapters.
+          next(block.call) unless job.queue_adapter.is_a?(ActiveJob::QueueAdapters::GoodJobAdapter)
```
bensheldon, line 23: I think queue_adapter lives on the class rather than the instance:`job.class.queue_adapter`.

Also, I think this should be the first guard clause inside of the `around_enqueue` block.

lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -40,6 +43,9 @@ module Concurrency
           key = job.good_job_concurrency_key
           next if key.blank?
 
+          # Don't attempt to enforce concurrency limits with other queue adapters.
+          next unless job.queue_adapter.is_a?(ActiveJob::QueueAdapters::GoodJobAdapter)
```
bensheldon, line 47: I think this should be the first guard clause inside of the `before_perform` block.

bensheldon (commented):
Nice discovery! This is a great improvement with a little cleanup.

lib/good_job/active_job_extensions/concurrency.rb
codyrobbins: Thanks—I moved the guards up to be the first check!

It looks like `queue_adapter` on instances delegates to the class, which was working for me on Rails 6.1, but maybe that’s a recent addition—I’ll see if `job.class.queue_adapter` fixes those test failures.

bensheldon:
Released in [`v1.13.1`](https://github.com/bensheldon/good_job/releases/tag/v1.13.1)

Thank you! 🙏

PR #348, opened 2021-08-24
Suppress backtrace of ConcurrencyExceededError

### LINKED ISSUE #347: Is there any value in seeing a backtrace for ConcurrencyExceededError?
Hi @bensheldon!

I'm curious if you (or other users) find any value in seeing the full backtrace for concurrency errors? In my case, concurrency errors are expected and common so I don't really view them as "errors" to be further investigated. I am using the following to silence the backtraces:

```
# config/initializers/good_job.rb

module GoodJob
  module ActiveJobExtensions
    module Concurrency
      class ConcurrencyExceededError < StandardError
        def backtrace
          [] # suppress backtrace
        end
      end
    end
  end
end
```

Does it make sense to do any of the following?

- hardcode this in GoodJob, or
- include an option to accomplish this in GoodJob via config, or
- include some documentation mentioning that the above code (or something similar) can suppress the backtraces?

reczy:
Closes #347

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -4,7 +4,11 @@ module GoodJob
     module Concurrency
       extend ActiveSupport::Concern
 
-      ConcurrencyExceededError = Class.new(StandardError)
+      class ConcurrencyExceededError < StandardError
+        def backtrace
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
👍

bensheldon:
Released in `v2.0.1`

PR #356, opened 2021-08-30
Implement `GoodJob.cleanup_preserved_jobs`, fixes #351

### LINKED ISSUE #351: Expose CLI `cleanup_preserved_jobs` functionality via `GoodJob`?
Hey! I've just finished reading through the README and it pretty much answered all my questions. Good job (pun intended) on the documentation!

However, I was wondering if there was a way to officially expose a method to use in a rake task or job to clean up old jobs. The [README suggests](https://github.com/bensheldon/good_job/tree/9517669d77459b5fcd4fbdac91519ae7f9d3ce70#monitor-and-preserve-worked-jobs) using:

```Ruby
GoodJob::Job.finished(1.day.ago).delete_all
```

However, the CLI task does a bit more, namely instrumenting via `ActiveSupport::Notifications`:

https://github.com/bensheldon/good_job/blob/811fb264bd29626331caaa9609bf17151e9ca005/lib/good_job/cli.rb#L139-L146

Of course I can just copy the code, but since that functionality seems to be desired behavior since it's implemented in the CLI, maybe it's a good idea to expose this directly? Would also make it easier to schedule this via GoodJob itself.

Another question would be, is there a reason why `.delete_all` is enough and we don't need callbacks via `.destroy_all`? I assume there's just no need for them as the jobs have no associations to destroy or nullify, but just wanted to make sure.

Thanks!

aried3r:


lib/good_job.rb
```diff
@@ -114,6 +115,26 @@ def self._shutdown_all(executables, method_name = :shutdown, timeout: -1)
...
+  # Deletes preserved job records.
+  # By default, GoodJob deletes job records when the job is performed and this
+  # method is not necessary. However, when `GoodJob.preserve_job_records = true`,
+  # the jobs will be preserved in the database. This is useful when wanting to
+  # analyze or inspect job performance.
+  # If you are preserving job records this way, use this method regularly to
+  # delete old records and preserve space in your database.
+  # @params before_seconds_ago [nil,Numeric] Jobs olders than this will be deleted (default: +86400+).
+  # @return [Integer] Number of jobs that were deleted.
+  def self.cleanup_preserved_jobs(before_seconds_ago: nil)
```
bensheldon, line 127: I'm curious if you think it would be good to have a better kwarg name here. I want to suggest something like 

- `cleanup_preserved_jobs(before: 3.days.ago)` or
- `cleanup_preserved_jobs(older_than: 3.days)` <-- I think I like this one.
- Something better?

The longer option names were really a result of wanting the CLI interface to be clear given that there are limitations in what can be passed through `ENV`. But as a Ruby method, it's easy to pass in a timestamp, or duration. 

It's a kwarg so it's trivial to preserve both options, and there is a benefit to being symetrical with the CLI options, but I recognize it's a little clunky.

Not a blocker of this PR, I wanted to get your perspective :D

aried3r: > I'm curious if you think it would be good to have a better kwarg name here.

Definitely!

When writing this, I was also considering `older_than` but wanted to stick with the CLI option name to avoid having to discuss a potential other name. So I'm very happy you brought it up!

I don't think we necessarily need to keep both names, but I'm open to that. I'll work on the changes and you can have another look if you want to keep this symmetrical after all.

README.md
```diff
@@ -724,7 +724,8 @@ It is also necessary to delete these preserved jobs from the database after a ce
 - For example, in a Rake task:
 
     ```ruby
-    GoodJob::Job.finished(1.day.ago).delete_all
+    GoodJob.cleanup_preserved_jobs # Will keep 1 day of job records by default.
+    GoodJob.cleanup_preserved_jobs(before_seconds_ago: 7.days.seconds) # It also takes custom arguments.
```
bensheldon, line 728: This documentation should also be updated:

```suggestion
    GoodJob.cleanup_preserved_jobs(older_than: 7.days) # It also takes custom arguments.
```

bensheldon:
@aried3r Thank you! 🎉  Released in`v2.0.3`.

Issue #359, opened 2021-09-01
Cron-like jobs not always executed, possible reasons?


aried3r:
I'm not sure if this is perhaps just a configuration error, but we're seeing some issues with cron-like jobs not always being executed.

We're still experimenting a bit with GoodJob, so configuration is still changing and not final. Currently we have it all in an initializer because we don't want to have configuration in multiple places for now.

Regular jobs using `.perform_later` work as expected from what we can see, but we're having issues with some cron-like jobs.

As of now have _very mild_ workloads regarding jobs. Perhaps a couple hundred a day, less than 500 even. This will increase as we add more background jobs in our app. And, as you'll see in our configuration below, 2 cron jobs that execute a total of 6 times per week.
Could this perhaps have anything to do with when GoodJob "wakes up", to handle cron-like jobs? I'm not exactly sure how GoodJob works under the hood for these sort of jobs, so our best guess was that we don't have enough workload for GoodJob to "wake up" more often and thus miss the scheduled time.

Could you maybe shed some light on how it works internally in `:async` mode or in general if there is no difference? We'd love to know when and what is triggering the `GoodJob started cron with 2 jobs.` log messages. I've found the [source](https://github.com/bensheldon/good_job/blob/d68de094c506172d07d2352327998f85d5208dad/lib/good_job/cron_manager.rb#L43-L49) but haven't yet tracked down when exactly that is being executed.

Our configuration looks like this:

```Ruby
# config/initializers/good_job.rb
# frozen_string_literal: true

Rails.application.configure do
  config.active_job.queue_adapter = :good_job

  config.good_job.execution_mode = :async
  config.good_job.max_threads = 5
  # Disable polling in development.
  config.good_job.poll_interval = Rails.env.development? ? -1 : 10 # seconds
  config.good_job.shutdown_timeout 60 # seconds
  config.good_job.enable_cron = true
  config.good_job.cron = {
      reminders: { cron: '0 9 * * MON-FRI', class: 'ReminderNotifierJob' },
      clean_up_preserved_jobs: { cron: '0 19 * * SUN', class: 'CleanUpPreservedJobsJob' }
  }
  config.good_job.queues = '*'
end

GoodJob.active_record_parent_class = 'ApplicationRecord'
GoodJob.logger = Logger.new(Rails.root.join('log', 'good_job.log'))

# Useful for analyzing the jobs in the dashboard.
GoodJob.preserve_job_records = true

GoodJob.retry_on_unhandled_error = false

[...]

PR #367, opened 2021-09-08
Add dashboard for cron-style jobs

aried3r:
Hey!

Since I'm still working on #359, I thought I could implement a very basic dashboard for cron-style jobs. Looks like this:

Screenshot 2021-09-08 at 14-46-04 Good Job Dashboard

`args` and `set` are not being displayed since they could be procs.

What do you think?

### DIFF
--- a/engine/app/controllers/good_job/dashboards_controller.rb
+++ b/engine/app/controllers/good_job/dashboards_controller.rb
@@ -102,10 +102,5 @@ module GoodJob
         end,
       }
     end
-
-    def cron_jobs
-      configuration = GoodJob::Configuration.new({})
-      @cron_jobs = configuration.cron
-    end
   end
 end
--- a/engine/app/controllers/good_job/dashboards_controller.rb
+++ b/engine/app/controllers/good_job/dashboards_controller.rb
@@ -102,5 +102,10 @@ module GoodJob
         end,
       }
     end
+
+    def cron_jobs
+      configuration = GoodJob::Configuration.new({})
+      @cron_jobs = configuration.cron
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


aried3r:
Hmm, the failed test seems like a fluke. I don't think this change should have made it time out.

engine/app/views/layouts/good_job/base.html.erb
```diff
@@ -23,9 +23,15 @@
       <div class="collapse navbar-collapse" id="navbarSupportedContent">
         <ul class="navbar-nav me-auto">
           <li class="nav-item">
-            <%= link_to root_path, class: ["nav-link", ("active" if current_page?(root_path))] do %>
-              All jobs <span class="badge bg-secondary">More views coming soon</span>
-            <% end %>
+            <%= link_to "All Jobs", root_path, class: ["nav-link", ("active" if current_page?(root_path))] %>
+          </li>
+          <li class="nav-item">
+            <%= link_to "Cron-style Jobs", cron_jobs_path, class: ["nav-link", ("active" if current_page?(root_path))] %>
```
bensheldon, line 29: Maybe "Cron Schedules"

engine/app/controllers/good_job/dashboards_controller.rb
```diff
@@ -102,5 +102,10 @@ def index
         end,
       }
     end
+
+    def cron_jobs
```
bensheldon, line 106: Ideally this would be in its own controller: `CronSchedulesController`.

bensheldon (approved):
This is awesome! Thank you.  I'm happy to merge this in right now...and I also left two comments that you're welcome to do now, or we can defer (I don't want perfection to be the enemy of good).

aried3r:
Good suggestions! I didn't know which style you'd prefer, so I didn't create too many new files. I'll try to apply this changes either today or tomorrow.

bensheldon:
@aried3r Thank you so much for this! 🙌

PR #377, opened 2021-09-13
Rename `GoodJob::Job` to `GoodJob::Execution`

### LINKED ISSUE #376: Rename `GoodJob::Job` to be `GoodJob::Execution`
A source of conceptual complexity is that a `GoodJob::Job` does not directly map to an ActiveJob Job. A `GoodJob::Job` is a single run or "execution" of an ActiveJob job, which if retried, will have multiple entries in the `good_jobs` database table. This is very evident in places in which local variables are named `active_jobs` and `good_jobs` in order to clearly differentiate
 
By renaming the model to be `Execution` I think it would help clear up some confusion. I also think it could help accelerate development of the Dashboard because I'm a bit fuzzy in my head about how to present the data. Being clear about something being an (ActiveJob) Job, and a (record in the `good_jobs` table) Execution would help.

- I'm open to alternative naming suggestions
- I won't be renaming the underlying database table, that will continue to be `good_jobs` (renaming tables is operationally messy so I'd like to avoid that)
- I think it should be a straightforward renaming of classes/variables and no other issues. The configuration that is part of the public API is non-specific afaik.
- This would be annoying for anyone who is using `GoodJob::Job` in their own code, but I don't consider that part of the public API. Maybe I'll find out.

bensheldon:
Closes #376.

- Preserves but deprecates `GoodJob::Job`.
- Also renames `GoodJob::CurrentExecution` to `GoodJob::CurrentThread`
- Updates Dashboard controllers and views/partials; puts the "Active Job ID" column first in the table view; wraps UUIDs with `<code>` tags
- Renames scopes`GoodJob::Execution.with_job_class(class_name)` to `GoodJob::Execution.job_class(class_name)`


morgoth:
I'm wondering how many people are using internals of the gem.
In example we have some hand crafted simple monitoring tool in place to check for jobs, so we're using `GoodJob::Job` directly, but we also have tests for it so we will notice the change.

The other example is the hirefire gem (for autoscalling heroku dynos) that uses the class directly https://github.com/hirefire/hirefire-resource/blob/master/lib/hirefire/macro/good_job.rb

Do you plan to release it as 2.2 version? Maybe we should deprecate the `Job` somehow first?

bensheldon:
@morgoth thanks for drawing my attention to that. You're right, there should be a deprecation notice on this, and probably preserve the presence of the `GoodJob::Job` class until v3.

bensheldon:
I'm now preserving/deprecating `GoodJob::Job` instead of removing it (`class Job < Execution`).

morgoth (approved):

PR #379, opened 2021-09-15
Link to execution id, not active_job_id

nnc:
Hi @bensheldon,

Thanks for this great project!

After upgrading from `2.1.0` to `2.2.0` I started seeing this error when accessing the dashboard:
```
No route matches {:action=>"show", :controller=>"good_job/active_jobs", :id=>nil}, possible unmatched constraints: [:id]

Backtrace:
good_job (2.2.0) engine/app/views/good_job/shared/_executions_table.erb:31
good_job (2.2.0) engine/app/views/good_job/shared/_executions_table.erb:23
good_job (2.2.0) engine/app/views/good_job/dashboards/index.html.erb:41
```

No idea if this is actually correct, but it did fix the error for me.

### DIFF
--- a/engine/app/views/good_job/shared/_executions_table.erb
+++ b/engine/app/views/good_job/shared/_executions_table.erb
@@ -28,7 +28,7 @@
               <% end %>
             </td>
             <td>
-              <%= link_to active_job_path(execution.active_job_id, anchor: dom_id(execution)) do %>
+              <%= link_to active_job_path(execution.id, anchor: dom_id(execution)) do %>
                 <code><%= execution.id %></code>
               <% end %>
             </td>


bensheldon:
@nnc thanks for reporting this! hmm, that shouldn't be broken, and `active_job_id` is the correct accessor for that path.  

I think what you're seeing is the result of having a null `active_job_id` column value, which shouldn't exist after the version `2.0` upgrade.

Could you share the results of this? 

```ruby
total_executions = GoodJob::Execution.count
executions_without_active_job_id = GoodJob::Execution.where(active_job_id: nil).count

puts "total: #{total_executions}, without_active_job_id: #{executions_without_active_job_id}"
```

If you do see executions without an `active_job_id` value, you can update it by running this:

```ruby
UPDATE_BATCH_SIZE = 500
loop do
  break if GoodJob::Execution.where(active_job_id: nil, finished_at: nil).limit(UPDATE_BATCH_SIZE).update_all("active_job_id = (serialized_params->>'job_id')::uuid").zero?
end
```

nnc:
Thanks for explaining that @bensheldon. And yes, you're right.

I think the main reason for this problem is that I have `GoodJob.preserve_job_records = true`, so all those finished executions were causing this error, even after running the code you so helpfully provided.

Now that I ran `good_job cleanup_preserved_jobs --before-seconds-ago=1` there's no error anymore.

PR #381, opened 2021-09-17
Preserve page filter when deleting execution

morgoth:
Example: when deleting jobs one by one with error state it is now easier as we redirect back to the same page


bensheldon (approved):
👍

PR #387, opened 2021-09-25
Update GH Test Matrix with latest JRuby 9.3.0.0

tedhexaflow:


tedhexaflow:
Hi Ben, 

Shall we also keep the latest version of JRuby 9.2? (9.2.17.0)

bensheldon:
@tedhexaflow Good question, my understanding of JRuby has changed. Now that I understand JRuby a little better, I think we should track `9.2.latest` and `9.3.latest`.

For MRI Ruby, my policy is to keep the `2.6.latest`, `2.7.latest`, `3.0.latest`.


My mistake earlier was that I believed that there were functional differences between `9.2.13.0` and `9.2.16.0`. But now I see that the entire `JRuby 9.2` series was intended to be equivalent to `MRI 2.5`, and the `JRuby 9.3` series is intended to be compatible with `MRI 2.6`.

tedhexaflow:
Alrighty! 
I'll remove 9.2.13.0 and keep the latest version of 9.2 and 9.3.
For MRI Ruby, we can use 2.6, 2.7 in the matrix, and it will use the latest version. 
But I'm not sure we can do the same for JRuby. Let me have a look.

tedhexaflow:
Hi @bensheldon 
Cool. It worked. 
You can merge the PR 😀
Thanks 🙏

bensheldon:
Thank you! 🙌

Issue #388, opened 2021-09-29
"WARNING: you don't own a lock of type ExclusiveLock" in Development


newbenji:
First of all great job. 
Working on migration some jobs to good_job.
but i get this warning out.

WARNING:  you don't own a lock of type ExclusiveLock

But i cant find documentation to support it, or im not good at finding it

bensheldon:
@newbenji hmm, that's unexpected for you to receive that warning:

```
WARNING: you don't own a lock of type ExclusiveLock
```

The warning is output by the database adapter. It means that the current database connection (thread) is trying to unlock an Advisory Lock that does not exist. 

Can you say more about where you're seeing this?

bensheldon:
Are you seeing this in your development environment, or in production?

Could you try adding this to an initializer (`config/initializers/good_job.rb`):

```ruby
# Output Postgres notifications through Rails.logger instead of stderr
ActiveSupport.on_load :active_record do
  ActiveRecord::ConnectionAdapters::AbstractAdapter.set_callback :checkout, :before, lambda { |conn|
    raw_connection = conn.raw_connection
    next unless raw_connection.respond_to? :set_notice_receiver

    raw_connection.set_notice_receiver do |result|
      Rails.logger.info(result.error_message)
    end
  }
end
```

...and then share the logs that are produced (the initializer above will make sure that the Postgres warning shows up in the logs, rather than going to stderr).

Also, I wrote out some of these hints too for some lock causing problems I've run across: https://github.com/bensheldon/good_job/issues/329#issuecomment-898087131

mchristen:
I see this in my development environment! It seems to happen around the times I change code and the code reloading takes place.

bensheldon:
Thank you @mchristen! I was able to reproduce it in development with code reloading. When a Rails reloads the code, the database connection that is used to take the initial advisory lock is not the same connection used to release the lock. I'm only seeing this when code reloading is enabled , not in a Production configuration (fhew!)

I will put together a fix for this.

bensheldon:
I have a fix for this released in https://github.com/bensheldon/good_job/releases/tag/v2.3.1

newbenji:
thx awsome and fast response. i must aggree i only saw the error on development

mikehale:
I'm seeing this when running in development mode and using the external execution mode.

bensheldon:
[...]

PR #389, opened 2021-09-30
Wrap Scheduler task execution with Rails `reloader` instead of `executor` to avoid database connection changing during code reload

bensheldon:
Connects to #388. 

Related to #95, and _hopefully_ this PR does not produce deadlocks.

### DIFF
--- a/lib/good_job/scheduler.rb
+++ b/lib/good_job/scheduler.rb
@@ -230,7 +230,7 @@ module GoodJob # :nodoc:
     # @return [void]
     def create_task(delay = 0)
       future = Concurrent::ScheduledTask.new(delay, args: [performer], executor: executor, timer_set: timer_set) do |thr_performer|
-        Rails.application.executor.wrap do
+        Rails.application.reloader.wrap do
           thr_performer.next
         end
       end


jjb:
@bensheldon

I was recently reading https://guides.rubyonrails.org/threading_and_code_execution.html#reloader

and came across  "Rails automatically wraps web requests and Active Job workers"

what do you know about how the need for this interacts with that alleged behavior?

I wonder if that statement only applies to the shipped async adapter and whoever wrote the docs wasn't precise?

bensheldon:
@jjb great question! I think the documentation is written correctly, but GoodJob is a special case because GoodJob uses ActiveRecord to fetch-and-lock the serialized job data _before_  it invokes ActiveJob. GoodJob has to wrap its code with the Rails Executor in order for _GoodJob_ to work correctly, and then when ActiveJob is invoked, ActiveJob's usage of Rails Executor is a no-op.  E.g. in psuedo code:

```Ruby
create_thread do
  Rails.application.executor.wrap do
    GoodJob::Execution.fetch_and_lock_job do |job_params|
      ActiveJob.execute(job_params) do |job|
        Rails.application.executor.wrap do # noop
          job.perform
        end
      end
    end
  end
end
```

I think the Rails docs are assuming that the Thread is being invoked directly with the serialized job params primitives e.g.

```ruby

Thread.new(job_params) do |thr_job_params|
  ActiveJob.execute(thr_job_params) do |job|
    Rails.application.executor.wrap do # noop
      job.perform
    end
  end
end
```

Issue #382, opened 2021-09-17
Fix flakey test that times out


bensheldon:
This test frequently exceeds the 15min timeout on Github Actions: 

https://github.com/bensheldon/good_job/blob/4d0fab3c4adf54f62495649e2efee903fdd5fcfd/spec/integration/server_spec.rb#L20-L34

I don't _think_ it's a bisectable problem, but is something inherent in the test itself (but I could be wrong). The test itself shells out and starts a separate Rails process, and then waits for specific log output. There are a few similar tests, but this one alone ("production async") seems to be be the one timing out.

Examples:
- https://github.com/bensheldon/good_job/runs/3323594460?check_suite_focus=true#step:9:1039
- https://github.com/bensheldon/good_job/runs/3247004793?check_suite_focus=true#step:9:741
- https://github.com/bensheldon/good_job/runs/3323594460?check_suite_focus=true#step:9:1039

PR #391, opened 2021-10-01
Display cron schedules properties in dashboard

aried3r:
I saw this in [sidekiq-scheduler](https://github.com/ondrejbartas/sidekiq-cron#web-ui-for-cron-jobs) and thought this is actually quite useful for us at work.

Since the README calls `:set` additional properties, I went with that name, but I'm open to other suggestions.

The code is a bit crude, but I thought maybe a PR is easier for discussing this than just an issue.

If this is something you're interested in and once the initial review is done, I'd also add the same for `args` perhaps?

Here it is in action:

https://user-images.githubusercontent.com/1301152/135626719-9826a95e-4ffe-4e3f-a59f-d9eb13a15e1c.mp4

### DIFF
--- a/engine/app/views/good_job/cron_schedules/index.html.erb
+++ b/engine/app/views/good_job/cron_schedules/index.html.erb
@@ -5,6 +5,13 @@
         <thead>
           <th>Cron Job Name</th>
           <th>Configuration</th>
+          <th>
+            Properties&nbsp;
+            <%= tag.button "Toggle", type: "button", class: "btn btn-sm btn-outline-primary", role: "button",
+                           data: { bs_toggle: "collapse", bs_target: ".job-properties" },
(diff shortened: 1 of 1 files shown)


bensheldon:
@aried3r Thanks for submitting this. I think it's fantastic. Some suggestions:

- For clarity, I think the header should match the configuration attribute. E.g. "Set" instead of "Properties".
- What do you think of only having an expanding toggle if there is configuration? e.g. If there is "None" or "Lambda/Callable", show that directly.

That's all. This is definitely an improvement.

aried3r:
I've removed the `<pre>` tag for when it's not a `Hash`. WDYT?

https://user-images.githubusercontent.com/1301152/135649373-d7df7c84-b984-43b9-bfa5-6225710e241e.mp4

bensheldon (approved):
Great! I'll merge this and allow you to do args separately. 

Btw, this spurred some thoughts that I captured in #382.

PR #394, opened 2021-10-01
Display schedule time relative to now.

morgoth:
Uses https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-time_ago_in_words

### DIFF
--- a/engine/app/helpers/good_job/application_helper.rb
+++ b/engine/app/helpers/good_job/application_helper.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 module GoodJob
   module ApplicationHelper
+    def relative_time(timestamp)
+      text = timestamp.future? ? "in #{time_ago_in_words(timestamp)}" : "#{time_ago_in_words(timestamp)} ago"
+      tag.time(text, datetime: timestamp, title: timestamp)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


engine/app/helpers/good_job/application_helper.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 module GoodJob
   module ApplicationHelper
+    def relative_time(timestamp)
+      text = timestamp.future? ? "in #{time_ago_in_words(timestamp)}" : "#{time_ago_in_words(timestamp)} ago"
+      content_tag(:time, text, datetime: timestamp, title: timestamp)
```
bensheldon, line 6: I think `content_tag` is deprecated in favor of `tag.time`.

bensheldon (approved):
Big improvement. I'm a fan of relative time with timestamp title. 

You might know better than me about the future compatibility of `content_tag`.

engine/app/helpers/good_job/application_helper.rb
morgoth: I changed it. I wasn't sure in which version of Rails `tag` was introduced, but we're already using it in engine, so it is fine.

bensheldon:
😍
 
Screen Shot 2021-10-01 at 5 47 41 PM

PR #395, opened 2021-10-01
Correct icon for alert flash

morgoth:


bensheldon:
Nice catch! Thank you!

PR #396, opened 2021-10-02
Display cron schedules args in dashboard

aried3r:
As promised the follow-up. Now there's quite a bit of code duplication. Do you want me to deduplicate the code or do we go by the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))?

Removed the `#` from `aria-controls` as the list of ID references doesn't use the `#` afaik.

Added args to the cleanup job so that we cover all cases for this column again.

Video shows that the buttons still trigger what they're supposed to.

https://user-images.githubusercontent.com/1301152/135731088-adb068ad-5ff3-4ff7-841f-d9c6b4fcd5c9.mp4

### DIFF
--- a/engine/app/views/good_job/cron_schedules/index.html.erb
+++ b/engine/app/views/good_job/cron_schedules/index.html.erb
@@ -9,7 +9,14 @@
             Set&nbsp;
             <%= tag.button "Toggle", type: "button", class: "btn btn-sm btn-outline-primary", role: "button",
                            data: { bs_toggle: "collapse", bs_target: ".job-properties" },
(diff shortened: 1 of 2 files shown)


bensheldon (approved):
I think this is strong as-is 👍 

Personally, I find that extracting/abstracting reusable html partials/helper logic can get gnarly quickly, and then become an obstacle to contributors who want to suggest UI improvements at the HTML level, but have to drop into some twisty Ruby code to do so. 

I'd like to keep the barrier to contributing to the Dashboard low at least for now while I'm still learning about how people want to use it.

PR #397, opened 2021-10-04
Correct app file

morgoth:
I believe it was some copy paste leftover

[ci skip]


bensheldon (approved):
Yep, copy-pasted. Thank you.

PR #398, opened 2021-10-04
Display info about used timezone.

morgoth:
It might be confusing if one is using UTC in the app, but is placed in
different timezone. Make it clear by expclitily stating it.


bensheldon (approved):
👏

PR #401, opened 2021-10-05
Update GH Test Matrix with more PG versions

tedhexaflow:


tedhexaflow:
Hi @bensheldon 

Is it ok having 21 checks? Or should we just keep the two latest versions? (13 and 14)

Should I also remove (ruby 2.7 with pg 10.8) from the matrix?

.github/workflows/test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         ruby: [2.5, 2.6, 2.7, 3.0, jruby-9.2, jruby-9.3]
-        pg: [12.5]
+        pg: [12, 13, 14]
         include:
           - ruby: 2.7
             pg: 10.8
```
bensheldon, line 41:
```suggestion
        pg: [14]
        include:
          - ruby: 3.0
            pg: 10.8
          - ruby: 3.0
            pg: 11
          - ruby: 3.0
            pg: 12
          - ruby: 3.0
            pg: 13
```

tedhexaflow: Thanks, @bensheldon,

I've added 10,11,12,13 also for the latest JRuby.

Hope you are ok with that. Thanks.

bensheldon:
Thank you!!

PR #440, opened 2021-10-29
Add fulltext search filter

### LINKED ISSUE #432: Add a search bar to Dashboard
Can be part of the filter interface as another filterable query param.

Should use fulltext search (`ts_*`), but skip the index for now and see how performance is.

bensheldon:
Closes #432.

~~Bit of a spike, but wanted to open up for feedback.~~

Some quirks:
- When searching jobs, it only searches the properties of the most recent execution. So while the "Last Error" (i.e. the error from the previous execution) of a job will appear in the dashboard, it's not queryable.
-  There is no search indexing, so this results in a row scan.

### DIFF
--- a/engine/app/views/good_job/shared/_filter.erb
+++ b/engine/app/views/good_job/shared/_filter.erb
@@ -1,6 +1,6 @@
 <div class='card mb-2'>
   <div class='card-body d-flex flex-wrap'>
-    <div class='me-4'>
+    <div class='mb-2 me-4'>
       <small>Filter by job class</small>
       <br>
       <% filter.job_classes.each do |name, count| %>
@@ -16,7 +16,7 @@
       <% end %>
     </div>
 
-    <div class='me-4'>
+    <div class='mb-2 me-4'>
       <small>Filter by state</small>
       <br>
       <% filter.states.each do |name, count| %>
@@ -32,7 +32,7 @@
       <% end %>
     </div>
 
-    <div>
+    <div class='mb-2 me-4'>
       <small>Filter by queue</small>
       <br>
       <% filter.queues.each do |name, count| %>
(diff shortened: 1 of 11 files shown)


engine/app/views/good_job/shared/_filter.erb
```diff
@@ -47,5 +47,20 @@
         <% end %>
       <% end %>
     </div>
+
+    <div>
+      <%= form_with(url: "", method: :get) do |form| %>
```
morgoth, line 52: you could add here `local: true` as this value might differ because it can be configured globally in the app https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-form_with-label-form_with+options

engine/app/views/good_job/shared/_filter.erb
```diff
@@ -47,5 +47,20 @@
...
+        <% filter.to_params(query: nil).each do |key, value| %>
+          <%= form.hidden_field(key.to_sym, value: value) %>
+        <% end %>
+
+        <small><%= form.label :query, "Search" %></small>
+        <div class="input-group input-group-sm">
+          <%= form.text_field :query, value: params[:query], class: "form-control" %>
```
morgoth, line 59: maybe `search_field`? https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html#method-i-search_field
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

morgoth (approved):

engine/app/views/good_job/shared/_filter.erb
bensheldon: Good catch! Thanks!

engine/app/views/good_job/shared/_filter.erb
bensheldon: 👍

PR #441, opened 2021-10-30
Allow for cron schedules to be expressed using fugit natural language parsing

### LINKED ISSUE #439: Ability to express cron schedule using fugit natural language parser
Hi Ben

First of all thank you for `good_job` 🙏 It's been running rock solid and I've recently migrated all my Kubernetes cron-jobs to good_job recurring tasks.

Question: Would you accept a pull request that made it possible to use the [fugit natural language parser](https://github.com/floraison/fugit#fugitnat)? Without having given it a lot of thought we could check in [next_at](https://github.com/bensheldon/good_job/blob/main/lib/good_job/cron_entry.rb#L44-L47) whether `cron` is already an instance of `Fugit::Cron`. That way, in the schedule configuration you could do:

```
config.good_job.cron = {
  # Every 15 minutes, enqueue `ExampleJob.set(priority: -10).perform_later(42, name: "Alice")`
  frequent_task: { # each recurring job must have a unique key
    cron: Fugit::Nat.parse('every day at five'),
    class: "ExampleJob", # reference the Job class with a string
    args: [42, { name: "Alice" }], # arguments to pass; can also be a proc e.g. `-> { { when: Time.now } }`
    set: { priority: -10 }, # additional ActiveJob properties; can also be a lambda/proc e.g. `-> { { priority: [1,2].sample } }`
    description: "Something helpful", # optional description that appears in Dashboard (coming soon!)
  },
  another_task: {
    cron: "0 0,12 * * *",
    class: "AnotherJob",
  },
  # etc.
}
```

I realise it would couple the gem more tightly to `fugit` so I can totally understand if you feel the benefits are not worth it. :)

jgrau:
@bensheldon something like this? Please feel free to nitpick!

Closes #439

### DIFF
--- a/lib/good_job/cron_entry.rb
+++ b/lib/good_job/cron_entry.rb
@@ -25,6 +25,8 @@ module GoodJob # :nodoc:
 
     def initialize(params = {})
       @params = params
+        
+      raise ArgumentError, "Invalid cron format: '#{cron}'" unless fugit.instance_of?(Fugit::Cron)
     end
 
     def key
@@ -60,6 +62,10 @@ module GoodJob # :nodoc:
     
     def schedule
       fugit.original
+    end
+
+    def fugit
+      @fugit ||= Fugit.parse(cron)
     end
 
     def jobs
@@ -104,16 +110,6 @@ module GoodJob # :nodoc:
         args: display_property(args),
         description: display_property(description),
       }
-    end
-
-    def fugit
-      @fugit ||= begin
-        parsed_cron = Fugit.parse(cron)
-        
-        raise ArgumentError, "Invalid cron format: '#{cron}'" unless parsed_cron.instance_of?(Fugit::Cron)
-        
-        parsed_cron
-      end
     end
     
     private
--- a/lib/good_job/cron_entry.rb
+++ b/lib/good_job/cron_entry.rb
@@ -25,7 +25,7 @@ module GoodJob # :nodoc:
 
     def initialize(params = {})
       @params = params
-        
+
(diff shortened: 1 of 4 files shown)


jgrau:
@bensheldon I did two iterations on this: the first that assumed the user would supply the `Fugit::Cron` instance as the configuration and the second using `Fugit.parse` which works with both the cron-style and natural language. <- that's the reason for the 2 commits.

lib/good_job/cron_entry.rb
```diff
@@ -96,13 +99,23 @@ def display_properties
...
         args: display_property(args),
         description: display_property(description),
       }
     end
 
+    def fugit
+      @fugit ||= begin
+        parsed_cron = Fugit.parse(cron)
+        
+        raise ArgumentError, "Invalid cron format: '#{cron}'" unless parsed_cron.instance_of?(Fugit::Cron)
```
bensheldon, line 113: I think this ArgumentError would be clearer if placed in #initialize.

bensheldon (approved):
This is fantastic! I like the version that still takes a string, but with a broader syntax. I left one comment about surfacing the ArgumentError earlier in execution.

Let me know your thoughts on that and I'll merge this 🙌🏻

lib/good_job/cron_entry.rb
jgrau: I agree :) Updated. And fixed linting errors.

bensheldon:
@jgrau Thank you! 🎉  This has been released in GoodJob [`v2.6.0`](https://github.com/bensheldon/good_job/releases/tag/v2.6.0)

Issue #403, opened 2021-10-05
HTTP health check endpoint


bouk:
Hi @bensheldon, thanks for the great gem.

We're using good_job in an environment that requires an HTTP health check to make sure the service is healthy. For Sidekiq there's a plugin available that does something similar: https://github.com/arturictus/sidekiq_alive

I couldn't figure out a good way to build a plugin like this for good_job, so I was wondering if you've dealt with this before and have any tips, or could suggest a good way to integrate an endpoint like this with good_job.

bensheldon:
I am really excited by this challenge. Both because I think it's operationally important to have probes, and because it's an interesting technical challenge.

For the webserver part of the probe:

- If you're using GoodJob in `async` mode (i.e. running jobs inside the webserver), you can add the code below to your webserver's existing probe.
- For the CLI, it would need to have a tiny server built into it. That seems like it would be 20 lines of code to do with Rack... though I'm curious what the tiniest Rails API server implementation would be because the CLI is already initializing Rails probably YAGNI). 

For the content of the probes:

 - Readiness: the job schedulers have initialized and are ready to work jobs:
     ```ruby
     GoodJob::Scheduler.instances.all?(&:running?)
     ```
 - Liveliness: I have experienced Ops people arguing about what a Liveness probe should entail and whether it should check on service dependencies (e.g. database). This [sort of thing](https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html). If I accept that liveliness includes a working database connection, then I would use:
 
    ```ruby
    GoodJob::Scheduler.instances.all?(&:running?) &&
      GoodJob::Notifier.instances.all?(&:listening?)
    ```
    
    
@bouk Is this a project you'd want to do? I think the smallest scope would be adding a simple Rack server (like [this](https://github.com/arturictus/sidekiq_alive/blob/master/lib/sidekiq_alive/server.rb)), starting the Rack server inside of a Concurrent::Future thread somewhere around [here](https://github.com/bensheldon/good_job/blob/f114b973f996da02ae0322608421ccc89668e7d4/lib/good_job/cli.rb#L95), and maybe making it conditional on some configuration like `good_job.config.probe_port`.

bensheldon:
@aried3r Thanks! Yes, it should be `GoodJob::Scheduler.instances.any? && GoodJob::Scheduler.instances.all?(&:running?)`

Can you open an issue for what you're seeing about the instances not being started?

bensheldon:
[...]

PR #452, opened 2021-11-06
Add http probe for CLI healthcheck/readiness/liveliness

bensheldon:
Connects to #403. Adds separate endpoints that are descriptive of GoodJob's state (started/connected) because I do not want to be opinionated about what constitutes an appropriate probe. 

This PR uses WEBrick as the http server. Ruby 3.0 removes webrick from the stdlib and thus it has to be included in the gemspec.

### DIFF
--- a/good_job.gemspec
+++ b/good_job.gemspec
@@ -54,6 +54,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fugit", ">= 1.1"
   spec.add_dependency "railties", ">= 5.2.0"
   spec.add_dependency "thor", ">= 0.14.1"
+  spec.add_dependency "webrick", ">= 1.3"
   spec.add_dependency "zeitwerk", ">= 2.0"
 
   spec.add_development_dependency "benchmark-ips"
--- a/lib/good_job/probe_server.rb
+++ b/lib/good_job/probe_server.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module GoodJob
+  class ProbeServer
+    RACK_SERVER = 'webrick'
+
+    def self.task_observer(time, output, thread_error) # rubocop:disable Lint/UnusedMethodArgument
+      return if thread_error.is_a? Concurrent::CancelledOperationError
+
+      GoodJob.on_thread_error.call(thread_error) if thread_error && GoodJob.on_thread_error.respond_to?(:call)
+    end
+
+    def initialize(port:)
+      @port = port
+    end
+
+    def start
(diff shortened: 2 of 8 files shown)


good_job.gemspec
```diff
@@ -54,6 +54,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fugit", ">= 1.1"
   spec.add_dependency "railties", ">= 5.2.0"
   spec.add_dependency "thor", ">= 0.14.1"
+  spec.add_dependency "webrick", ">= 1.3"
```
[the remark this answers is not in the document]
bensheldon: I didn't put much thought into it, though I'm not quite sure how I came to `v1.3` as it looks like that shipped with Ruby 2.4, which GoodJob does not support. 

I think of the gemspec as descriptive of compatibility; I think security should be considered at the application that is consuming the gemspec. Trying to balance:

- if the application is not previously including webrick, that bundler is not defaulting to an insecure version (I don't know how Bundler decides whether to pull the latest version, or just whatever it has lying around locally)
- if the application is previously including webrick, not forcing the application to update the version of webrick because that might be a blocker for the adopting/updating GoodJob in a controlled environment.

The other thing I did consider, rather than pulling in webrick, was just pasting in one of the many "rack-compatible http server in 30 lines of ruby" and packaging it directly. These health check requests are very, very simple and really only intended to be used internally.

lib/good_job/probe_server.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      GoodJob.on_thread_error.call(thread_error) if thread_error && GoodJob.on_thread_error.respond_to?(:call)
+    end
+
+    def initialize(port:)
+      @port = port
+    end
+
+    def start
+      @handler = Rack::Handler.get(RACK_SERVER)
```
bouk, line 18: @bensheldon if you use `Rack::Handler.default` here you don't need the dependency on `webrick` and the handler can be set with an environment flag

bensheldon: Ooh, that's interesting. When trying that change, it defaults to Puma, but Puma intercepts `ctrl-c`/`Interrupt` which becomes really problematic :-(

This isn't intended to be anything more than a container healthcheck, which makes me think that I should explore pasting in a "rack-compatible http server in 30 lines of ruby" snippet of code and avoid the dependency altogether.

bouk: Ah I see, then webrick makes more sense 👍🏻

PR #456, opened 2021-11-24
Do not delete csp policies when checking csp policies

JonathanFrias:
I noticed that updating the good_job dashboard broke the javascript, due to it being blocked by the content security policy.
For applications that use a custom CSP, good_job breaks. It is due to very strange code in the `action_dispatch` gem. If you check the policies it triggers a side effect. See below the offending method source.

This pull request enabled good_job to check the csp policies without actually removing them.

This is the code in `action_dispatch` that causes the issue. Really strange to me that checking the policies' values causes them to be removed.

```ruby
From: /Users/jonathan/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/actionpack-6.1.4/lib/action_dispatch/http/content_security_policy.rb:164:
Owner: ActionDispatch::ContentSecurityPolicy
Visibility: public
Signature: default_src(*sources)
Number of lines: 7

define_method(name) do |*sources|
  if sources.first
    @directives[directive] = apply_mappings(sources)
  else
    @directives.delete(directive)
  end
end
```

### DIFF
--- a/engine/app/controllers/good_job/base_controller.rb
+++ b/engine/app/controllers/good_job/base_controller.rb
@@ -6,16 +6,16 @@ module GoodJob
     around_action :switch_locale
 
     content_security_policy do |policy|
-      policy.default_src(:none) if policy.default_src.blank?
-      policy.connect_src(:self) if policy.connect_src.blank?
-      policy.base_uri(:none) if policy.base_uri.blank?
-      policy.font_src(:self) if policy.font_src.blank?
-      policy.img_src(:self, :data) if policy.img_src.blank?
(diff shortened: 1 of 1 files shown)


bensheldon:
@JonathanFrias wow. Thanks for digging into this bug. That interface is surprising! I'll review and merge. Thank you!

bensheldon:
@JonathanFrias thanks again! I just released this in `v2.7.1`.

JonathanFrias:
Thank you!! Now I can just update to the latest! :P

Issue #412, opened 2021-10-08
GoodJob should automatically clean up after itself and delete old job records


bouk:
It would be useful if there was a built-in ActiveJob that just calls `GoodJob.cleanup_preserved_jobs`, which could then be used with the Cron feature. That way it would be very easy to set up periodic finished job cleaning.

AxelTheGerman:
Love this feature, just one word of caution. Running automagically without the developer being able to change the default could cause issues in production, e.g. using resources to clean up at a bad time (think of classic GC)

bensheldon:
@AxelTheGerman thank you for the comment! Do you have any ideas for what would be good configuration options? I'm exposing two in my spike on the feature:

- `cleanup_interval_jobs` (integer) Number of jobs a Scheduler will execute before cleaning up preserved jobs. Defaults to `nil`. Can also be set with  the environment variable `GOOD_JOB_CLEANUP_INTERVAL_JOBS`.
- `cleanup_interval_seconds` (integer) Number of seconds a Scheduler will wait before cleaning up preserved jobs. Defaults to `nil`. Can also be set with  the environment variable `GOOD_JOB_CLEANUP_INTERVAL_SECONDS`.

AxelTheGerman:
That's probably a good starting point!

So it will be whichever is hit first, i.e. 1000 jobs or 1 hour.

The potential issues I mentioned in my previous post probably won't happen until someone is running `good_job` under high load - at which point they could turn it off and implement their own clean up. Or then we could come up with a better "algorithm" to decide when to clean up

🚀

AxelTheGerman:
> until GoodJob 3.0, when it will 1 hour

Not sure 1 hour is enough for systems with low throughput. In my apps I don't have many jobs running so I'd probably keep them around for 1 day - 1 week just in case.

But again, it's configurable - via code or ENV (which I love)

Issue #380, opened 2021-09-16
Inconsistencies in configuration settings


paul:
Different config settings seem to behave differently based on how they're defined. I was trying to investigate why GoodJob wasn't preserving completed jobs, even though I have the setting enabled.

In `config/initializers/good_job.rb`, I have this:

```ruby
MyApp::Application.configure do
  config.active_job.queue_adapter = :good_job

  config.good_job.tap do |good_job|
    good_job.execution_mode = :external
    good_job.preserve_job_records = true

    good_job.active_record_parent_class = "ApplicationRecord"
    good_job.on_thread_error = ->(exception) { Honeybadger.notify(exception) }

    # Enabled by the procfile on "worker", but on heroku we only want to run it
    # on the first worker dyno
    dyno = ENV["DYNO"]
    if dyno && dyno != "worker.1"
      good_job.enable_cron = false
    end

    good_job.cron = {
      refresh_slack_tokens: {
        cron:  "0 * * * *",
        class: "RefreshSlackAppTokens",
      }
    }
  end
end
```

I'm enabling `preserve_job_records`, and its set when I examine the config, but when I look at the value on GoodJob, its not set:

```
[1] pry(main)> Rails.configuration.good_job
=> {:cron=>{:refresh_slack_tokens=>{:cron=>"0 * * * *", :class=>"RefreshSlackAppTokens"}},
 :execution_mode=>:external,
 :active_record_parent_class=>"ApplicationRecord",
 :on_thread_error=>#<Proc:0x00005607b3af1160 /home/rando/Code/textus/flobot/config/initializers/good_job.rb:12 (lambda)>,
 :preserve_job_records=>true}
[2] pry(main)> GoodJob.preserve_job_records
=> false
```

Other settings set the same way, like `execution_mode`, do what I expect and change how GoodJob runs. It seems it allows `preserve_job_records` to be set on config, but it has no effect. Alternatively, if I attempt to set all these values on GoodJob directly some of them (like `GoodJob.execution_mode = :extenal`) fail with "undefined method".

It looks like GoodJob defines several `mattr_accessors`, but which settings are expected to be configured on `config`, and which on `GoodJob` directly?

bensheldon:
Hi @paul. Thank you for opening this issue. 

[...]

PR #460, opened 2021-11-27
Allow GoodJob global configuration accessors to also be set via Rails config hash

bensheldon:
~~Deprecates all `GoodJob` configuration accessors, in favor of Rails configuration hash.~~

I was originally planning to deprecate them now, but I plan with GoodJob 3.0 to change them as a result of #412 so I'll deprecate most of them (with the exception of `logger` and `on_thread_error`) when I do that.

- Creates a callable `GoodJob._on_thread_error` method
- Wraps all of the Railtie code `initializer` blocks
- Fixes some app configuration from #454 

Connects to #380.

### DIFF
--- a/lib/good_job/railtie.rb
+++ b/lib/good_job/railtie.rb
@@ -7,7 +7,7 @@ module GoodJob
 
     initializer "good_job.logger" do |_app|
       ActiveSupport.on_load(:good_job) do
-        self.logger = ::Rails.logger
+        self.logger = ::Rails.logger if GoodJob.logger == GoodJob::DEFAULT_LOGGER
       end
       GoodJob::LogSubscriber.attach_to :good_job
     end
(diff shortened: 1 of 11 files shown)


lib/good_job/railtie.rb
```diff
@@ -23,6 +23,11 @@ class Railtie < ::Rails::Railtie
     end
 
     config.after_initialize do
+      GoodJob.logger = Rails.application.config.good_job.logger if Rails.application.config.good_job.logger.present?
+      GoodJob.on_thread_error = Rails.application.config.good_job.on_thread_error if Rails.application.config.good_job.on_thread_error.present?
+      GoodJob.preserve_job_records = Rails.application.config.good_job.preserve_job_records if Rails.application.config.good_job.preserve_job_records.present?
+      GoodJob.retry_on_unhandled_error = Rails.application.config.good_job.retry_on_unhandled_error if Rails.application.config.good_job.retry_on_unhandled_error.present?
```
bensheldon, line 29: This is the meat of the change, with the effect of overriding the values assigned to `GoodJob` with values in Rails config.

PR #477, opened 2022-01-03
Add indexes to `good_jobs.finished_at` and have `GoodJob.cleanup_preserved_jobs` delete all executions for a given job

### LINKED ISSUE #476: finished_at should be indexed and clean up should clean up all of a job's executions
Thanks for adding the new option to automatically cleanup preserved jobs!

I noticed that `GoodJob::Execution.finished(timestamp).delete_all` inside of `cleanup_preserved_jobs` is querying `finished_at` but that column isn't indexed resulting in a table scan. I'm thinking `finished_at` needs an index (possibly conditional on being `not null`)?

bensheldon:
Closes #476

### DIFF
--- a/lib/generators/good_job/templates/install/migrations/create_good_jobs.rb.erb
+++ b/lib/generators/good_job/templates/install/migrations/create_good_jobs.rb.erb
@@ -32,5 +32,7 @@ class CreateGoodJobs < ActiveRecord::Migration<%= migration_version %>
     add_index :good_jobs, :concurrency_key, where: "(finished_at IS NULL)", name: :index_good_jobs_on_concurrency_key_when_unfinished
     add_index :good_jobs, [:cron_key, :created_at], name: :index_good_jobs_on_cron_key_and_created_at
     add_index :good_jobs, [:cron_key, :cron_at], name: :index_good_jobs_on_cron_key_and_cron_at, unique: true
+    add_index :good_jobs, [:active_job_id], name: :index_good_jobs_on_active_job_id
+    add_index :good_jobs, [:finished_at], where: "retried_good_job_id IS NULL AND finished_at IS NOT NULL", name: :index_good_jobs_jobs_on_finished_at
   end
 end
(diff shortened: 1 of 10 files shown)


6temes:
Hi @bensheldon 

I just realized that, after running the latest migrations, I have two indices:

```
    t.index ["active_job_id", "created_at"], name: "index_good_jobs_on_active_job_id_and_created_at"
    t.index ["active_job_id"], name: "index_good_jobs_on_active_job_id"
```

I may be wrong, but wouldn't `index_good_jobs_on_active_job_id` be redundant? If I understand well, `index_good_jobs_on_active_job_id_and_created_at` is already indexing the `active_job_id` and, having both indices would only slow down writes.

bensheldon:
@6temes you're right, those are redundant, and the singular/non-composite index could be dropped.

nsilva-ta:
Hi @bensheldon 

Why do we have an array with only one symbol for the column name?

```ruby
    add_index :good_jobs, [:finished_at], where: "retried_good_job_id IS NULL AND finished_at IS NOT NULL", name: :index_good_jobs_jobs_on_finished_at
```

reczy:
Hi @nsilva-ta,

Both ways of writing in the column_name are acceptable (symbol or array of symbols) - [see here](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_index).

In fact, if you look at your schema after the migration, you'll see that they're all represented in array format:
```
    t.index ["active_job_id", "created_at"], name: "index_good_jobs_on_active_job_id_and_created_at"
    t.index ["active_job_id"], name: "index_good_jobs_on_active_job_id"
    t.index ["concurrency_key"], name: "index_good_jobs_on_concurrency_key_when_unfinished", where: "(finished_at IS NULL)"
    t.index ["cron_key", "created_at"], name: "index_good_jobs_on_cron_key_and_created_at"
    t.index ["cron_key", "cron_at"], name: "index_good_jobs_on_cron_key_and_cron_at", unique: true
    t.index ["finished_at"], name: "index_good_jobs_jobs_on_finished_at", where: "((retried_good_job_id IS NULL) AND (finished_at IS NOT NULL))"
    t.index ["queue_name", "scheduled_at"], name: "index_good_jobs_on_queue_name_and_scheduled_at", where: "(finished_at IS NULL)"
    t.index ["scheduled_at"], name: "index_good_jobs_on_scheduled_at", where: "(finished_at IS NULL)"
```

PR #479, opened 2022-01-06
Add JRuby / JDBC support for LISTEN 

bensheldon:
- Follows pattern of https://jdbc.postgresql.org/documentation/81/listennotify.html to poll the database for notifications in the absence of native `wait_for_notify`.
- To poll, the adapter makes a noop `SELECT 1` query every 1 second.

@tedhexaflow fyi, I'd love your feedback on this


tedhexaflow:
Hi @bensheldon,

I'm not very familiar with postgre Listen/Notify features. 

Is there any specific you would like me to check?

Thanks.

bensheldon:
@tedhexaflow I think mainly I'm curious about:

- I'm assuming that the JDBC interface is very stable.  Should I be doing some deeper duck-typing/method testing to ensure that `getNotifications` exists? The docs make me think it's stable.
- Is there any downside you can think of to polling the database every second with a noop query? I assume that Java is even better at parallelization than MRI would be.

And I guess more generally, _I_ think LISTEN/NOTIFY for JRuby is exciting and brings parity between MRI and JRuby in GoodJob's benefits of low-latency job execution, so any confidence boosting there is appreciated 😄 

And lastly, I wonder if I should suggest that this be implemented in Rails ActionCable Postgres adapter too, because that's what inspired GoodJob's implementation: https://github.com/rails/rails/blob/ce3db04cd74ed267fb083a1342672194912ea87d/actioncable/lib/action_cable/subscription_adapter/postgresql.rb#L106-L108

tedhexaflow:
Hi @bensheldon 

Thanks for the detail.

I think I am not the right person to provide helpful feedback for those questions. 

Let me try to reach Charles @headius from JRuby and ask for help to provide feedback.

headius:
@bensheldon @tedhexaflow I don't know much about this library or about Postgres notifications in general, but...

* Anything in JDBC should be stable and work well with JRuby, so I assume notifications are included in that.
* Background polling should not affect performance of foreground workers/threads in JRuby, so I don't see a problem there.

We're happy to help make JRuby support happen, if that means answering questions, fixing JRuby-related issues, or understanding Java APIs. Stop by our Matrix channel if you need anything!

bensheldon:
@headius thank you! That's really helpful context. I joined the Matrix channel and I'll definitely bring my questions there. Thanks again!

tedhexaflow:
Thank you so much 🙏 @headius.

PR #481, opened 2022-01-07
Bug Fix when using Concurrency

ryanmansfield:
good_job_pr

Came across this issue in development, where .pluck on line 65 does not seem to be working properly,
it returns allowed_active_job_ids = [{'active_job_id' => XXXX}] instead of [XXX].

It's not very dry doing the .flat_map again on the next line, but it works properly.


bensheldon:
@ryanmansfield thanks for the PR! That's unexpected! AR's `#pluck` should return an array of values. I wonder why it's not. I'll poke at this a bit.

bensheldon:
@ryanmansfield strange, I get what I expected when running locally, which is an array of values. What version of Ruby on Rails are you running? 

Screen Shot 2022-01-07 at 4 53 55 PM

You did uncover that this line isn't covered by the tests.

ryanmansfield:
@bensheldon my appologies, further investigation on my end shows that the default nature of pluck was overridden on the app I am working on.

bensheldon:
@ryanmansfield no sweat! Three positive things came out of this:

1. We found the underlying issue.
2. I learned that there is an important yet untested line of code that needs tests.
3. In researching this, I discovered https://github.com/girishso/pluck_to_hash which might be a good alternative to the overriding in your app.

PR #484, opened 2022-01-14
$ symbol gets copied when clicking on the copy button

zeevy:
copying and pasting of setup commands not working due to "$" symbol


bensheldon:
@zeevy Excellent observation. Thank you!

PR #485, opened 2022-01-15
Add details about exactly where to require the engine.

phallstrom:
This clarifies exactly where to require goodjob in config/application.rb
and adds a note about how to resolve the issue when the dashboard won't
reload after saving a file.

Context: I skimmed the README when I set it up and made my
application.rb look like this:

```require "rails"

require "active_model/railtie"
require "active_job/railtie"
require "active_record/railtie"
require "active_storage/engine"
require "action_controller/railtie"
require "action_mailer/railtie"
require "action_text/engine"
require "action_view/railtie"
require "action_cable/engine"
require "sprockets/railtie"

Bundler.require(*Rails.groups)

require "good_job/engine"
```

With this setup, the dashboard would work fine until I saved any file
(ie. a model). After that reloading the dashboard would give this error:

Routing Error
uninitialized constant GoodJob::ExecutionsController

By moving the require line _before_ the `Bundler.require` line, this
problem went away.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ _🚧 GoodJob's dashboard is a work in progress. Please contribute ideas and cod
 
 GoodJob includes a Dashboard as a mountable `Rails::Engine`.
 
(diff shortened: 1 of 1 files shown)


bensheldon:
Thank you for this Readme improvement! 🙌

Issue #408, opened 2021-10-07
Internationalize/I18n the Dashboard Engine


bensheldon:
It would be nice to have English UI strings extracted to a `yml` file to allow support for other locales.

We're liable to run into some RTL and spacing challenges once other translations are added, but we can address those as they arise. Ideally dates/times should be localized (as we they are currently locked to English in #407)

bensheldon:
@query-string I'd love the help. Maybe you could scope down to doing the Cron screen of the dashboard. 

I'm imagining there might be a messy merge because all the strings will end up in the same en.yml, but I can get us past that barrier when it happens.

effgenij:
@bensheldon @query-string If it's not urgent, I'll have it done by the end of the week. If it is urgent, you can pass it on to someone else.

Liberatys:
@bensheldon is this still a feature you'd like to see? :D I'd like to start contributing to good_job(as I'm using it in all my new projects ^^ 👍🏻  Great job btw). So I think this would be a good first issue to setup the project and get to know the structure and process :D

bensheldon:
@Liberatys yep! Still a feature I'd like to see. Thank you!

JuanVqz:
@bensheldon one more try hahaha, I would like to work on this, but I'll be honest I have no experience doing things in gems, (that is part of the reason I just want to contribute). could you help me if I get stuck? I know maybe I'll ask several times but if you agree I can tackle it.

bensheldon:
@JuanVqz I'm happy to help. Btw, if you just want to help, this might be a better starting issue: https://github.com/bensheldon/good_job/issues/341

JuanVqz:
> @JuanVqz I'm happy to help. Btw, if you just want to help, this might be a better starting issue: https://github.com/bensheldon/good_job/issues/341

Thank you for your faster response, since you will be helping me I think, I can tackle both of them. 😃

bensheldon:
I think we can call this done for now. Improvements are still invited, but the main thrust has been completed 🎉

PR #497, opened 2022-01-26
Internationalize/I18n the Dashboard Engine

JuanVqz:
@bensheldon, I had a little bit of progress, so I would like to share it with you

Related #408 

Links:
[ruby on rails docs](https://guides.rubyonrails.org/i18n.html)


JuanVqz:
@bensheldon what do you think about the variable's name in the locales/en.yml file.

bensheldon:
@JuanVqz That looks fantastic! You did what I was imagining: Used relative keys (e.g. `.key`) that are namespaced to the template 👍 

For the key names themselves, generic naming is ideal that describes the purpose of the string, rather than the contents. E.g. 

- "timezone_notice"
- "current_time"
- "coming_soon"

JuanVqz:
I have already fixed the test that struggled in the CI.

bensheldon:
This looks great. I think this would be a good time for you to pause on extracting strings, and next look into UI and routing to display the different locales.

JuanVqz:
> This looks great. I think this would be a good time for you to pause on extracting strings, and next look into UI and routing to display the different locales. 

Yes, I agree, it seems a good time to pause it. What do you mean by looking into IU, should I add a select field with available languages?

Also, I have a question, until now I have been seen the /good_job page though the rspec test when running the system spec, but I'm wondering how can I see it in a rails app, I saw in the read me that I should point to the local gem but I don't fully understand how can I do it.

bensheldon:
> What do you mean by looking into UI, should I add a select field with available languages?

Yep. I think simplest would just be a drop down menu in the Navbar with simple links. The Bootstrap docs should show you how to do that. 

Great question about running it. You can run GoodJob's dashboard directly from the good job repo:

```
bin/setup
bin/test_app s # runs rails server
```

JuanVqz:
> Great question about running it. You can run GoodJob's dashboard directly from the good job repo:
> 
> ```
> bin/setup
> bin/test_app s # runs rails server
> ```

this is super useful, thank you for sharing it.

I already started playing with the select field in the view, and I noticed that we may use the `locale` functionality in at least two ways, and I'm wondering which one would you like to have,

1 `localhost:3000/en/good_job`
  I think, I will need to scope the routes in the route file
  ```ruby
    # route.rb
    scope :lang do
      # routes
    end
  ```

2 `localhost:3000/good_job?locale=en`

Also, I'm wondering if the select field, should change the language once the user changes the option, or could we have a button to trigger the selection...

bensheldon:
@JuanVqz oops, I didn't answer your question. I think option #2 is good for now. 

This is the dropdown menu I think you should use, which would contain links: https://getbootstrap.com/docs/5.1/components/navs-tabs/#using-dropdowns

JuanVqz:
> @JuanVqz oops, I didn't answer your question. I think option #2 is good for now.
> 
> This is the dropdown menu I think you should use, which would contain links: https://getbootstrap.com/docs/5.1/components/navs-tabs/#using-dropdowns

There are some failing tests but it seems like we have some progress here

Peek 2022-03-28 23-06

bensheldon:
@JuanVqz AWESOME! I will dive in later this week and get those tests passing. This looks really good.

JuanVqz:
> @JuanVqz AWESOME! I will dive in later this week and get those tests passing. This looks really good.

it seems like I got it green, there were some minor issues, also, I added the footer partial which didn't exist when I started working on this feature, now it seems ready to go...I'm open to making changes if needed. :+1:

engine/app/controllers/good_job/base_controller.rb
```diff
@@ -30,8 +30,8 @@ def default_url_options(options = {})
 
     private
 
-    def set_locale
-      I18n.locale = current_locale
+    def switch_locale(&action)
+      I18n.with_locale(current_locale, &action)
```
JuanVqz, line 34: what do you prefer `with_locale`?

bensheldon: Good question. I like that `with_locale` only changes the locale within the context of the block and then returns to the previous value. Because the Goodjob engine is running in the context of a parent application, I don't want to change the `I18n.locale` value globally because it might have an unexpected side effect in the parent application.

JuanVqz: Good to know, yes, we don't want unexpected behavior 👍

Thank you for the info

bensheldon:
@JuanVqz thank you! I feel great about this. Any items before I merge and release?

Thank you so much!!

JuanVqz:
> @JuanVqz thank you! I feel great about this. Any items before I merge and release?
> 
> Thank you so much!! 

I'm happy to be part of it, I don't think so, I believe is ready to merge! 

Thank you for the opportunity to contribute, I'll be around if I can take another issue ⚠️

JuanVqz:
:heart: thank you!!!

engine/app/controllers/good_job/base_controller.rb
```diff
@@ -24,10 +24,28 @@ class BaseController < ActionController::Base # rubocop:disable Rails/Applicatio
       request.content_security_policy_nonce_generator = ->(_request) { SecureRandom.base64(16) }
     end
 
+    def default_url_options(options = {})
```
igas, line 27: Is there a particular need for it to be public? Now it's treated as an action.

bensheldon: @igas good catch. Could you make a PR for that?

PR #498, opened 2022-01-26
Test with Ruby 3.1

aried3r:


Appraisals
```diff
@@ -31,3 +34,12 @@ if ruby_27_or_higher && !jruby
   #   gem "rails", github: "rails/rails", branch: "main"
   # end
 end
+
+if ruby_31_or_higher && !jruby
+  appraise "rails-7.0-ruby-3.1" do
+    gem "capybara", "~> 3.36" # For Ruby 3.1 support https://github.com/teamcapybara/capybara/pull/2468
+    gem "psych", "~> 4.0"
```
aried3r, line 41: This should not be necessary, as Ruby 3.1 ships with psych 4. https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

The error that we're seeing in the test run comes from the test app, which uses aliases. See also https://gist.github.com/yahonda/2776d8d7b6ea7045359f38c10449937b#rails-60z and https://github.com/rails/rails/issues/44090.

I'll try to reproduce this locally but I'm not exactly sure where `test_app` takes its gems from. Is it really from appraisal or from this file? https://github.com/bensheldon/good_job/blob/main/Gemfile.lock It would explain the error perhaps, but would also mean that CI always uses the same rails version at least for tests using `test_app`.

gemfiles/rails_7.0_ruby_3.1.gemfile
```diff
@@ -0,0 +1,29 @@
+# This file was generated by Appraisal
+
+source "https://rubygems.org"
+
+gem "activerecord-jdbcpostgresql-adapter", platforms: [:jruby]
+gem "appraisal", branch: "fix-bundle-env", git: "https://github.com/bensheldon/appraisal.git"
+gem "pg", platforms: [:mri, :mingw, :x64_mingw]
+gem "rails", "~> 7.0.1"
+gem "capybara", "~> 3.36"
```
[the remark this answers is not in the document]
JuanVqz: Hi @aried3r, If may I, the linter is asking for ordering the gems in alphabetic order, (I sorted only the ones are shown in the snippet but it should be done for all the file content)  

```ruby
gem "capybara", "~> 3.36"
gem "rails", "~> 7.0.1"
```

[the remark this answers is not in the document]
aried3r: Hey @JuanVqz! Thanks for pointing this out. However this file is automatically being generated by the `appraisal` gem and not handwritten. I fear any manual change would be overwritten once the gemfiles are being regenerated.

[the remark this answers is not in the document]
bensheldon: I think that the `gemfiles` directory can be ignored by Rubocop because the files are generated by Appraisal 👍

Appraisals
bensheldon: I added `puts Rails.gem_version` to an initializer inside of `test_app` and then ran `LOUD=1 bundle exec appraisal rspec`. It was putting out the expected Rails versions from the Appraisal gemfiles. 

I just did the most shallow of reproduction locally, by checking out your PR and trying to run `bundle exec rspec` and I _did_ get the psych error:

```
❯ bundle exec rspec

An error occurred while loading ./spec/app/jobs/example_job_spec.rb.
Failure/Error: Rails.application.initialize!

Psych::BadAlias:
  Unknown alias: defaults
# ./spec/test_app/config/environment.rb:5:in `<top (required)>'
# ./spec/rails_helper.rb:14:in `require'
# ./spec/rails_helper.rb:14:in `<top (required)>'
# ./spec/app/jobs/example_job_spec.rb:2:in `require'
# ./spec/app/jobs/example_job_spec.rb:2:in `<top (required)>'
```

bensheldon: Ah, ok, I understand now.

As part of the test setup, before invoking `bundle exec appraisal rspec` the test runner sets up the database. That command is executed outside of Appraisal by the Rails version in the root Gemfile, which is Rails 6.1 which isn't compatible with Ruby 3.1.

https://github.com/bensheldon/good_job/blob/6f0a3fa828e1937184241de8d3d6b9b8d7ce12d2/.github/workflows/test.yml#L99-L100

I think the simplest, though maybe embarrassingly simple solution would be to change that command to be `bundle exec appraisal spec/test_app/bin/rails db:test:prepare`. It would run across all the appraisals unnecessarily, but I think the command should be idempotent and it means not having to duplicate the ruby/dependency logic in more places.

aried3r: Yeah, that's what I thought. I'll try that, thank you.

bensheldon:
I already think the matrix is a bit excessive 😬

I think we only need one version of CRuby and one version of JRuby to test across different different versions of Postgres. Everything else can be locked to a specific Postgres version.

aried3r:
> I think we only need one version of CRuby and one version of JRuby to test across different different versions of Postgres.

Is it coincidence that it was the previously latest CRuby? Or in other words, should it always be the latest stable Ruby version (now 3.1) that's tested across different versions of Postgres or should we leave it with 3.0?

aried3r:
I believe the test failure is unrelated to this PR and simply a timeout. Would you like anything else before this can get merged?

bensheldon (approved):
@aried3r this is fantastic. Thank you! I'll re-run CI to get rid of that unrelated timeout and merge. Thank you again! 🙏

Issue #341, opened 2021-08-18
Lockable should accept an explicit keys on class methods too


bensheldon:
I'd like to be able to create an arbitrary advisory lock from the Class; currently, that's only possible from an instance.

e.g. `MyClass.with_advisory_lock ...`

Would help clean up this: https://github.com/bensheldon/good_job/blob/8e1a0f334dc83aae6a04faa521c9c9953c0a9783/lib/good_job/active_job_extensions/concurrency.rb#L25

pandwoter:
Hey, I've tried to dig into this and so far I see two options: 
1. Copy instance variant of the `Lockable#with_advisory_lock` method to class level (this would bloat class level methods in my opinion)
2. We modify a `Lockable.advisory_lock` scope to include only one record if an additional argument was passed (eg key: ...). 
![image](https://user-images.githubusercontent.com/44208683/151324252-345bce4b-8897-4df5-b664-824ee640e8c3.png)
This approach has a downside: we perform all of the hassles with `cte_table` even for one record.

I think both variants is not worth it and it would be better to keep everything as it is right now (creating a useless instance of an `Execution` is cheaper than going with either of these variants)

Let me know what do you think @bensheldon

bensheldon:
@pandwoter Thanks for looking into this. I'm ok with Option 1.

pandwoter:
@bensheldon well, I'll prepare a PR nearest time then 👍 , thanks for the response

PR #499, opened 2022-01-28
Add arbitrary lock on class level too

pandwoter:
https://github.com/bensheldon/good_job/issues/341


spec/lib/good_job/lockable_spec.rb
```diff
@@ -199,6 +199,80 @@
     end
   end
 
+  describe '.advisory_record_lock' do
+    it 'results in a locked record' do
+      model_class.advisory_record_lock!(key: "good_jobs-#{execution.active_job_id}")
```
pandwoter, line 204: @bensheldon, I don't really like adding this prefix here `good_jobs-` and I also don't want to modify the passed key on a class level, because in a use case which you have shown in the issue itself:
https://github.com/bensheldon/good_job/blob/6f0a3fa828e1937184241de8d3d6b9b8d7ce12d2/lib/good_job/active_job_extensions/concurrency.rb#L33-L42
we are passing key with the included prefix

bensheldon: I agree that makes sense. I think it's fine to have `key` operate slightly differently at the class level than the instance level; in my mind this interface is still exploratory.

- Class level `key`: no prefix, generic.
- Instance level `key`: prefixed, instance-specific.

You can use `execution.lockable_key` in this test rather than manually prefixing the string.

bensheldon (commented):
@pandwoter This is a great start. 

I noticed that you created a new method called `.with_advisory_record_lock`. I'd like the `key:` parameter added to the existing `.with_advisory_lock` class method. e.g. "it can be passed a column _or_ a key" It's going to arguably overload it, but I don't want to introduce another method.

And I'm glad that you added additional `.advisory_record_lock` and `advisory_record_unlock` methods. I'd like those all to be renamed `.advisory_lock` so that they are symmetrical with the instance methods.

Also, I think an additional refactor, once the class methods are dialed in, is to have the instance methods delegate to the class methods where possible. 

Thank you!

pandwoter:
>I noticed that you created a new method called .with_advisory_record_lock. I'd like the key: parameter added to the existing .with_advisory_lock class method. e.g. "it can be passed a column or a key" It's going to arguably overload it, but I don't want to introduce another method.

Done! ✅  

> And I'm glad that you added additional .advisory_record_lock and advisory_record_unlock methods. I'd like those all to be renamed .advisory_lock so that they are symmetrical with the instance methods.

Yeah, I named this method like so due that we already have `advisory_lock` scope and we can't define a method with this exact name 😢 (https://github.com/bensheldon/good_job/blob/68fb9603d895b18b588b11285f52d7f83c28679a/lib/good_job/lockable.rb#L40)


> Also, I think an additional refactor, once the class methods are dialed in, is to have the instance methods delegate to the class methods where possible.

Done ✅ 

Any thoughts regarding these scopes @bensheldon? To have exact same interface on a class and instance level we need to do something with it...

bensheldon:
🎉

PR #510, opened 2022-02-02
README: Limiting concurrency - fetch symbol instead of string

BenSto:
Hi,

first of all thanks for the great work, indeed a good job! 

I found a minor mistake in the README regarding the concurrency limitation when using a keyword argument. Currently it says to use `[]` with a string, but the key in the arguments hash is a symbol, so the expression always turns into `nil` . 
Like proposed in this PR it should be a symbol instead of a string and I think it is also better to use `.fetch` instead of `[]`, so it raises immediately a `KeyError` instead of turning silently into `nil`, if the key does not exist.

Feel free to just adapt the README and close this PR.


bensheldon:
@BenSto thanks for flagging this! I just made a few tweaks and will merge. Thank you! 🎉

PR #517, opened 2022-02-09
Document safer setting for retry_on_unhandled_error

### LINKED ISSUE #505: Make default retry behaviour safer
Currently the default for retries is:
> By default, GoodJob will automatically and immediately retry a job when an exception is raised to GoodJob.

The [documentation](https://github.com/bensheldon/good_job#retries) on this behaviour is excellent and quite extensive and does explain nicely how to limit the number of retries using ActiveJob settings.

In our case we configured all Jobs (ApplicationJobs and MailerJobs) to retry a limited number of times. On final retry the exception is not raised but reportet to our exception tracker. We did NOT set `GoodJob.retry_on_unhandled_error = false` as it did not seem necessary (our Unit tests showed the configured number of retries for failing tests).

Unfortunately we use other gems which also enqueue jobs (in this case searchkick). These do NOT inherit from ApplicationJob but from `ActiveJob::Base` (which makes sense for libraries). But without configuring `GoodJob.retry_on_unhandled_error = false` this means these jobs are retried an infinite amount of times (immediately). This we only noticed after a couple million exceptions had reached our tracker (and almost exceeded our quota). After investigating this issue we furthermore had problems replicating the behaviour in test environment (thus even if we had thought of covering this corner with a test it would have not showed up).

I think, in most cases it is not a good idea to infinitely retry jobs (at least not immediately). This at the least introduces high load and in the worst case repeats actions visible to the outside (how about sending out emails). Therefore in my opinion it would be safer to set `GoodJob.retry_on_unhandled_error = false` by default. To enable retries one should then use ActiveJob settings to configure what is needed (maybe change documentation to limited number of retries with Infinite only as a additional note).

If this not what is intended with good_job (safe may also mean safe to not "miss" a job) I would suggest adding a warning in the documentation.

tamaloa:
`retry_on_unhandled_error` current default (`false`) may lead to failing jobs being retried infinitely and without pause thereby at least causing high server load.

This documents the behaviour in README and advised on setting a safe value so hopefully less people are caught by this problem.

fixes #505

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Available configuration options are:
 - `cleanup_interval_seconds` (integer) Number of seconds a Scheduler will wait before cleaning up preserved jobs. Defaults to `nil`. Can also be set with  the environment variable `GOOD_JOB_CLEANUP_INTERVAL_SECONDS`.
 - `logger` ([Rails Logger](https://api.rubyonrails.org/classes/ActiveSupport/Logger.html)) lets you set a custom logger for GoodJob. It should be an instance of a Rails `Logger` (Default: `Rails.logger`).
 - `preserve_job_records` (boolean) keeps job records in your database even after jobs are completed. (Default: `false`)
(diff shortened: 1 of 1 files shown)


bensheldon:
Thank you for the improvement!

PR #518, opened 2022-02-10
Dashboard: update search filters and some small UI updates

multiplegeorges:
This PR updates a few things in the Dashboard UI:
-  Change the search filter UI in the dashboard to be select/dropdown menus. This should scale better for a large amount of jobs and queues. 
- Updated the headline and empty states on the Cron Schedules index and Processes index.
- Adds a simple badge helper for job state badges.
- Adds an empty state for the Executions and Jobs tables.
- Adds specs for the filter changes.

Interface preferences can be very subjective. 
With that in mind, I'd love to get others' input on these changes and have a discussion on the overall direction of the dashboard.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ doc
 .yardoc
 gemfiles/*.gemfile.lock
 tmp/rspec_examples.txt
+.DS_Store
--- a/engine/app/filters/good_job/executions_filter.rb
+++ b/engine/app/filters/good_job/executions_filter.rb
@@ -2,7 +2,7 @@
 module GoodJob
   class ExecutionsFilter < BaseFilter
     def states
-      {
+      @states ||= {
         'finished' => base_query.finished.count,
         'unfinished' => base_query.unfinished.count,
         'running' => base_query.running.count,
--- a/engine/app/filters/good_job/executions_filter.rb
+++ b/engine/app/filters/good_job/executions_filter.rb
@@ -2,7 +2,7 @@
 module GoodJob
   class ExecutionsFilter < BaseFilter
     def states
-      @states ||= {
+      @_states ||= {
         'finished' => base_query.finished.count,
         'unfinished' => base_query.unfinished.count,
         'running' => base_query.running.count,
--- a/engine/app/helpers/good_job/application_helper.rb
+++ b/engine/app/helpers/good_job/application_helper.rb
@@ -5,5 +5,20 @@ module GoodJob
       text = timestamp.future? ? "in #{time_ago_in_words(timestamp)}" : "#{time_ago_in_words(timestamp)} ago"
       tag.time(text, datetime: timestamp, title: timestamp)
     end
+
+    def status_badge(status)
+      classes = case status
+                when :finished
+                  "badge rounded-pill bg-success"
+                when :queued, :scheduled, :retried
+                  "badge rounded-pill bg-secondary"
+                when :running
+                  "badge rounded-pill bg-primary"
+                when :discarded
+                  "badge rounded-pill bg-danger"
+                end
+
+      content_tag :span, status.to_s, class: classes
+    end
   end
 end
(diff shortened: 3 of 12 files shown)


bensheldon:
@multiplegeorges I really like where this is going. I can help get the tests cleaned up, and we could get this merged.

Things I like:
- It handles the situation where there is a lot of job classes and queue names. 
- The badges are nice and I think improving visual scannability is good
- Overall, I really, really appreciate the contributions to the Dashboard. I think it's an ergonomic challenge but also a place for GoodJob to shine. 

Things that bother me:
- (small/easily fixable) there is some accessibility issues with the labels. I think you hacked around to get the visual layout lined up, but I want to make sure the formatting is valid: everything has a label, no empty labels, though we can use aria-hidden and sr-only to accomplish that. 
 - (bigger/more difficult). I don't love putting the States specifically in a dropdown. There is the (currently unrealized) opportunity to be more explanatory about the meaning of the states, because I think that will help people make the most of GoodJob. As links, it's easier to like add an info hover or something. Though I also recognize it's really hard, technically, to mix links and form items, so I'm presenting this less as "here's a solution to use instead" and more of "dropdowns make it harder to achieve this goal I'm thinking about in the future".

Directionally/really big picture stuff:
- Jobs vs Executions needs to be straightened out. The challenge is that I think people's mental model is "jobs", but when debugging a problem (either with a specific job, or overall "why are all my jobs not running as I expected?") it's "executions" where someone should be looking. 
- The chart is a bit of a placeholder. I think it's useful as "what is the overall shape of these logs", but I think there should be a separate page of charts tailored to specific operation questions (e.g. enqueued vs executed, errors, latency, etc.)
- I wonder if the layout of the dashboard should be changed from the filter being on top, to 2-columns (filter on the left, table on the right). That would provide more vertical space for filters, but then challenges me with how it should look on small screens/responsively.

So those are the big picture things to explain. If we can get the tests to pass and work through the WCAG errors I'm good accepting this.

multiplegeorges:
@bensheldon I've fixed the layout in a non-hacky way and improved the accessibility of the labels for the filters.

Thanks for taking a look at the specs, I had them running green locally. Not sure what's up there.

#### Jobs vs Executions

I definitely have the mental model of a Job+retries rather than a series of executions and was initially confused by how the dashboard presents the data. 

I think it could be simplified by moving execution details to the show page of a Job and displaying the job + head execution details in the index table. Previous executions are useful to see for debugging purposes, as you mentioned, but I think any action a user will take will be done on the head execution, so conceptually surfacing that as the main thing a user interacts with is the direction I would go.

#### States

This was also a bit confusing to me due to the Job/Execution split. I initially didn't understand why there were different states available to me depending on which page I was looking. It seemed like all the states of a Job would apply to Executions too.

That said, I agree that making the entire list visible at a glance is much more discoverable and is probably a better UI.

#### Layout

For filterable index pages, I think a two column layout with a vertical set of filters on the left makes a lot of sense. It gives us a lot more flexibility in how we layout the filters. Button-style controls for states would fit nicely, for example.

However, going to this layout will remove some space from the tables. I think that can be mitigated by changing the table rows to be more information dense, so there's no loss there.

#### Graphs

If you are good with consolidating executions onto the show page of a Job, then the root path can become a dashboard page that shows the high-level state of everything: all job states with counts, a couple of graphs, and a snapshot of currently running jobs. 

I'd also like to put an auto-refresh function on that page. I can see myself putting that page up on another monitor and just having it auto-refreshing to keep tabs on things.

bensheldon:
@multiplegeorges can you please check the "maintainer can modify" checkbox on this PR so I can push up a fix for the tests? Thank you!

multiplegeorges:
I believe it is set. Here's what I'm seeing:
Screen Shot 2022-02-15 at 11 32 52 AM

bensheldon:
@multiplegeorges thanks for checking! looks like I had a problem locally with my git. I just successfully pushed up an update.

multiplegeorges:
Looks good to me! Thanks for the help with the tests.

:shipit:

PR #522, opened 2022-02-15
Fix Benchmark job throughput script  

douglara:


bensheldon:
Thanks! Looks like some of my exploration got in there accidentally. 

I'm curious, are you benchmarking?

douglara:
Heheh thanks!

I'm created this feature that implement a fairer way to distribute load between the queues in cases overload using something like the round-robin algorithm and i test this resource using this benchmarking 😀

PR #525, opened 2022-02-18
Added a chapter on how to prepare for production.

### LINKED ISSUE #523: Best practices in deploying and monitoring a queue
Hi there,
first of all, thank you for all the work that everyone has dedicated into this project!!! :heart: 

The reason I'm reaching out is because we're running a bunch of queues with GoodJob on Heroku, and recently we started seeing jobs piling up and not being processed.

After a bit of digging, we found that GooJob would end up out of available connections in the pool (which is totally fair):
> ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.045 seconds); all pooled connections were in use

But what's scary is that the running processes would be reported as healthy by Heroku, even though the performance of the queue degraded (eg. 500+ jobs pending for 10-20m, while normally it would take minutes to process such a load)!

Based on this, my immediate takeaways are:
 1. the workers do not return the connections back into the pool
 2. the monitoring we're having in place is not ok

Anybody else had to go through something similar? 

Should we explicitly try to release any connection back into the pool? Should we use the built-in health-check instead of the generic pid-based health-check?

stas:
Based on the feedback from #523 I tried to put together a chapter with best practices on how to prepare running GoodJob in production.

@bensheldon @reczy would appreciate you taking a look at this. Thank you!

Closes #523


README.md
```diff
@@ -320,6 +321,42 @@ GoodJob.retry_on_unhandled_error = false
...
+   of a job
+ * 1 connection dedicated to the cron scheduler if you're running it
+ * 1 connection dedicated to the cron executor if you're running cron
+
+If you're running the queue along with the Rails process, please adjust
+the pool size taking into account the number of web server threads!
+
+It is important to understand that the queue process won't crash if the
+connections pool is exhausted, instead it will report an exception (eg.
+`ActiveRecord::ConnectionTimeoutError`)!
```
bensheldon, line 349: I'd like this to be isolated inside of the section on database connection pool size. 

Overall, I'm imagining the section you're proposing here is more of a checklist of items that explains _why_ you should configure them for production, but the _how and details_ lives in the section related to that specific feature.

stas: @bensheldon thanks for reviewing this, I followed your suggestions and moved it as the _Database Connections_ sub-section.

As for the reasoning behind it, we basically were wrapping our heads around trying to compile a production-ready setup and still failed to grasp the whole connections usage part. It can be seen as a _checklist_, though to me, I just want to make sure our findings are documented, as I'll definitely be returning to them as I use `good_jobs` in more projects :wink:

bensheldon (approved):
👍

PR #535, opened 2022-03-02
Enable DB table names customization

dimvic:
With this patch it is possible to assign custom table names for `GoodJob::Process` and `GoodJob::ActiveJobJob` and have `good_job` function as usual.

It enables the use of a common database for two or more schedulers while keeping the dashboard functional.

My motive was working on two apps that share a database and both need to run a scheduler and we want to run `good_job` because well, it's too good for our use case :). I believe this is a legit, while ugly, requirement in some projects.

There is no need to complicate things by making the migrations take the custom table names into account for upgrades. It is possible, but the added complexity would possibly make migrations fragile and certainly harder to reason about. After all, anyone who opts to change the table name this way would review migrations coming from updates before running them. Even if not, they would fail and the person would iterate.

The table names have been added without quoting into the SQL code because, well, it's a table name.

I like to believe that the only reason to consider not including this is that future code would need to use table names via model classes instead of hardcoded, which is not necessarily a bad thing.

### DIFF
--- a/engine/app/charts/good_job/scheduled_by_queue_chart.rb
+++ b/engine/app/charts/good_job/scheduled_by_queue_chart.rb
@@ -9,6 +9,7 @@ module GoodJob
     def data
       end_time = Time.current
       start_time = end_time - 1.day
+      table_name = GoodJob::ActiveJobJob.table_name
 
       count_query = Arel.sql(GoodJob::Execution.pg_or_jdbc_query(<<~SQL.squish))
         SELECT *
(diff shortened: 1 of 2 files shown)


bensheldon:
@dimvic fyi, in your application, you should be overriding the table name for `GoodJob::Execution` too. GoodJob's scheduler works off of executions (which is an individual performance of a job). The awkwardly named (till the next major release of GoodJob) ActiveJobJob is used for representing a more ergonomic model in the Dashboard.

This PR is great btw, just wanted to clarify the models.

dimvic:
Thanks! Would you like me to add tests for this? Would only need a couple more introduced in `spec/engine`.

bensheldon:
@dimvic yep, that's exactly right. I don't think this needs tests because it's largely how ActiveRecord works.

It does make me think that I should make `ActiveJobJob` delegate its tablename to `Execution` so that you only have to set that on the Execution itself. Let me merge/release this change to unblock you, and then I'll make another PR for your feedback.

bensheldon:
@dimvic I released [`v2.11.2`](https://github.com/bensheldon/good_job/releases/tag/v2.11.2) with this change 🎉

dimvic:
Thank you very very much. I agree that since the table name is shared between the two it is would be beneficial to have it delegated, but I don't see much more use to it other than making it easier to customise table names. Ping me if you'd like me to do something more about this, I'd be more than happy to.

PR #536, opened 2022-03-03
Delegate `ActiveJobJob.table_name` to `Execution` and prevent it from being directly assignable

bensheldon:
Follow-up to #535


bensheldon:
@dimvic fyi, I'm going to get this in and then cut a release.

PR #542, opened 2022-03-21
Use bundle add instead

glaucocustodio:
As per https://github.com/rubygems/rubygems/pull/5337, we can simplify the steps of adding a gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,19 +65,13 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
 
 ## Set up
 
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
@glaucocustodio thank you for the improvement! I always for get that bundler command 😄 

I made a few tweaks:

- Explained what the command does
- Changed the markdown numbered list back to using `1.` for all elements in the list.  That's a trick to ensure that if one changes elements of the list, it keeps the diff isolated to just the content changes.

Issue #507, opened 2022-01-31
Release GoodJob 3.0


bensheldon:
Changes to GoodJob 3.0:

### Changed default behavior of retries, job preservation, and automatic cleanup. #545 

|                  | 2.0 Defaults | 3.0 Defaults |
|------------------|--------------|--------------|
| Retry Job          | ✅            | ❌            |
| Job Preservation | ❌            | ✅            |
| Cleanup          | ❌            | ✅            |

My goal for GoodJob 3.0 is to change the defaults
 - Enable the automatic `cleanup_interval` so that GoodJob will delete old jobs (and thus resolve my worry about filling up the database)
  - GoodJob _will_ store failed/errored jobs in the database until they're automatically cleaned up. 
 - GoodJob will _not_ retry jobs on failure unless it's been configured in ActiveJob.
 
 Extracted from #495 
 
### Other changes:

- Class name changes:
  - `GoodJob::ActiveJobJob` -> `GoodJob::Job` (`GoodJob::Job` was previously deprecated as an alias of `GoodJob::Execution`)
- Interface changes:
  - `GoodJob::Adapter` no longer can be initialized with custom execution options (queues, max_threads, poll_interval).
- Instrumentation changes
  - `perform_job.good_job` removes `good_job:` parameter, use `execution:` instead.
- Support changes:
  - Drop support for Rails 5.2
  
  
### Tracking

- [x] #641
- [x] #637
- [x] #633 
- [x] #545
- [x] #642

aried3r:
Would there be any tangible maintenance benefit in dropping [EOL PostgreSQL versions](https://endoflife.date/postgresql)? Currently only v9.6 but soon v10 as well.

I only found https://github.com/bensheldon/good_job/commit/d142b4b23c33ec034795be236e5d3fdc4db7d002 in a quick search, which doesn't seem so bad.

strzibny:
Hi, not sure where's the best place to ask, but will there be an option to keep the job status forever? Or you think it's problematic to run GoodJob like that?

PR #545, opened 2022-03-25
By default, preserve job records and automatically them clean up

bensheldon:
Connects to #507


fameoflight:
It make sense to include a default job that clean up old job, then people can put them in cron schedule as needed. 

For this purpose it might make sense to generate models also for GoodJobs at the very least

bensheldon:
@fameoflight when GoodJob 3.0 is released, GoodJob will do the cleaning up of jobs automatically on its own.

Issue #550, opened 2022-04-04
undefined method `relative_time' when include_all_helpers is false


shouichi:
When `include_all_helpers` is set to `false`, only a helper which matches the name of the controller is included.

https://api.rubyonrails.org/classes/ActionController/Helpers.html

GoodJob engine defines some helpers in `ApplicationHelper`, thus only the `ApplicationController` (and its children) can access those helpers. Currently, the base controller is named `BaseController` and those helpers are not available.

Thank you.

bensheldon:
@shouichi thank you for reporting this! There's likely a lot of Rails configuration assumptions being made 😓 

I can make a quick fix for this (explicitly requiring the helpers in the BaseController), unless you'd like to make the PR.

shouichi:
Thanks for the reply!

Sure, we can quickly fix it by explicitly including the helper. I have a small question before making that change. Can we rename BaseController to ApplicationController (which is more common I believe)? That way, we can follow the rails convention.

bensheldon:
Yep, BaseController can be renamed. I don't have much experience with Engine conventions, but change is easy 😊

shouichi:
Thanks for the feedback. I'll submit a PR!

PR #553, opened 2022-04-05
Rename `GoodJob::BaseController` to `GoodJob::ApplicationController`

shouichi:
Problem:

When `config.action_controller.include_all_helpers = false`, helpers are
not available by default which leads to `undefined method 'relative_time'`.
This is because the controller is named `BaseController` and the helper
is named `ApplicationHelper`.

> https://api.rubyonrails.org/classes/ActionController/Helpers.html.
> In previous versions of Rails the controller will include a helper
> which matches the name of the controller, e.g., MyController will
> automatically include MyHelper.

Solution:

Rename `BaseController` to `ApplicationController`. By following the
rails naming convention, necessary helpers are automatically included by
rails.

Close https://github.com/bensheldon/good_job/issues/550.


shouichi:
@bensheldon thanks for the follow-up commits!

PR #558, opened 2022-04-05
Deprecate Adapter configuration of job execution/cron

bensheldon:
Connects to #380 and #421.

This configuration was never documented in the Readme, and I'm not sure if anyone uses it.

My current thinking is that the ActiveJob Adapter should only concern itself with enqueuing jobs. And GoodJob's global configuration will cover executing jobs and cron, etc. This also helps move closer to a goal of GoodJob have a single global supervisor for Schedulers/CronManager, etc. which will make tracking Process lifecycle much easier.


bensheldon:
@Hugo-Hache thank you for the comment! You've made me realize that I have made a mistake: I should _not_ deprecate the `execution_mode:` parameter because (especially for testing), a developer will need to configure how the job is executed e.g. `execution_mode:` is in the context of enqueuing. 

I'll make a PR to undo the deprecation fo that parameter.

PR #562, opened 2022-04-11
FIX: make 'default_url_options' method private

friendlyantz:
Further to [this conversation.](https://github.com/bensheldon/good_job/pull/497/files#r843531378) with @igas

minor fix to move `default_url_options` under private


bensheldon (approved):
@friendlyantz Thank you for hopping on this! Very helpful 🙌

fyi, I removed the `before_action` addition because it isn't needed to private the method.

PR #566, opened 2022-04-18
Remove WIP comments from dashboard

bkeepers:
Thanks for all your work on good_job! I've been using it on a few projects lately and really like it. As the primary maintainer for delayed_job for many years, I have a special appreciation for the care and attention that you and the community have put into it.

I'm hoping to contribute more, but as my first pull request, I suggest removing the WIP messages from the dashboard. It looks like the one has been there for about two years now.


bensheldon (approved):
@bkeepers As a longtime user of Delayed Job myself, I am humbled 😊 

Thank you for your contribution! I _have_ been wondering when the appropriate time is to remove the message... so thank you for pressing on that 🎉

PR #572, opened 2022-04-18
Dashboard UI updates: sticky navbar, statuses as tabs

bkeepers:
I spent a few hours today playing around with a couple things that I would love to see improved with the dashboard.

1. Make navbar sticky instead of footer (which doesn't really contain actionable information)
2. Elevate the status filter to tabs. That is what I personally filter by most often. 
3. Move all filters above the chart since they affect the chart data.

image


bensheldon:
@bkeepers I think this looks fantastic. I think changing the States to Tabs is a big improvement. I like that the States are all immediately visible/discoverable rather than obscured within a dropdown. 

I'm more than happy to merge this as is (not let perfect be the enemy of _better_). I'll share some other thoughts in the [discussion you opened](https://github.com/bensheldon/good_job/discussions/571).

engine/config/locales/en.yml
```diff
@@ -45,9 +45,9 @@ en:
         last_update_html: Last updated <time id="page-updated-at" datetime="%{time}">%{time}</time>
         wording: Remember, you're doing a Good Job too!
       navbar:
-        cron_schedules: Cron Schedules
-        executions: All Executions
-        jobs: All Jobs
+        cron_schedules: Schedules
+        executions: Executions
+        jobs: Jobs
```
bkeepers, line 50: I forgot to call this out in my initial summary. Are you ok with shortening these changes for the navbar?

engine/config/locales/es.yml
```diff
@@ -45,7 +45,7 @@ es:
         last_update_html: Última actualización <time id="page-updated-at" datetime="%{time}">%{time}</time>
         wording: "¡Recuerda, también tú estás haciendo un buen trabajo!"
       navbar:
-        cron_schedules: Tareas Programadas
+        cron_schedules: Horarios
```
bkeepers, line 48: I know a little Spanish, but not in a programming context. Can someone verify this?

bkeepers (commented):
@bensheldon sounds good. This should be ready whenever you are.

engine/config/locales/en.yml
bensheldon: I think "Cron Schedules" should be shortened to "Cron". That's the keyword people are looking for.

engine/config/locales/es.yml
bensheldon: Using "cron", it doesn't need to be translated 😉

bensheldon (approved):
👍 Thank you! 

Lemme know if I can help get the tests passing. I'm eager to get this merged.

bensheldon:
@bkeepers btw, while it's fresh in your mind, I'd love if you have any feedback/suggestions for improving the [Readme's contributor section](https://github.com/bensheldon/good_job#contribute) for getting started.

bkeepers:
> @bkeepers btw, while it's fresh in your mind, I'd love if you have any feedback/suggestions for improving the [Readme's contributor section](https://github.com/bensheldon/good_job#contribute) for getting started.

At first I didn't realize there was a test rails app that generates jobs, so that would have saved me some time getting another app set up and in a state where I could see jobs in various states. Just documenting `bin/test_app server` would be great.

multiplegeorges:
@bkeepers Just updated the gem and saw that the filter dropdowns I added were gone, but the tabs are way better! Looks great, much more discoverable and surfacing the number in each is great for scanning. 

👍

bensheldon:
@multiplegeorges yay! Glad you like it too! 🎉 

@bkeepers I was staring at the tab counts working on #578 and realized that I think the tab counts are incorrect: the numbers should be within the context of the filter, not absolute. If I made that change to have counts relative to the filter, I think the other filter options (Job Class, Queues, Search) should go _above_ the tabs.  Functionally I don't think anything would change though 🤔 

These counts should be based off of `filtered_query` not `base_query`:

https://github.com/bensheldon/good_job/blob/c548fb705e40e47be704fd7726ac923d449b9783/engine/app/filters/good_job/jobs_filter.rb#L4-L13

PR #576, opened 2022-04-21
Remove executions from the dashboard

bkeepers:
Per discussion in https://github.com/bensheldon/good_job/discussions/571#discussioncomment-2588328, this removes Executions from the dashboard.


engine/config/routes.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 GoodJob::Engine.routes.draw do
-  root to: 'executions#index'
+  root to: redirect(path: 'jobs')
```
bkeepers, line 3: Since there is talk of eventually adding a different home page for the dashboard, I made this a redirect to the jobs tab so we don't later break any links pointing to specific filters.

bensheldon (approved):
👍 Thank you!

PR #577, opened 2022-04-21
Dashboard: Use toasts to show notices and alerts

bkeepers:
The dashboard changes in #572 and coming up in #575 are a little awkward with alerts. This switches to using [toasts](https://getbootstrap.com/docs/5.1/components/toasts/) instead of [alerts](https://getbootstrap.com/docs/5.1/components/alerts/).

## Before

image

## After

good_job toast

### DIFF
--- a/engine/app/assets/scripts.js
+++ b/engine/app/assets/scripts.js
@@ -17,6 +17,7 @@ const GoodJob = {
     GoodJob.addListeners();
     GoodJob.pollUpdates();
     GoodJob.renderCharts(true);
+    GoodJob.showToasts();
   },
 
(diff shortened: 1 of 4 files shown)


bensheldon (approved):
Looks good and simple and why I initially wanted to build off Bootstrap 😄 

A downside is that it means we'll either need to enable JS on system tests that check for toasts, or to allow `visible: false` on the capybara matchers. I'm comfortable with the latter. The _nice_ thing to do would be to make a custom matcher like `expect(page).to have_toast("some content")` but that's definitely lagniappe.

PR #580, opened 2022-04-25
Namespaces assets per Rails docs

kylekthompson:
When upgrading to 2.13.1, we were running into this error on our application pages (outside of where the engine was mounted): `Uncaught TypeError: Failed to resolve module specifier "document_ready". Relative references must start with either "/", "./", or "../".`

Per the [Rails Guide on assets in engines](https://guides.rubyonrails.org/engines.html#assets), the assets, like other components, should be namespaced. Oddly, this does seem contradictory with the same section in the docs which later indicate that "the application's asset would take precedence and the engine's one would be ignored".

I've tested this change against our own application by pointing our gemfile at my local copy of this gem with the patch applied, but I'm happy to test in other ways (beyond running the test suite as part of this PR) if you can help guide me through how you usually test these sorts of changes.


engine/app/controllers/good_job/assets_controller.rb
```diff
@@ -35,11 +35,11 @@ def rails_ujs_js
     end
 
     def scripts_js
-      render file: GoodJob::Engine.root.join("app", "assets", "scripts.js")
+      render file: GoodJob::Engine.root.join("app", "assets", "good_job", "scripts.js")
     end
 
     def style_css
-      render file: GoodJob::Engine.root.join("app", "assets", "style.css")
+      render file: GoodJob::Engine.root.join("app", "assets", "good_job", "style.css")
```
kylekthompson, line 42: I went ahead and namespaced these as well, though they weren't conflicting with any of _our_ assets as far as I can tell.

engine/app/controllers/good_job/assets_controller.rb
```diff
@@ -35,11 +35,11 @@ def rails_ujs_js
     end
```
kylekthompson, line 35: I'm not sure if these practices for the engine's assets also apply to the _vendored_ assets. Perhaps these should also be namespaced (see the lines above this comment)?

bensheldon:
@kylekthompson Interesting! I think probably everything (including vendored) should be namespaced. 

Are you using importmaps-rails? That looks like an importmap error?

It's a little odd because GoodJob tries hard to avoid using any asset pipeline. This makes me wonder if I should rename the `assets` directory if there is an implicit assumption that those resources should be pipelined by the host application.

...but to confirm: you're saying that this change fixes the problem for you? That's good enough for me at the moment.

bensheldon:
@kylekthompson I can finish out this PR and get it passing.

bensheldon:
@kylekthompson thank you, again! This has been released: https://github.com/bensheldon/good_job/releases/tag/v2.13.2

kylekthompson:
Thanks so much for taking that over the finish line, sorry to give you more work!

bensheldon:
No prob! If it's blocking you, it's gotta be fixed! 😀

aki77:
@bensheldon 

It seems that `Rails.application.config.assets.paths` is added including namespace.

```
irb(main):001:0> Rails.application.config.assets.paths
=>
[".../app/assets/config",
 ".../vendor/bundle/ruby/3.1.0/gems/good_job-3.4.3/app/assets/good_job",
 ".../vendor/bundle/ruby/3.1.0/gems/blazer-2.6.4/app/assets/fonts",
```

If you are not using an asset pipeline, I think it would be better to rename the directory to something other than `assets`.
For example, `frontend`?

PR #582, opened 2022-04-26
Refactor Dashboard Live Poll javascript

bensheldon:
Follow up to #526.

- Creates javascript class
- Stores state in the dom as much as possible (Stimulus style)
- Replaces `#id=key` replacement with `data-live-poll-region={name}`. This might make name collisions slightly less visible. 
- More exemplary... and makes me wonder if I should simply vendor StimulusJS to be _even more_ exemplary.


bensheldon:
> Thoughts on programmatically toggling off live poll if a user checks a checkbox or clicks on an action button? Maybe even if search bar is dirty?

@reczy ooooh, that's an awesome suggestion. I can add a listener on all form input changes, and use that to toggle off polling. 

Hmmm. Should polling be re-enabled on the next page load?

reczy:
hmm... Good question. My initial gut reaction was "of course" but now that you're making me think about it, I'm not so sure.

Graphs, charts, summary details - in my mind, totally makes sense to have live poll default on. But I'm trying to think of other examples of tabular data that automatically refresh while the user is using it...

I guess tail logs, email inboxes... but both of those pretty much just add data to one end of things. The refreshes here can yield pretty wholesale/jarring changes. If you're trying to read through / analyze the table data or perform some manual action, you definitely don't want an unexpected refresh.

I think I just talked myself out of wanting default on live poll for the table.

Maybe live poll on for charts/graphs/summary data, off for the table? If you want to get fancy, maybe show a notice that says click here to show updates (I guess you could look at the good job table size or check last updated)?

Just ideas - I don't really have any strong opinions or wants here other than hoping to avoid an unexpected table refresh while I'm looking through it.

bensheldon:
@reczy Thanks for thinking through that :-) We agree. 

I think the use-case for current Live Poll functionality is "Put GoodJob's Dashboard on a Wallboard and passively look at it" not "GoodJob's Dashboard is reactively real-time while being operated by a developer". 

Getting to "reactively real-time" will be something different, both from a design perspective (for all of the interruptive reasons you list) and technically (e.g. messaging-pushing and dom-diffing, rather than interval polling and region replacement)

bensheldon:
@reczy I'm going to move your request into an Issue so it's a little easier to track. Should be simple to do now that the refactor makes the change easy 🙌

Issue #551, opened 2022-04-04
Improve the "Gem development" section of README?


shouichi:
It is instructed to run `bin/setup` but it does not always work (depends on the ruby version). For example, with ruby 3.1.x, it does not work because the `matrix` gem was removed. Instead `bundle exec appraisal` should be used.

https://github.com/bensheldon/good_job#gem-development

Also, it might be a good idea to move this entire section to "CONTRIBUTING file".

https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

Thank you.

bensheldon:
@shouichi thank you for the feedback and suggestions! I'll work on improving this section.

shouichi:
Thank you! Feel free to close this when you think it's done :+1:

PR #587, opened 2022-05-01
Include `ruby` version in development Gemfile; update to Ruby 3.0.4

bensheldon:
I'm not sure if this will have any negative repercussions (especially with Appraisals), but it was mentioned as friction in #551.


bensheldon:
The `ruby` declaration really doesn't like only `major.minor` versions without the trailing `.patch`.

PR #598, opened 2022-05-13
Bind probe server to all interfaces

### LINKED ISSUE #592: CLI healtheck only listening on localhost, not reachable for Kubernetes
A cli health check was introduced in #452 after discussion in #403. The readme describes how to add it to Kubernetes. We did exactly that, but the health check does not work.

We can enter the kubernetes-worker-pod and perform `curl localhost:7001`, which works just fine. However, a `curl <ip>:7001` does not work, with a `connection refused`.

Unfortunately, Kubernetes will try to ping the pod via its ip, and is therefore not successful, which prevents the containers from starting, as the health checks are never successful.

Do people use this health check with Kubernetes, and if so, how did they solve this problem? 😆 
If not, how could we allow the binding on all interfaces? We tracked this down to this code:

https://github.com/bensheldon/good_job/blob/9b61ea61d17f7d3d6c1d6d77ea2e2fe368995800/lib/good_job/probe_server.rb#L17-L24

Where we wildly guessed that providing `BindAddress: "0.0.0.0"` in line 20, which will then be passed on to Webrick here:

https://github.com/ruby/webrick/blob/968bc8c4392c0ff8fa8147ca50bc47c74a811886/lib/webrick/server.rb#L111

However, I do not really understand Rack and Webrick, so that is just a first guess, and before experimenting with that (which always involves a deployment...), I first wanted to raise this as an issue here!

Let me know if you need any more information, and best regards 🙂

Timmitry:
Closes #592

The probe server was only reachable under `localhost` and `127.0.0.1`, but not by the local ip address of the system, which causes problem if e.g. the Kubernetes probes try to connect via that ip address.

This PR fixes this by adding a `Host: '0.0.0.0'`-argument to the Rack/Webrick-Server, which means "bind to all interfaces".

### DIFF
--- a/spec/lib/good_job/probe_server_spec.rb
+++ b/spec/lib/good_job/probe_server_spec.rb
@@ -11,12 +11,20 @@ RSpec.describe GoodJob::ProbeServer do
   end
 
   describe '#start' do
-    it 'starts a webrick server' do
+    it 'starts a webrick server that binds to all interfaces' do
       probe_server.start
       wait_until(max: 1) { expect(probe_server).to be_running }
 
-      response = Net::HTTP.get("127.0.0.1", "/", port)
-      expect(response).to eq("OK")
+      ip_addresses = Socket.ip_address_list.select(&:ipv4?).map(&:ip_address)
+      expect(ip_addresses.size).to be >= 2
+      expect(ip_addresses).to include("127.0.0.1")
+
+      aggregate_failures do
+        ip_addresses.each do |ip_address|
+          response = Net::HTTP.get(ip_address, "/", port)
+          expect(response).to eq("OK")
+        end
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/lib/good_job/probe_server_spec.rb
```diff
@@ -11,12 +11,20 @@
...
 
   describe '#start' do
-    it 'starts a webrick server' do
+    it 'starts a webrick server that binds to all interfaces' do
       probe_server.start
       wait_until(max: 1) { expect(probe_server).to be_running }
 
-      response = Net::HTTP.get("127.0.0.1", "/", port)
-      expect(response).to eq("OK")
+      ip_addresses = Socket.ip_address_list.select(&:ipv4?).map(&:ip_address)
```
Timmitry, line 18: Not sure if this is too much magic 🧙 But I wanted to test this somehow...

bensheldon: I like it! 👍

bensheldon (approved):
Thank you!

Issue #605, opened 2022-05-23
dashboard/engine – i18n: Wrong translation scope


Jay-Schneider:
Hi there,

I encountered a problem when updating good_job from `2.14.2` to `2.15.0` regarding the hereby introduced internationalization.

The application I am running is configured to raise an exception instead of falling back to display the key in question.
When I visit the dashboard I receive a `I18n::MissingTranslationData in GoodJob::Jobs#index` stating `translation missing: en.status.scheduled`.

I figured out that this is actually a problem in the gem, where the view refers to `scope: '.status'` while the locale file actually has the translations nested under `good_job.status` (which might be the better choice here imho).
 https://github.com/bensheldon/good_job/blob/265f5bfabb07f91295840a8f3c826a3dcdd2b64e/engine/app/views/good_job/shared/_filter.erb#L10-L13

https://github.com/bensheldon/good_job/blob/265f5bfabb07f91295840a8f3c826a3dcdd2b64e/engine/config/locales/en.yml#L48-L65

So maybe the fix just constists of adjusting to scope to `.goodjob.status`?

Additionally one might consider configuring i18n to raise during tests so that such mistakes are not concealed.

Best regards 👋

bensheldon:
@Jay-Schneider This has been released in [v2.5.1](https://github.com/bensheldon/good_job/releases/tag/v2.15.1)

PR #607, opened 2022-05-23
Fix/i18n status scopes

Jay-Schneider:
As suggested in #605 I enabled the config setting `action_view.raise_on_missing_translations` for the test_app in development and test. This way mistakes as the mentioned mismatch of translation nesting and usage should become apparent.

Also this PR fixes the bug described in #605 by using the correct scope for status translations.

Sadly I had slight issues running the tests locally (with certain rails versions) and so I hope Github Actions is happy anyways.

I can confirm though, that the config adjustment in 5b4ee61 detects the wrong translations. The output is not perfect because the exception is not properly shown in the rspec output but instead the server seems to render a ` "We're sorry, but something went wrong."` 500-page.

### DIFF
--- a/engine/app/assets/good_job/style.css
+++ b/engine/app/assets/good_job/style.css
@@ -34,3 +34,8 @@
 .toast-container {
   z-index: 1;
 }
+
+.translation_missing {
+  background-color: red;
+  color: white;
+}
--- a/engine/app/helpers/good_job/application_helper.rb
+++ b/engine/app/helpers/good_job/application_helper.rb
@@ -41,7 +41,7 @@ module GoodJob
     }.freeze
 
     def status_badge(status)
-      content_tag :span, status_icon(status, class: "text-white") + t(status, scope: '.good_job.status'),
+      content_tag :span, status_icon(status, class: "text-white") + t(status, scope: 'good_job.status'),
                   class: "badge rounded-pill bg-#{STATUS_COLOR.fetch(status)} d-inline-flex gap-2 ps-1 pe-3 align-items-center"
     end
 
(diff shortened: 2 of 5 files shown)


Jay-Schneider:
Okay it seems like the configuration option changes over the rails versions. In Rails 6.x it is called 
```
config.action_view.raise_on_missing_translations = true
```

while in Rails 7 this is not allowed anymore and it should be
```
config.i18n.raise_on_missing_translations = true
```
(which of course in turn won't work for Rails < 6.1). Actually Rails 6.1. accepts both versions and deprecates the first one.

But I feel like we don't want to drop support for Rails 6.0 because of this 😅 

@bensheldon  Since I lack experience in gem development, I am not sure how to handle this properly.
I would probably use a `if Rails.version[0] == "7" ...` but am happy to solve this in another way if you'd give me some pointers 😄

bensheldon:
@Jay-Schneider Yep, you can create a version-based carve-out. Here's an example of where that's been done before: https://github.com/bensheldon/good_job/blob/ee6716b248f8e70e5de899bfa7d827a7d938edd7/spec/lib/good_job/execution_spec.rb#L290

Also: I'd prefer if it only raises in Test environment, not in Development. A common workflow is for someone to add the i18n tokens, and then for me to get the translations; raising in Dev would make that workflow painful.

Jay-Schneider:
> A common workflow is for someone to add the i18n tokens, and then for me to get the translations; raising in Dev would make that workflow painful.

Okay so even the translations for the main language (english in most cases I suppose) will be provided by you and not by the contributing dev? Because if you select `en` locale it is enough to have those present, there is (at this point) no need to have them filled in in all the locale files yet.
So a contributor could just add the english "translations" (aka key-value pairs) to the `en.yml` and code away without exceptions.
Using i18n-tasks to find missing translations and/or adding them is a separate step.

To be clear: I am fine with removing the config for the development environment, I just wanted to make sure there is no misunderstanding regarding this.

PR #620, opened 2022-06-18
Fix `:inline` mode with future behavior to run unscheduled jobs immediately

bensheldon:
Also addresses the inline deprecations in the test suite.

### DIFF
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -65,7 +65,7 @@ module GoodJob
       scheduled_at = timestamp ? Time.zone.at(timestamp) : nil
 
       if execute_inline?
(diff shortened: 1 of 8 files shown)


lib/good_job/adapter.rb
```diff
@@ -65,7 +65,7 @@ def enqueue_at(active_job, timestamp)
       scheduled_at = timestamp ? Time.zone.at(timestamp) : nil
 
       if execute_inline?
-        future_scheduled = (scheduled_at.nil? || scheduled_at > Time.current)
+        future_scheduled = scheduled_at && scheduled_at > Time.current
```
bensheldon, line 68: Oops!

PR #649, opened 2022-06-28
Allow cron entries to be temporarily disabled and re-enabled through the Dashboard

### LINKED ISSUE #540: Cron Schedule jobs add disable action
Hi and thanks for this awesome gem!

My good_job version is `2.11.2`.

I used to use the Sidekiq，and sidekiq has these functions in recurring jobs: `Enqueue Now` and `Disable`

It's very useful for my work.

image

But with good_job, it seems only has the `Enqueue Now` function,  **is there any plans to add `disable` action to schedule jobs?**

image

Thanks.

alex-klepa:
Closes #540.

Screen Shot 2022-07-26 at 1 08 30 PM


bensheldon (commented):
@alex-klepa This is great! Thank you for tackling this. I have some feedback, some maybe more difficult than others. I'd love to get this completed, so let me know if I can help:

1. Instead of having a cron-specific table in the database, I'd like to have a generic `good_job_settings` table, with the following fields:
- `id` (uuid)
- `key` (string, with uniqueness index)
- `value` (jsonb, default null/nil)
- timestamps

2. For this particular feature, at the time that someone toggled on/off, it would create/update a `GoodJob::Setting` record for key=`disabled_cron_entry_keys`, value=`["key1", "key2"]`. And insert or remove cron-key values from that jsonb array. 

(that's the main design change I'd like to see)

3. Instead of a `toggle` action, I'd like to have discrete actions for `enable`, `disable`. That ensures that if 2 separate people (or two separate tabs) are operating in the dashboard, their intent is deliberate.

bensheldon:
@alex-klepa I'm checking in. Can you make those changes I suggested? If not, I'll take over. Thanks again! 🙌🏻

alex-klepa:
Hi @bensheldon, I don't have time at the moment, if you could take over, would be great!

bensheldon:
@alex-klepa no sweat! I really, really appreciate you proposing this and I'll get it over the finish line. Thank you!

PR #672, opened 2022-07-20
Dashboard: improvements to jobs index and show pages

bkeepers:
This PR includes some updates I had started working on months ago and finally got around to merging with master. It's mostly a design refresh for the jobs#index list, but includes some updates to jobs#show for consistency. There is still room for improvement here, but I think this is a good step forward.

<table>
  <tr>
    <th>jobs#index: Before</th><th>jobs#index: After</th>
  </tr>
  <tr>
    <td>
image
</td>
    <td>image</td>
</tr>
<tr>
    <th>jobs#show: Before</th><th>jobs#show: After</th>
</tr>
<tr>
<td>
image
</td>
<td>image</td>
</tr>
</table>


app/views/good_job/jobs/_table.erb
```diff
@@ -1,122 +1,161 @@
...
+    <% if jobs.present? %>
+      <% jobs.each do |job| %>
+        <%= tag.div id: dom_id(job), class: "list-group-item list-group-item-action py-3" do %>
+          <div class="row align-items-center">
+            <div class="col-auto">
+              <%= check_box_tag 'job_ids[]', job.id, false, data: { "checkbox-toggle-each": "job_ids" } %>
+            </div>
+            <div class="col-4">
+              <%= tag.code link_to(job.id, job_path(job), class: "small text-muted text-decoration-none") %>
+              <%= tag.h5 tag.code(job.job_class), class: 'mb-0' %>
```
bkeepers, line 69: FIXME: link to job

app/views/good_job/jobs/_table.erb
```diff
@@ -1,122 +1,161 @@
...
-                  <div class="text-nowrap">
-                    <% if job.status.in? [:scheduled, :retried, :queued] %>
-                      <%= link_to reschedule_job_path(job.id), method: :put, class: "btn btn-sm btn-outline-primary", title: "Reschedule job", data: { confirm: "Confirm reschedule", disable: true } do %>
-                        <%= render_icon "skip_forward" %>
-                      <% end %>
-                    <% else %>
-                      <button class="btn btn-sm btn-outline-secondary" disabled><%= render_icon "skip_forward" %></button>
+    <% if jobs.present? %>
+      <% jobs.each do |job| %>
+        <%= tag.div id: dom_id(job), class: "list-group-item list-group-item-action py-3" do %>
```
bkeepers, line 62: FIXME: clicking on row should either link to job or toggle checkbox 🤔

lib/models/good_job/job.rb
```diff
@@ -8,6 +8,7 @@ module GoodJob
   class Job < BaseRecord
     include Filterable
     include Lockable
+    include Reportable
```
bensheldon, line 11: Maybe a quibble, but I don't think the latency/runtime stats quite make sense for a `Job` record because a Job is a container for multiple Executions. 

Thinking about my generic recommendations for optimizing queues, I think the number to report is the total latency for the entire job e.g. the time from when the first execution was _expected_ to run, to the time that the job eventually finished.

bkeepers: You're right. I'll remove them from the table for now, and…

I'd love to discuss what I was trying to get at by adding them: I'm looking for a snapshot of queue latency and runtime as operational metrics which indicate if there are enough workers/resources. This was just an easy way to quickly scan them to get a rough overview. It makes a lot more sense to include them on an overview dashboard rather than in the jobs list. If I remember right, resque-web had a dashboard that showed current latency for each queue.

Also, I think it makes sense for a `Job` to sum the queue latency and runtime for each `Execution`, but I'm not sure total wall clock time from start to finish is a helpful operational metric since it includes intentional backoff. Either way, this total probably only makes sense on individual job pages and an overview dashboard.

PR #675, opened 2022-07-22
Add cron_enabled attribute to good_job and pass it to process current state

saksham-jain:
- Add cron_enabled attribute to good_job.
- Set this attribute on initialization using enable_cron key in configuration.
- Add new key cron_enabled to process current_state and use GoodJob.cron_enabled to set this key.

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -69,14 +69,6 @@ module GoodJob
   #   @return [Boolean, nil]
   mattr_accessor :retry_on_unhandled_error, default: false
 
-  # @!attribute [rw] cron_enabled
-  #   @!scope class
-  #   Use this to know if cron is enabled for the process (default: +false+).
-  #   If +true+, it means cron is enabled.
-  #   If +false+, it means cron disabled.
-  #   @return [Boolean, nil]
-  mattr_accessor :cron_enabled, default: false
-
   # @!attribute [rw] on_thread_error
   #   @!scope class
   #   This callable will be called when an exception reaches GoodJob (default: +nil+).
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -69,6 +69,14 @@ module GoodJob
   #   @return [Boolean, nil]
   mattr_accessor :retry_on_unhandled_error, default: false
 
+  # @!attribute [rw] cron_enabled
+  #   @!scope class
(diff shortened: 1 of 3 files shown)


lib/good_job.rb
```diff
@@ -59,6 +59,14 @@ module GoodJob
   #   @return [Boolean, nil]
   mattr_accessor :retry_on_unhandled_error, default: false
 
+  # @!attribute [rw] cron_enabled
+  #   @!scope class
+  #   Use this to know if cron is enabled for the process (default: +false+).
+  #   If +true+, it means cron is enabled.
+  #   If +false+, it means cron disabled.
+  #   @return [Boolean, nil]
+  mattr_accessor :cron_enabled, default: false
```
bensheldon, line 68: I'm trying to move away from these global accessors.

Take a look at `GoodJob::Configuration` and see if you can instantiate one of those instead to get the configuration.

bensheldon (changes requested):

lib/good_job.rb
saksham-jain: Should we directly add new key in process's current_state without creating this global accessor, like this -> `cron_enabled: Rails.application.config.good_job[:enable_cron]`

bensheldon: I'm working on #681, which should allow us to use `GoodJob.configuration.enable_cron?`. I'll let you know when it gets merged.

bensheldon: It's been merged. If you merge main or rebase, you can use `GoodJob.configuration`

saksham-jain: Okay, This made things easy.

bensheldon:
@saksham-jain could you please fix up the lint warning? This looks good 👍🏻

bensheldon (approved):
👍🏻 thank you!

bensheldon:
@saksham-jain Thank you for the PR. I really appreciate this! Are you interested in further issues?

Just having worked through with this, I see a few jumping off points for improvement:

- Rename that configuration option to be `Configuration#cron_enabled?` so it's consistent and the less awkward construction.
- Could you go through `Configuration` and propose other configuration options you think would be meaningful to expose on the process state for debugging in the dashboard?

saksham-jain:
Thanks, @bensheldon. I will check these points.

PR #679, opened 2022-07-25
Show basename of proctitle

bkeepers:
This is just a quick tweak to the processes page to shorter the `proctitle` to just the basename.

## Before

image

## After

image


app/views/good_job/processes/index.html.erb
```diff
@@ -27,7 +27,7 @@
           <div class="col">
             <code class="font-monospace">
```
bensheldon, line 28: What would you think of also putting the full proctitle in a `title` attribute; similar to how relative timestamps also have the full timestamp available.

bkeepers: 👍 good call. Added in 2914ecc

PR #695, opened 2022-08-10
Add additional details to Concurrency Control explanation

bensheldon:
Connects to https://github.com/bensheldon/good_job/discussions/694


shouichi (commented):
If I understand it correctly, we should emphasize that the primary mechanism of concurrency control is the number of threads/processes. Then explain that using `good_job_control_concurrency_with` as an additional barrier is a good idea. Make sense?

bensheldon:
@shouichi Hmmm. It really depends on your concurrency needs. Someone may just want to prevent two jobs with the same arguments from being performed at the same time.

Issue #684, opened 2022-07-29
Feature: dynamic options for good_job_control_concurrency_with


baka-san:
Currently only `key` can be set dynamically for `good_job_control_concurrency_with`, i.e. `perform_limit` and `enqueue_limit` cannot be. I have a generic job which is being queued by dozens of cron jobs, each which have different concurrency requirements.

So I wonder if it is/would be possible to do something like:

```ruby
  good_job_control_concurrency_with(
    key: -> { "CrawlPostJob_#{arguments.last[:spider_class]}" }, # this currently works
    perform_limit: -> { arguments.last[:concurrency] }, # this doesn't work at the moment
  )
111

baka-san:
Sorry, I haven't read the contributing section in your README yet, but I dug into this a bit. I manually tested with the dashboard and enqueuing jobs from the console and it seems to allow dynamic settings as I asked about in my initial post.

```ruby
# lib/good_job/active_job_extensions/concurrency.rb

# frozen_string_literal: true
module GoodJob
  module ActiveJobExtensions
    module Concurrency
      extend ActiveSupport::Concern

      class ConcurrencyExceededError < StandardError
        def backtrace
          [] # suppress backtrace
        end
      end

      module Prepends
        def deserialize(job_data)
          super
          self.good_job_concurrency_key = job_data['good_job_concurrency_key']
          self.good_job_concurrency_total_limit = job_data['good_job_concurrency_total_limit']
          self.good_job_concurrency_enqueue_limit = job_data['good_job_concurrency_enqueue_limit']
          self.good_job_concurrency_perform_limit = job_data['good_job_concurrency_perform_limit']
        end
      end

      included do
        prepend Prepends

        class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
        attr_writer :good_job_concurrency_key, :good_job_concurrency_total_limit,
          :good_job_concurrency_enqueue_limit, :good_job_concurrency_perform_limit

        around_enqueue do |job, block|
          # Don't attempt to enforce concurrency limits with other queue adapters.
          next(block.call) unless job.class.queue_adapter.is_a?(GoodJob::Adapter)

          # Always allow jobs to be retried because the current job's execution will complete momentarily
          next(block.call) if CurrentThread.active_job_id == job.job_id

          enqueue_limit = job.good_job_concurrency_enqueue_limit
          total_limit = job.good_job_concurrency_total_limit

[...]

PR #696, opened 2022-08-12
Allow concurrency limits to be configured dynamically with lambda/proc

baka-san:
Adds the ability to set dynamic concurrency limits (i.e. `total_limit`, `enqueue_limit`, and `perform_limit`) via a proc similar to how a key can be set via a proc. Discussed in issue https://github.com/bensheldon/good_job/issues/684.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,15 +406,18 @@ class MyJob < ApplicationJob
 
   good_job_control_concurrency_with(
     # Maximum number of unfinished jobs to allow with the concurrency key
+    # Can be an Integer or Lambda/Proc that is invoked in the context of the job
     total_limit: 1,
 
     # Or, if more control is needed:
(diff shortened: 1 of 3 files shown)


baka-san:
@bensheldon Thanks for solving that issue. Sorry about that.

I'm having some problems setting up the local environment, hence running the test suite. Xcode and rbtrace aren't playing together nicely. Which OS are you using?

bensheldon:
@baka-san I'm using MacOS 12.5. I should clean up the Gemfile's development section because rbtrace isn't necessary, only nice to have for debugging deadlocks.

bensheldon:
@baka-san I just removed some [unnecessary dev dependencies](#700). Let me know if that fixes things for you. This PR looks good right now so I'll merge it when CI passes 👍🏻

Issue #697, opened 2022-08-12
Jobs enqueued via dashboard ignores app default_locale


esasse:
My app uses `pt-BR` as default_locale and `en` just as fallback. When I enqueue a job from the cron entries list on the dashboard, it runs using `en` instead of `pt-BR`. 

<img width="233" alt="image" src="https://user-images.githubusercontent.com/148989/184384847-3cbdba72-d0e2-4124-9a67-147aca0d3199.png">

I'm using async mode and I suppose dashboard is changing app's default_locale. How can I avoid this?

Thanks!

bensheldon:
Oh that's interesting! 

ActiveJob implicitly includes the context's locale when [enqueuing jobs](https://edgeguides.rubyonrails.org/active_job_basics.html#internationalization). 

I think the fix should be as simple as wrapping cron's enqueue with an explicit `I18n.with_locale(I18n.default_locale) do ...`:

https://github.com/bensheldon/good_job/blob/65dc794ea6dd2207db949fbb444bf4cb4e9f4d1d/app/models/good_job/cron_entry.rb#L109

Would you be able to submit a PR to fix that?

esasse:
> Would you be able to submit a PR to fix that?

I'll try! 😀

PR #698, opened 2022-08-12
Enqueues jobs with I18n default locale

esasse:
Closes https://github.com/bensheldon/good_job/issues/697


bensheldon:
@esasse thank you! This is great 🎉

PR #702, opened 2022-08-15
Describe pessimistic usecases

shouichi:
See https://github.com/bensheldon/good_job/pull/695#pullrequestreview-1070520303.


bensheldon:
Thank you!

PR #707, opened 2022-08-20
Keep locale param when submitting dashboard filter

aki77:
Behavior before modification

### DIFF
--- a/app/views/good_job/shared/_filter.erb
+++ b/app/views/good_job/shared/_filter.erb
@@ -5,6 +5,7 @@
     <%= form_with(url: "", method: :get, local: true, id: "filter_form", class: "container-fluid") do |form| %>
       <%= hidden_field_tag :poll, params[:poll] %>
       <%= hidden_field_tag :state, params[:state] %>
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
Thanks for catching that! 🙌🏻 

btw, what are you using to see the DOM changes highlighted in red?

bensheldon:
Released: https://github.com/bensheldon/good_job/releases/tag/v3.4.4

aki77:
Thanks!
The DOM highlighting is due to the following changes.
2b584df6941cbe4ad6231ec09f33236c3e8d63fa

Whenever good_job is used from a language other than the one provided by default, it is always highlighted.
If highlighting is for good_job developers, I think it would be better if only developers explicitly enable highlighting.

bensheldon:
@aki77 oh 🤦🏻 I had forgotten that's GoodJob code. Yes, that highlight should be removed; the intent was to help with internationalizing GoodJob, but I wasn't thinking about the side effects.

I'll remove it.

PR #711, opened 2022-09-11
Sentry integration Docs

remy727:
`sentry-raven` was outdated and `Raven` was renamed to `Sentry`.


bensheldon:
@remy727 thank you! I just (finally!) migrated my personal projects from `sentry-raven` to `sentry-ruby` so this is good timing 😄

Issue #691, opened 2022-08-08
Implement batch functionality


bensheldon:
Connects to discussion in #417; please keep discussion in that issue. This issue is focused on implementation.

```ruby
batch = GoodJob::Batch.new
batch.callback_class = "MyCallBackJob" # => Results in `MyCallbackJob.perform_later(batch)`
batch.callback_queue_name = "default"
batch.callback_priority = 0
batch.properties = { user: current_user } # additional state that is serialized/deserialized using ActiveJob/GlobalId
batch.add do
  MyJob.perform_later
  YourJob.perform_later
end

more_job = MoreJob.perform_later
batch.add(more_job)

batch.enqueue # jobs may be performed but the callback won't be triggered till `Batch#enqueue` is called

# Or more succinctly
GoodJob::Batch.enqueue(on_finish: "MyCallbackJob", user: current_user) do |batch|
  MyJob.perform_later
  YourJob.perform_later
  MoreJob.perform_later
end

batch.jobs # => All of the GoodJob::Jobs in the batch
batch.enqueued_at # => When the batch is first enqueued
batch.discarded_at # => When the _first_ job is discarded
batch.completed_at # => When all jobs are completed successfully or discarded

# When the _first_ job within the batch is discarded:
#   MyCallbackJob.perform_later(batch)
# When _all_ jobs within the batch are completed successfully or discarded:
#   MyCallbackJob.perform_later(batch)
```

## Work
- [x] Add a callback after an execution performed _and unlocked_ #706
- [x] Implement a batch enqueue that can bulk insert multiple jobs at the same time
- [ ] ...

Issue #417, opened 2021-10-10
Support for Batches?


mollerhoj:
Considering migrationg from Sidekiq Pro: Does this gem support some equivalent of sidekiq batches?

Thats is, can I set a job to run once a particular set of other jobs have finished?

mollerhoj:
Sure. We're extracting data from a REST API where every resource in a collection (of say 20.000 items) has it's own endpoint. The pipeline is as follows:

- 1) Fetch a list of items
- 2) Send a request to each items endpoint.
- 3) Register that the job is done.

This is a common and embarrassingly parallel task thats completely IO bound, and thus much faster to do with many jobs rather than a single one.

Our main issue with sidekiq is the reliance on redis - It's much simpler to keep everything in postgres rather than having to sync the knowledge (of what job has run/is running/is connected to a particular model etc) of our job system db and our CRUD db manually.

bensheldon:
That's helpful. I'm wondering if this sort of interface would be sufficient?

```ruby
class FanOutJob < ApplicationJob
  include GoodJob::ActiveJobExtensions::Batch

  after_batch do |id:, key:|
    id  # => a UUID; all the jobs enqueued in the same batch will have the same UUID
    key # => a String that is entirely within your control to set data, serialize/etc. 
        #    to pass some data/state to the completion of the batched set of jobs

    CrudModel.find_by(key).do_something_after_the_batch_finished
  end

  def perform(*args)
    # do the work
  end
end

batch_job_args = [args_for_job_1, args_for_job_2, args_for_job_3]
FanOutJob.perform_later_in_batch(batch_job_args, key: the_crud_model.id) # => returns the batch_id, maybe?
```

In the implementation I'm imagining, the only state that would be passed forward to the batch completion would be an arbitrary key (the String), and a UUID (which would be used by the backend to trigger the `after_batch` callback once all the jobs with the same UUID had completed).

Some questions I have:

- How would you handle errors? e.g. if one of the jobs errors, should that job/error be passed to the `after_batch` handler? Should successful jobs be passed too? Is the ActiveJob instance objects enough (they won't have a record of the instance objects), or would you want the GoodJob::Execution itself (which is kinda-sorta a private API).
- Is a single "after every job in the batch finishes/errors" enough? Are there other lifecycle events that would be required?


[...]

PR #712, opened 2022-09-12
Implement `GoodJob::Batch`

bensheldon:
Connects to #417, #691.

- [x] Add description field
- [x] Add Dashboard view
- [x] Add cleanup for batches
- [x] Rename batch `params` to `properties`
- [x] Make migration-less patch-release compatible


README.md
```diff
@@ -396,6 +397,74 @@ end
...
+    # When these jobs have finished, it will enqueue your `MyBatchCallbackJob.perform_later(batch, options)`
+    class MyBatchCallbackJob < ApplicationJob
+      def perform(batch, options = {})
+        # Do something with the <GoodJob::Batch> instance; options is a hash is reserved for future use
+        batch.properties[:user] # => <User id: 1, ...>
+      end
+    end
+    ```
+
+- Jobs can be added to an existing batch. Jobs in a batch are enqueued and performed immediately, though the final callback job will not be enqueued until `GoodJob::Batch#enqueue` is called.
```
mollerhoj, line 421: I find these instructions unclear. To me, it looks like the code below contains a race condition. What happens if all the jobs finishes before `enqueue` is called?.

On another note: What happens if you call `enqueue` twice on the same batch?
What happens if you call `enqueue` on a batch without any jobs?

README.md
```diff
@@ -396,6 +397,74 @@ end
...
+    batch = GoodJob::Batch.add do
+      10.times { MyJob.perform_later }
+    end
+    batch.add do
+      10.times { OtherJob.perform_later }
+    end
+    batch.enqueue(MyBatchCallbackJob, age: 42)
+    ```
+
+- If you need to access the batch within a model, include [`GoodJob::ActiveJobExtensions::Batches`](lib/good_job/active_job_extensions/batches.rb) in your job class:
```
mollerhoj, line 433: Am I allowed to add jobs dynamically to the batch? (If yes, let's add an example in the documentation, this is extremely useful.)

mollerhoj (commented):
I have not have time yet to go over the code in detail, but I've written some questions that popped up when reading the documentation.

All in all, this looks very, very good to me. Nice API, seems like a pretty clean implementation, and an absolute KILLER feature.

I think the documentation would be even better with an example of a complex workflow (using batches to run both serial and parallel jobs in succession.) E.g, show how one would implement this workflow:


workflow

README.md
bensheldon: Great callout! I will cover that in the documentation. Briefly:

- the callback job won't be enabled until `#enqueue` is called on the batch. 
- calling `#enqueue` on a batch without any jobs will immediately enqueue the callback job
- `#enqueue` can be called multiple times, which will re-enable the callback job to be enqueued
- `#add` will add jobs to the batch without triggering an enqueue

Do you think this kind of example makes sense for creating a multistage batch?

```ruby
class MyBatchJob < ApplicationJob
  def perform(batch, options = {})
    if batch.properties[:stage] == 1
      batch.properties[:stage] = 2
      batch.enqueue do
        10.times { OtherJob.perform_later}
      end
    elsif batch.properties[:stage] == 2
      batch.properties[:stage] = 3
      batch.enqueue do
        OtherJob.perform_later
      end
    else
      # all done
    end 
  end
end

GoodJob::Batch.enqueue(MyBatchJob, stage: 1)
```

PR #719, opened 2022-10-07
When not preserving job records, ensure all prior executions are deleted after successful retry

ylansegal:
When records are configured to be preserved on `on_unhandled_error`, and later retried, the previous executions stay in the database and become "orphaned". I believe they should be deleted on a successful retry. This would bring it inline with the behavior for destroying a failed job (eg. via the UI) which results in previous executions also being deleted.

This is my first contribution to this repo. I took a stab at writing a spec that exposes the issue and a fix via a ActiveRecord callback. I am open to suggestions for improvement.

Thank you for making `good_job` available and open source!

### DIFF
--- a/app/models/good_job/execution.rb
+++ b/app/models/good_job/execution.rb
@@ -66,7 +66,7 @@ module GoodJob
     end
 
     belongs_to :job, class_name: 'GoodJob::Job', foreign_key: 'active_job_id', primary_key: 'active_job_id', optional: true, inverse_of: :executions
-    after_destroy -> { self.class.active_job_id(active_job_id).delete_all }
+    after_destroy -> { self.class.active_job_id(active_job_id).delete_all }, if: -> { @_destroy_job }
 
     # Get Jobs with given ActiveJob ID
     # @!method active_job_id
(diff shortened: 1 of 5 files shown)


bensheldon:
@ylansegal thank you for opening up this PR. I think you've discovered a bug 🎉 

I need to think about this a bit more. I don't think the fix here is quite right because I think if a a job is retried (which will generate a new execution) this PR will delete that new execution, which isn't desired.

Here is where the `Execution#destroy` is called after performing the job (the subsequent execution gets created/enqueued as part of the ActiveJob performance): 

https://github.com/bensheldon/good_job/blob/e13915e9c356ffa6df5eabd44dfac5b723ddaaaf/app/models/good_job/execution.rb#L294-L306

I'm thinking this logic needs to be unrolled to more clearly separate "This is what happened with the job/execution (e.g. succeeded, retried, discarded/handled, unhandled" and then within that have the logic of "and based on the configuration, this is what should happen with the record". I might push up into your branch unless you want to dig into that more.

ylansegal:
> I need to think about this a bit more. I don't think the fix here is quite right because I think if a a job is retried (which will generate a new execution) this PR will delete that new execution, which isn't desired.
> 

I definitely had not considered retries fully. My thinking was that failed jobs are preserved, which would not trigger the callback. That seems to be an assumption on how I have configured GoodJob in my project.  

> I'm thinking this logic needs to be unrolled to more clearly separate "This is what happened with the job/execution (e.g. succeeded, retried, discarded/handled, unhandled" and then within that have the logic of "and based on the configuration, this is what should happen with the record". I might push up into your branch unless you want to dig into that more.

Feel free to push to that branch! You definitely have more knowledge of the internals.

bensheldon:
Released in [v3.4.8](https://github.com/bensheldon/good_job/releases/tag/v3.4.8) 🎉  Thank you for the fix 🙌🏻

bensheldon:
@ylansegal sorry, I need to revert this because of an unexpected issue in #728. I will figure out a better implementation and get this behavior re-added.

ylansegal:
Thank you @bensheldon

PR #727, opened 2022-10-20
Add configurable limit (`queue_select_limit`) when querying candidate jobs

mitchellhenke:
Partner to #726 and the other part of #720

This piece is also very helpful in scaling the number of processes/threads, particularly when scheduling a bunch of jobs or during times when the queue is backed up.

I wasn't sure on the best place to pass the configuration so I tried to start at the top which ends up passing down through a few layers. I _think_ I understand the reasoning behind configuring it to the maximum number of potential threads across all processes, but I may not have gotten the documentation in the config quite right.

One thing that strikes me as an area to potentially improve is how to know when you may need to bump it up if you're unaware of this limit. Having a high default is good, but if someone is running 10,000 threads of jobs and aren't aware of this configuration, they could be adding a bunch of jobs and only processing 1,000 at a time. I don't have any good ideas at the moment though.

Thanks again for the effort you've put into GoodJob, it's really wonderful and easy to use!

### DIFF
--- a/app/models/good_job/lockable.rb
+++ b/app/models/good_job/lockable.rb
@@ -37,11 +37,14 @@ module GoodJob
       # @param function [String, Symbol]  Postgres Advisory Lock function name to use
       # @return [ActiveRecord::Relation]
       #   A relation selecting only the records that were locked.
-      scope :advisory_lock, (lambda do |column: _advisory_lockable_column, function: advisory_lockable_function|
+      scope :advisory_lock, (lambda do |column: _advisory_lockable_column, function: advisory_lockable_function, queue_select_limit: nil|
         original_query = self
 
         cte_table = Arel::Table.new(:rows)
         cte_query = original_query.select(primary_key, column).except(:limit)
+        if queue_select_limit
+          cte_query = cte_query.limit(queue_select_limit)
+        end
         cte_type = if supports_cte_materialization_specifiers?
                      'MATERIALIZED'
                    else
@@ -154,10 +157,10 @@ module GoodJob
       #   MyLockableRecord.order(created_at: :asc).limit(2).with_advisory_lock do |record|
       #     do_something_with record
       #   end
-      def with_advisory_lock(column: _advisory_lockable_column, function: advisory_lockable_function, unlock_session: false)
+      def with_advisory_lock(column: _advisory_lockable_column, function: advisory_lockable_function, unlock_session: false, queue_select_limit: nil)
         raise ArgumentError, "Must provide a block" unless block_given?
 
-        records = advisory_lock(column: column, function: function).to_a
+        records = advisory_lock(column: column, function: function, queue_select_limit: queue_select_limit).to_a
 
         begin
           unscoped { yield(records) }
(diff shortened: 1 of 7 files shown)


lib/good_job/configuration.rb
```diff
@@ -208,6 +210,20 @@ def cron_entries
...
+    # The number of queued jobs to select when polling for a job to run.
+    # This limit is intended to avoid locking a large number of rows when selecting eligible jobs
+    # from the queue. This value should be higher than the total number of threads across all good_job
+    # processes to ensure a thread can retrieve an eligible and unlocked job.
+    # @return [Integer, nil]
+    def queue_select_limit
+      (
+        options[:queue_select_limit ] ||
+          rails_config[:queue_select_limit ] ||
+          env['QUEUE_SELECT_LIMIT'] ||
```
bensheldon, line 222:
```suggestion
          env['GOOD_JOB_QUEUE_SELECT_LIMIT'] ||
```

app/models/good_job/lockable.rb
```diff
@@ -37,11 +37,14 @@ module Lockable
       # @param function [String, Symbol]  Postgres Advisory Lock function name to use
       # @return [ActiveRecord::Relation]
       #   A relation selecting only the records that were locked.
-      scope :advisory_lock, (lambda do |column: _advisory_lockable_column, function: advisory_lockable_function|
+      scope :advisory_lock, (lambda do |column: _advisory_lockable_column, function: advisory_lockable_function, queue_select_limit: nil|
```
bensheldon, line 40: Within `Lockable` I think the option should be named `select_limit` (rather than `queue_`) because I'm _intending_ `Lockable` to be extractable and usage/purpose neutral.

btw, nice naming of `queue_select_limit` generally. I was sweating the naming of this option 👍🏻

bensheldon (changes requested):
This is beautiful 😍  Thank you! I left some comments and here's some big questions:

- What if the value defaulted to `nil`, and then was configurable when someone noticed the performance problem at scale? I'm wondering if you considered that, and I'm basically looking for reassurance that setting a default value is the right call.
- I _think_ 1,000 is a good number. The total number of connections the database allows is likely the analogous bottleneck (I know people can highly provision RDS, but most PaaS stuff I've seen is more like 250 database connections). I'm asking around of the people who I know GoodJob is using at large scale (funnily enough, I asked @zachmargolis not realizing that you are working on the same project 😊 )

app/models/good_job/lockable.rb
mitchellhenke: Thank you! I renamed the variable/argument in that file.

mitchellhenke:
>     * What if the value defaulted to `nil`, and then was configurable when someone noticed the performance problem at scale? I'm wondering if you considered that, and I'm basically looking for reassurance that setting a default value is the right call.

I was kind of leaning towards that as well since it would maintain the existing behavior without potentially surprising anyone in a less intuitive way. I'd rather the performance degrade slowly as the queue grows than not select an eligible job from the default limit on the query.

>     * I _think_ 1,000 is a good number. The total number of connections the database allows is likely the analogous bottleneck (I know people can highly provision RDS, but most PaaS stuff I've seen is more like 250 database connections). I'm asking around of the people who I know GoodJob is using at large scale (funnily enough, I asked @zachmargolis not realizing that you are working on the same project 😊 )

Yeah, 1000 would be pretty safe (maybe on the high side?). With the index, the query is good and fetching 1,000 rows is hopefully not _too_ slow.

That said, I think I'm going to drop the default to make it `nil`.

Issue #728, opened 2022-10-20
3.4.8 release breaks job retrying


morgoth:
In this release https://github.com/bensheldon/good_job/compare/v3.4.7...v3.4.8 it is no longer possible to retry job.

We're having code like this:
```ruby
class EveryMinuteJob < ApplicationJob
  after_perform do
    enqueue wait: 1.minute
  end

  def perform
    # do something here
  end
end
```
This is no longer working, as the job somehow is not enqueued again

morgoth:
Thank you

PR #730, opened 2022-10-20
Redo: When not preserving job records, ensure all prior executions are deleted after successful retry

bensheldon:
This undoes the reversion of #719 in #729, informed by #728.

Newly, this covers the case when `ActiveJob::Base#reenqueue` is called directly during job execution (not via `retry_job`), which I had not considered initially.


app/models/good_job/execution.rb
```diff
@@ -296,13 +297,14 @@ def perform
       job_error = result.handled_error || result.unhandled_error
       self.error = [job_error.class, ERROR_MESSAGE_SEPARATOR, job_error.message].join if job_error
 
+      reenqueued = result.retried? || retried_good_job_id.present?
```
bensheldon, line 300: This is the significant change in this PR from the previously reverted PR.

Issue #710, opened 2022-09-10
[help wanted]: Preserve job logs


bensheldon:
Preserving logs during job execution has been requested in the past. 

I wanted to sketch out what this feature might look like, *with the hope that someone else would implement it*. I've broken it down by checkpoint because I don't want to review a huge PR all at once, but rather build up the functionality so it meets my standards for quality and maintainability. I'm happy to zoom pair on it too if you drive. I'm open to change too. 

## The big picture
Have an ActiveJob Extension that stores logs in the database for later review. 

- Logs will be stored in their own table
- A single execution can have multiple log records. Ideally it would flush a buffer to the database at some interval. So more than a single log record, but not so many that each log or line goes to its own record. 
- Logs should be displayable in the dashboard. And ideally be inspectable while the job is running. 
- The intention is to improve troubleshooting/debugging. Not for compliance or as a replacement for a true log drain. 

## Checkpoint 1

Goal: A ruby harness that tees job logs to a buffer object. 

- There exists an includeable concern that (can) contains all of the behavior to be mixed into a job class.  
- When a job runs, the existing ActiveJob logger is teed into a new in-memory buffer instance. The buffer will need to be threadsafe (queue object?), but would be nice if it can be accessed globally (eg a globally accessible object that has internal name-spacing for individual executions)
- At the end of a run, the buffer instance can be inspected to see it contains all the logs. 
- Nice to have: allow a lower level log level (eg debug) just for the logs being retained. 

## Checkpoint 2

Goal: Drain logs to the database in a database-connection-efficient manner. 

- Exists an object that creates a Background thread (concurrent future) that waits on the buffer object. Can look like the Notifier. 
- Unlike the Notifier, the thread doesn't need to check out a dedicated DB connection. 
- In the thread, it should wait on the buffer, then do an efficient insert to drain the buffer, sleep a configurable amount of time, then repeat. 
- At most, writing logs should use 1 database connection per process.   
- Logs should be cleaned up automatically, just like old job records. 

## Checkpoint 3

Goal: surface logs in the dashboard

- However you want, using Bootstrap and existing dashboard components. 
- Nice to have: Poll for live log updates.

bensheldon:
Here's a simple/naive test case for Checkpoint 1:

```ruby
# spec/lib/good_job/active_job_extensions/logging_spec.rb

# frozen_string_literal: true
require 'rails_helper'

RSpec.describe GoodJob::ActiveJobExtensions::Logging do
  before do
[...]

PR #733, opened 2022-10-27
ActiveJob extension for preserving logs

NeimadTL:
This draft allows me to have an idea of what the logging module will look like. It consists of the execution id, the logger info and the logger debug. Those three data, could be added into a model and would have a `belong_to` relationship with the Execution model on one end; and on the other end, the Excecution would have a`has_one` relationship with this model. And then, at each execution, a record of this model would be saved in the database.

Connects to #710


bensheldon:
@NeimadTL this looks great! I pushed up some changes so that the logs are split and stored in a Concurrent::Array. I think this hits Checkpoint 1 👍🏻 

Do you want to try for Checkpoint 2?

NeimadTL:
> @NeimadTL this looks great! I pushed up some changes so that the logs are split and stored in a Concurrent::Array. I think this hits Checkpoint 1 👍🏻
> 
> Do you want to try for Checkpoint 2?

Thanks @bensheldon. I'm definitely  up for checkpoint 2. Let's discuss tomorrow.

lib/good_job/active_job_extensions/logging.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+module GoodJob
+  module ActiveJobExtensions
+    module Logging
+      extend ActiveSupport::Concern
+
+      included do
+        around_perform do |job, block|
+          original_logger = job.logger
+          job.logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(LogDevice.new(job)).extend(ActiveSupport::Logger.broadcast(original_logger)))
```
NeimadTL, line 10: @bensheldon , the `around_perform` hook above makes sense to me but what do L10 and L11 do exactly please ?

lib/good_job/active_job_extensions/logging.rb
```diff
@@ -0,0 +1,38 @@
...
+          block.call
+        ensure
+          job.logger = original_logger
+        end
+      end
+
+      class LogDevice
+        cattr_accessor :logs, default: Concurrent::Array.new
```
NeimadTL, line 18: @bensheldon is the `c` in `cattr_accessor` to indicate that this is class attribute

lib/good_job/active_job_extensions/logging.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        def initialize(job)
+          @job = job
+        end
+
+        def write(message)
+          self.class.logs << [@job.provider_job_id, message.strip]
+        end
+
+        def close
```
NeimadTL, line 28: I don't think #close and #reopen are overrode method but I might be wrong here. Are those methods mandatory or you think we'll be using them in the future ?

NeimadTL (commented):
@bensheldon, just left some questions. Could you clarify them to me please?

lib/good_job/active_job_extensions/logging.rb
bensheldon: This is creating a new logger that outputs to the custom LogDevice, and then teeing the output via `broadcast` so that the output still goes to the old logger too.

I could probably break this up into multiple lines. e.g.

```ruby
custom_logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(LogDevice.new(job))
custom.logger.extend(ActiveSupport::Logger.broadcast(original_logger))

ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(LogDevice.new(job)).extend(ActiveSupport::Logger.broadcast(original_logger)))
```

lib/good_job/active_job_extensions/logging.rb
bensheldon: Yes, it's a class attribute. By living on the class it becomes a singleton that can be access globally.

lib/good_job/active_job_extensions/logging.rb
bensheldon: I'm trying to match the interface of Ruby's `LogDevice`:

https://github.com/ruby/ruby/blob/9001e53e68d282493f513ed67824e4014fd01d57/lib/logger/log_device.rb

That said, I wonder if it would be better to inherit from LogDevice to be clear that I'm trying to match/override the interface, rather than just be an implicit duck-type.

bensheldon:
@NeimadTL thank you for the awesome Code Review! You raised some great questions about how this could be clearer. I can make those changes :thumbsup:

bensheldon:
I'm going to close this for now and revisit in the future.

PR #754, opened 2022-12-03
Redirect to Jobs path after destroy

BilalBudhani:
Currently when a job is deleted the controller redirects back to that job page which throws not found error. 

This PR redirects back to the jobs index page which is the expected behaviour.

### DIFF
--- a/app/controllers/good_job/jobs_controller.rb
+++ b/app/controllers/good_job/jobs_controller.rb
@@ -80,7 +80,7 @@ module GoodJob
     def destroy
       @job = Job.find(params[:id])
       @job.destroy_job
-      redirect_to jobs_path, notice: "Job has been destroyed"
+      redirect_to jobs_path, notice: "Job has been destroyed" # rubocop:disable Rails/I18nLocaleTexts
     end
 
     private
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
Thank you for noticing this and fixing it! Really helpful 💖

That linter error can be disabled with a line comment (aside: maybe Rubocop should also warn on `redirect_back`).

bensheldon:
I made a PR to rubocop-rails: https://github.com/rubocop/rubocop-rails/pull/880

app/controllers/good_job/jobs_controller.rb
```diff
@@ -80,7 +80,7 @@ def retry
     def destroy
       @job = Job.find(params[:id])
       @job.destroy_job
-      redirect_back(fallback_location: jobs_path, notice: "Job has been destroyed")
+      redirect_to jobs_path, notice: "Job has been destroyed"
```
bensheldon, line 83:
```suggestion
      redirect_to jobs_path, notice: "Job has been destroyed" # rubocop:disable Rails/I18nLocaleTexts
```

BilalBudhani:
> That linter error can be disabled with a line comment (aside: maybe Rubocop should also warn on redirect_back).

> I made a PR to rubocop-rails: https://github.com/rubocop/rubocop-rails/pull/880

ah! right. Thanks for helping me with this. I'm in love with this gem ❤️

bensheldon:
@BilalBudhani thank you again! Released: https://github.com/bensheldon/good_job/releases/tag/v3.6.2

PR #756, opened 2022-12-04
Remove support for Ruby 2.5 and JRuby 9.2; reactivate appraisal tests for Rails HEAD

bensheldon:
- I'm unable to build either Ruby 2.5 or JRuby 9.2 locally 😞  
- Ruby 2.5 has been EOL since March 31, 2021. 
- JRuby 9.2 is no longer listed on the [JRuby Downloads](https://www.jruby.org/download) page.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,7 +67,7 @@ jobs:
       # Validate Development Environment
       - name: bin/setup
         run: bin/rails db:test:prepare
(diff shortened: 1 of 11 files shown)


morgoth:
I'm not familiar with jRuby ecosystem, but I would say yes.
Looks like Rails doesn't run CI tests at all for jRuby.

PR #768, opened 2022-12-11
Fix disabling of interval cleanups

zarqman:
This PR allows `cleanup_interval_jobs` and `cleanup_interval_seconds` to each be disabled by setting to `nil` or `false`. Since v3.0.0 (see #545), both would fallback to the default values.

Since disabling these worked in v2.x and `CleanupTracker` still allows `nil`, it seems this was an accidental regression. Tests added accordingly.

The ENV variable equivalents may also be disabled by setting to an empty string:
`env GOOD_JOB_CLEANUP_INTERVAL_JOBS=  good_job ...`
This was already true, but is now tested.

The primary use-case is to allow one cleanup interval to be used, with the other disabled (that's the case here).

If both are set to `nil`, then auto cleanup is disabled. In light of the recent change to always preserve cron jobs (#767), that could be bad. However, setting both to `nil` does allow intentionally running cleanup externally, which may be a useful configuration for some. Accordingly, no attempt has been made to prevent both being disabled.

Finally, tagged on an update to ensure macOS on Intel can use precompiled versions of gems like `nokogiri`. I don't know if you prefer having such housekeeping tasks as separate PRs or not. Happy to split off if needed.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -236,6 +236,8 @@ GEM
       racc (~> 1.4)
     nokogiri (1.13.10-java)
       racc (~> 1.4)
+    nokogiri (1.13.10-x86_64-darwin)
+      racc (~> 1.4)
     nokogiri (1.13.10-x86_64-linux)
       racc (~> 1.4)
     octokit (4.25.1)
@@ -399,6 +401,7 @@ PLATFORMS
   ruby
   universal-java-11
   universal-java-17
+  x86_64-darwin
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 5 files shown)


bensheldon (approved):
@zarqman thank you for finding this! Yep, this looks correct.

I do somewhat regret allowing `nil` to be a proper configuration value. I think ideally it would be `false` (for Ruby) or `-1` (for ENV). But it is what is is for now.

Thank you again! 🙏🏻

zarqman:
@bensheldon, funny you mention it... I started with just `false` as well, and then also realized the legacy nature of `nil`. Arguably `0` could disable as well.

I suppose it could be added to the v4 changes list for when the time comes? (Or if you'd like me to prepare such a PR now and leave it sitting for v4, I'd be happy to do that while it's all fresh in my mind.)

And, thanks for the quick merge!

bensheldon:
@zarqman if you wanted to add a deprecation notice, that would be ideal for me remembering to change it in v4. 

I think the CLI should use `-1` to disable it, and Ruby config should use `false`, then the configuration object would coerce all that to either false or a positive integer. 

Tell me I'm wrong here. But I think `0` would be a valid value to imply "run a clean up cycle after every execution".

zarqman:
In `cleanup_interval_jobs`, `0` is currently equivalent to `1`.
In `cleanup_interval_seconds`, `0` is currently equivalent to setting `cleanup_interval_jobs = 1`, which is, well, odd.

For the sake of clear, intuitive behavior, I think `0` and `-1` should both disable. If `1` is wanted, it should be set expressly. Setting `SOME_ENV=0` is also a pretty common way to indicate disabled.

PR coming shortly.

Issue #770, opened 2022-12-12
Unable to discard failed jobs which crashed with `ActiveJob::DeserializationError`


nickcampbell18:
Hi team, not 100% sure whether this issue is caused by Rails or GoodJob, but here goes!

### Steps to reproduce

1. Enqueue a job using a Global ID, then delete the record it pointed to. The job cannot run now because it would raise an `ActiveJob::DeserializationError` when a GoodJob executor picks it up
2. User visits the GoodJob dashboard, finds the job and clicks "Discard" (this could also happen if there was a failed execution first, but it's irrelevant to trigger this problem)

### Expected behaviour

User can discard the job, even though the arguments cannot deserialize

### Actual behaviour

The GoodJob web UI crashes with an exception, user gets a HTTP 500 error page. It looks like [calling `execution.active_job`](https://github.com/bensheldon/good_job/blob/b84ef5a0cdf5a5ef756238424edfaa3e5106cb81/app/models/good_job/job.rb#L209) is what triggers Rails to attempt to deserialize the arguments, and crash. Here's the relevant parts of the backtrace:

```
lib/active_record/relation/finder_methods.rb:381 raise_record_not_found_exception!
...
lib/active_record/relation/finder_methods.rb:69 find
lib/active_record/querying.rb:22 find
lib/active_record/core.rb:271 find
...
lib/global_id/locator.rb:17 locate
lib/active_job/arguments.rb:148 deserialize_global_id
...
lib/active_job/arguments.rb:43 deserialize
lib/active_job/core.rb:190 deserialize_arguments
lib/active_job/core.rb:180 deserialize_arguments_if_needed
app/models/good_job/execution.rb:321 block in active_job
app/models/good_job/execution.rb:320 active_job
app/models/good_job/job.rb:209 block in discard_job
app/models/good_job/lockable.rb:309 with_advisory_lock
app/models/good_job/job.rb:207 discard_job
app/controllers/good_job/jobs_controller.rb:64 discard
```

I'm aware this is possibly a niche issue. Rails generates the commented-out code `discard_on ActiveJob::DeserializationError` in `ApplicationRecord`, to encourage you to consider this by default, but in my application we had never enabled it globally and I would guess others also haven't, so I think there's good cause for fixing it.

[...]

PR #771, opened 2022-12-12
Ignore ActiveJob::DeserializationError when discarding jobs

nickcampbell18:
Previously, discarding a job would attempt to deserialize the job arguments and crash. We load the job to attempt to run the `#instrument` method, but unfortunately this means that the job could not be discarded.

This commit adds a flag to `Execution#active_job`, `ignore_deserialization_errors`, which says "I want to load this job but I don't mind if the arguments cannot be deserialized".

There are only two places currently where this method is called, and the other one (`retry_job`) will still throw errors when attempting to retry a job in this state. But now, when discarding, the errors are completely ignored and the job can be discarded.

Closes https://github.com/bensheldon/good_job/issues/770

### DIFF
--- a/app/models/good_job/execution.rb
+++ b/app/models/good_job/execution.rb
@@ -316,9 +316,16 @@ module GoodJob
       self.class.unscoped.unfinished.owns_advisory_locked.exists?(id: id)
     end
 
-    def active_job
+    # Build an ActiveJob instance and deserialize the arguments, using `#active_job_data`.
+    #
+    # @param ignore_deserialization_errors [Boolean]
+    #   Whether to ignore ActiveJob::DeserializationError when deserializing the arguments.
+    #   This is most useful if you aren't planning to use the arguments directly.
+    def active_job(ignore_deserialization_errors: false)
       ActiveJob::Base.deserialize(active_job_data).tap do |aj|
         aj.send(:deserialize_arguments_if_needed)
+      rescue ActiveJob::DeserializationError
+        raise unless ignore_deserialization_errors
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/models/good_job/execution.rb
```diff
@@ -316,10 +316,21 @@ def executable?
...
+    def active_job(ignore_deserialization_errors: false)
+      instance = ActiveJob::Base.deserialize(active_job_data)
+
+      begin
+        instance.send(:deserialize_arguments_if_needed)
+      rescue ActiveJob::DeserializationError
+        raise unless ignore_deserialization_errors
       end
+
+      instance
```
nickcampbell18, line 333: I deviated slightly from the suggestion on the issue here, I think it makes sense to return an instance of the job even if the arguments couldn't be deserialized (although this could introduce further bugs downstream if you tried to do other things with it....)

bensheldon: Yes! In strong agreement 👍🏻

bensheldon: Looking a bit closer at that lint error, I realized that the `tap` in my original example was ensuring that the activejob instance is returned regardless of what happens in the rescue

nickcampbell18: 🤦 you're right of course - I've reverted to that style.

bensheldon (approved):
👍🏻 Looks good to me! 

Exploring it a little more I realized that we didn't cover when the initial deserealization fails (e.g. the job class constant is missing). I also see some other general cleanup, like these methods maybe should be moved over from the `Execution` to the `Job` class (probably overlooked during a previous refactor)

This PR moves us forward though and addresses your issue, so 🥇 

...and I'll do the cleanup _later_ 😄

bensheldon:
Released! https://github.com/bensheldon/good_job/releases/tag/v3.7.2

nickcampbell18:
Thanks @bensheldon!

PR #789, opened 2023-01-06
Avoid including unnecessary pg_locks references when counting jobs

mkrfowler:
We've been observing some performance limitations with the GoodJob dashboard. I believe this to be largely a consequence of performing live counts on an insufficiently-pruned table relative to the available database resource, but I noted that the dashboard count queries include a join against `pg_locks`. In the event that there are a substantial number of e.g. `succeeded` jobs, this actually gets quite expensive, and isn't necessary to satisfy counts.

This PR is a bit of an opportunistic approach to avoid that cost.


bensheldon:
@mkrfowler thank you! The Dashboard is very unoptimized so I appreciate the help. I want to poke at it locally for a bit, but I think it'll be good 👍

PR #791, opened 2023-01-06
Ensure Concurrency Keys are string-like and return a better error when they cannot be cast to a string

### LINKED ISSUE #784: Concurrency key doesn't handle Hash: TypeError (can't cast Hash)
As the title says. The job arguments are serialized by ActiveJob to something the database understands but the same is not true for the concurrency key.

```rb
class ExampleJob < ApplicationJob
  good_job_control_concurrency_with(total_limit: 1, key: -> { arguments.first })

  def perform(arg)
  end
end

Job.new.perform({a: "test"})
# => All good
Job.perform_later({a: "test"})
# => Failed enqueuing Job to GoodJob(queue): TypeError (can't cast Hash)
```

Not sure if what I was trying to do should even be supported, but the error message I get as a result of doing this is not very clear and lead me to believe that the issues was with the parameters I was passing to the job.

Edit: Updated the issue description after figuring out what was actually going on.

Earlopain:
Closes #784 

If you pass something like a Hash or Array as the concurrency key, the resulting stack trace doesn't make it very clear what you did wrong:

```
   TypeError:
     can't cast Hash
   # ./app/models/good_job/lockable.rb:196:in `advisory_lock_key'
   # ./lib/good_job/active_job_extensions/concurrency.rb:51:in `block (2 levels) in <module:Concurrency>'
   # ./spec/lib/good_job/active_job_extensions/concurrency_spec.rb:157:in `block (4 levels) in <top (required)>'
   # ./spec/support/reset_rails_queue_adapter.rb:12:in `block (2 levels) in <top (required)>'
   # ./spec/support/reset_good_job.rb:19:in `block (2 levels) in <top (required)>'
   # ./spec/support/database_cleaner.rb:14:in `block (3 levels) in <top (required)>'
   # /usr/local/rvm/gems/default/gems/database_cleaner-core-2.0.1/lib/database_cleaner/strategy.rb:30:in `cleaning'
   # /usr/local/rvm/gems/default/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:34:in `block (2 levels) in cleaning'
   # /usr/local/rvm/gems/default/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaners.rb:35:in `cleaning'
   # ./spec/support/database_cleaner.rb:13:in `block (2 levels) in <top (required)>'
```

This made me believe that I'm not allowed to pass a Hash to the job itself. Now the error is much clearer:

```
 ArgumentError:
       Concurrency key must be a String, was Hash
```

This is technically a breaking change. Some types like Number, Time and Symbol were being cast to a String by rails automatically, this now raises an error. The readme says that the concurrency key must be a String, but it would be best for this to go in the next major version if accepted.

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -120,7 +120,7 @@ module GoodJob
         return if key.blank?
 
         key = key.respond_to?(:call) ? instance_exec(&key) : key
-        raise TypeError, "#{key.class} is not a valid type of concurrency key" unless VALID_TYPES.any? { |type| key.is_a?(type) }
+        raise TypeError, "Concurrency key must be a String; was a #{key.class}" unless VALID_TYPES.any? { |type| key.is_a?(type) }
 
         key
       end
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -4,6 +4,8 @@ module GoodJob
     module Concurrency
       extend ActiveSupport::Concern
 
+      VALID_TYPES = [String, Symbol, Numeric, Date, Time, TrueClass, FalseClass, NilClass].freeze
+
       class ConcurrencyExceededError < StandardError
         def backtrace
           [] # suppress backtrace
@@ -30,6 +32,11 @@ module GoodJob
           # Always allow jobs to be retried because the current job's execution will complete momentarily
           next(block.call) if CurrentThread.active_job_id == job.job_id
 
+          # Only generate the concurrency key on the initial enqueue in case it is dynamic
+          job.good_job_concurrency_key ||= job._good_job_concurrency_key
+          key = job.good_job_concurrency_key
+          next(block.call) if key.blank?
+
           enqueue_limit = job.class.good_job_concurrency_config[:enqueue_limit]
           enqueue_limit = instance_exec(&enqueue_limit) if enqueue_limit.respond_to?(:call)
           enqueue_limit = nil unless enqueue_limit.present? && (0...Float::INFINITY).cover?(enqueue_limit)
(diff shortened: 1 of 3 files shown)


lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -117,11 +117,10 @@ def _good_job_concurrency_key
...
 
-        if key.respond_to? :call
-          instance_exec(&key)
-        else
-          key
-        end
+        key = instance_exec(&key) if key.respond_to? :call
+        raise ArgumentError, "Concurrency key must be a String, was #{key.class}" unless key.is_a?(String)
+
+        key
```
bensheldon, line 123:
```suggestion
        key.to_s
```

bensheldon (commented):
@Earlopain Thinking about this, it's too much of a breaking change: objects that previously could be coerced to a string will start raising exceptions.

Simply doing the coercion would avoid the issue you had of the value being cast, rather than coerced into a string (though still leaves open it not being understood that a complex object is being stringified)

Earlopain:
> @Earlopain Thinking about this, it's too much of a breaking change: objects that previously could be coerced to a string will start raising exceptions.
> 
> Simply doing the coercion would avoid the issue you had of the value failing to cast (though still leaves open it not being understood that a complex object is being stringified)

I don't think doing `to_s` would be a good idea considering that you can pass model instances to jobs. It trades one issue for another:

```rb
class ExampleJob < ApplicationJob
  good_job_control_concurrency_with(perform_limit: 1, key: -> { arguments.first })

  def perform(arg)
    sleep 60
  end
end

ExampleJob.perform_later(ExampleModel.first)
# key: "#<ExampleModel:0x00007f892b76bea0>"
ExampleJob.perform_later(ExampleModel.first)
# key: "#<ExampleModel:0x00007f892d66ee38>"
```
Both jobs would execute in parallel.

I agree on it being a breaking change. I have no issues with this being put on the backlog for v4, whenever that might be.

bensheldon:
Hmmm. I don't want to _require_ it to be `is_a?(String)` because there are plenty of other classes that I think should be acceptable: String, Symbol, Numeric, Date/Time....

...though maybe that isn't as infinite as I feared and it would be fine to typecheck those.

I also wonder if it would be possible to catch the bad cast exception and raise an easier to understand error.

Earlopain:
I like the approach of just catching the error. I changed the commit to do just that.

PR #795, opened 2023-01-12
Add French translation

francois-ferrandis:
Hope that helps!

Note: it seems that some labels are missing from the I18n files, would you be interested in a PR that extracts them?

Thank you for this great gem ! :star_struck: :gem:

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -1,0 +1,83 @@
+---
+fr:
+  datetime:
+    distance_in_words:
+      about_x_hours:
+        one: environ 1 heure
+        other: environ %{count} heures
+      about_x_months:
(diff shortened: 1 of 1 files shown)


bensheldon:
@francois-ferrandis this is fantastic. Thank you!

And yes, GoodJob is insufficiently tokenized for i18n. I would be grateful if you wanted to extract Dashboard strings that are missing from translation files in a separate PR.

francois-ferrandis:
That's great! I'll make a PR when I have some free time. :man_dancing: 

About this PR, are we waiting for a confirmation from a french-speaking person before merging?

bensheldon:
@francois-ferrandis oops, I just got distracted. Thank you!

Issue #731, opened 2022-10-23
Potential locking race condition when using cron scheduler across multiple processes


mitchellhenke:
I've typically only seen this under load-testing scenarios with many job processes, but Postgres was spending long time in `pg_advisory_lock`. I was only able to find `pg_advisory_lock` being used in GoodJob for cron-scheduled jobs ([here](https://github.com/bensheldon/good_job/blob/main/lib/good_job/active_job_extensions/concurrency.rb#L51) and [here](https://github.com/bensheldon/good_job/blob/main/lib/good_job/active_job_extensions/concurrency.rb#L94)). It looked like the other locking calls used the non-blocking `pg_try_advisory_lock`.

The time spent getting the lock seems to primarily be from the `around_enqueue` lock when a new cron job is created and then all of the processes try to lock the cron key to see if it is enqueued, and they end up waiting for one another.

The best way I've been able to reproduce it locally is have a frequently running cron-job (I chose every minute), and then running `good_job start` for the sample application in a handful of terminal panes (6 was sufficient on my machine). I have a branch/commit [here](https://github.com/mitchellhenke/good_job/commit/ab33b679ecea3899259bcbf3fb44f5e1a15b4856). With 6 running good job processes, the average time spent calling `pg_advisory_lock` was around 33ms according to `pg_stat_statements`, and I suspect it grows as you add processes/cron jobs.

I've only used GoodJob with cron jobs where the concurrency `total_limit` is 1, so I'm not sure on a good way to help when using limits larger than that, but I have a small idea for the `1` case at least. Would it be feasible to switch to `pg_try_advisory_lock` if the `limit` is 1? I think there's no need to check the current concurrency for performing/enqueuing if someone else has the lock, so it should be fine for the current process to let it go? I've pushed up a branch with that idea [here](https://github.com/bensheldon/good_job/compare/main...mitchellhenke:try-lock-for-limit-1-cron-jobs?expand=1).

I'm a bit out of my depth on concurrency stuff like this, so do feel free to let me know if I'm off-base. Thank you again!

bensheldon:
@mitchellhenke thanks for digging into this! It's complicated 😄 

First, do you _need_ to be using concurrency controls with `total_limit: 1`? There was a point in GoodJob's history where apps had to use concurrency controls to prevent multiple cron-processes from enqueuing; that stopped being necessary in #423 (v2.5.0).  

[...]

PR #810, opened 2023-01-24
Add global enable_listen_notify configuration to disable both notify and listen

mitchellhenke:
Based on discussion in https://github.com/bensheldon/good_job/issues/731#issuecomment-1369108297, this is a draft implementation of allowing the LISTEN/NOTIFY functionality to be configurable

I need to add some tests, but I wasn't sure on my initial approach.


lib/good_job/notifier.rb
```diff
@@ -165,6 +166,8 @@ def create_executor
     end
 
     def listen(delay: 0)
+      return unless @enable_listening
```
bensheldon, line 169: This can't be an early return. Creating the future is necessary because the `Notifier` is a bit more of a general purpose Reactor now with more responsibilities than _just_ LISTENing e.g. this background thread creates/manages the `Process` record.

It will be messier but I think the `unless @enable_listening` needs to go within the Future/thread:

1. Calling `connection.execute("LISTEN #{CHANNEL}")` (and the ActiveSupport Notification wrapping it (though we still need the callback trigger
2. Within `#wait_for_notify` having `!@enable_listening` that cause the logic to fall down to the `sleep WAIT_INTERVAL` path
3. Calling `connection.execute("UNLISTEN *")`

bensheldon (commented):
@mitchellhenke I think all the pieces are here, but I left some comments about the particulars.

Also, this led me to create #811, which is very out of scope for your change, but I wanted to document the other observations I had too, in case it provides some more context for your changes.

lib/good_job/configuration.rb
```diff
@@ -331,6 +333,15 @@ def probe_port
         env['GOOD_JOB_PROBE_PORT']
     end
 
+    def enable_listen_notify
+      ActiveModel::Type::Boolean.new.cast(
+        options[:enable_listen_notify] ||
+        rails_config[:enable_listen_notify] ||
+        env['GOOD_JOB_ENABLE_LISTEN_NOTIFY'] ||
+        DEFAULT_ENABLE_LISTEN_NOTIFY
+      )
```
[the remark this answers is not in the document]
mitchellhenke: ahhhhh, yep, thank you

lib/good_job/notifier.rb
mitchellhenke: Thanks for explaining, I've added a couple commits to do the above and also tried avoiding checking out the connection. I'm not sure if it's right, and I couldn't figure out how to test it. `notifier.connection` was always `nil` whether or not `enable_listening` was enabled.

bensheldon:
@mitchellhenke I just pushed up a few tweaks to this. I'm happy with where it ended up, but I'd love your eyes on it one more time before I merge it.

Also, I just wanted to document that this will mean that the Probe Server's `status/connected` will return a HTTP error status when listening is not used. Are you using that functionality?

mitchellhenke:
> @mitchellhenke I just pushed up a few tweaks to this. I'm happy with where it ended up, but I'd love your eyes on it one more time before I merge it.

Thanks! It makes sense to me.

> Also, I just wanted to document that this will mean that the Probe Server's `status/connected` will return a HTTP error status when listening is not used. Are you using that functionality?

Ohhh, I wouldn't want to break that. I'm not using it currently, but might in the future. I think I understand this a bit better now. My primary goal at the outset was to limit the herd of queue selection queries that happens when enqueuing a lot of jobs (not necessarily in bulk).

It could be a lot simpler and still accomplish the goal by adding a `notify` argument to the `enqueue` (and `enqueue_at`?) methods to specify whether to skip or send the `NOTIFY`. Whether or not that should be globally configurable feels a bit debatable, but I'd be happy to open a PR that at least adds the argument.

bensheldon:
> It could be a lot simpler and still accomplish the goal by adding a `notify` argument to the `enqueue` (and `enqueue_at`?) methods to specify whether to skip or send the `NOTIFY`. 

I agree on the simpler aspect of getting this closer to the individual job, but we aren't able to (or I'm very reluctant to) modify the Adapter interface.

I'm thinking instead that it could be done as a job extension so that you could do something like this:

```ruby
MyJob.set(good_job_notify: false).perform_later
```

Does that more closely match your needs? I think I can whip that up pretty quick.

bensheldon:
@mitchellhenke I whipped up that extension 😄 : https://github.com/bensheldon/good_job/pull/814

mitchellhenke:
> @mitchellhenke I whipped up that extension 😄 : #814

Thank you! Is it possible to opt in to not notifying for all jobs? The limitation on retries has me worry about a cascade of failures becoming a cascade of notifications on retry.

bensheldon:
@mitchellhenke I updated #814 to preserve the setting across retries.

bensheldon:
@mitchellhenke I updated the PR to also be serialized/preserved into the job data so that it will persist across job retries.

bensheldon:
Whoops, didn't mean to close this.

mitchellhenke:
@bensheldon yeah, that looks good to me!

PR #814, opened 2023-01-28
Create ActiveJob extension to disable sending a NOTIFY on enqueue and retry

bensheldon:
Replacement of #810 at more granular level.


lib/good_job/active_job_extensions/notify.rb
```diff
@@ -0,0 +1,44 @@
...
+module GoodJob
+  module ActiveJobExtensions
+    # Allows configuring whether GoodJob should emit a NOTIFY event when a job is enqueued.
+    # Configuration will apply to initial enqueue and subsequent retries.
+    #
+    # @example
+    #   # Include the concern to your job class:
+    #   class MyJob < ApplicationJob
+    #     include GoodJob::ActiveJobExtensions::Notify
+    #   end
```
mitchellhenke, line 11: Does this mean if I don't include it, that'll be sufficient to not notify? Is this considered a breaking change?

bensheldon: No, this is more like "enabled by default, include this extension to change the default value to something else"

hmm, that does make me wonder if I should change it to `NotifyConfiguration` or something.

bensheldon: `NotifyOptions` it is!

PR #820, opened 2023-01-30
Abort enqueue when the concurrency limit is reached

TAGraves:
Curious to get your feedback on this!

We implemented concurrency keys in some of our jobs for the first time today, and I noticed that when a job isn't enqueued because the concurrency limit is reached, Rails still logs a line like
```
Enqueued TestJob (Job ID: 39afef71-ad97-4a53-adad-2ba05636d440) to (default) with arguments: {:name=>"Alice"}
```

This happens because ActiveJob [logs as long as enqueue didn't raise an error or get aborted](https://github.com/rails/rails/blob/fee14347348ad99ff6260225aaebc8809a983fa9/activejob/lib/active_job/log_subscriber.rb#L8-L25). Maybe the best fix actually lies there although it's not totally clear to me how best to patch that.

Anyway, this PR uses `before_enqueue` instead of `around_enqueue` and throws `:abort` when a job won't be enqueued. That instead causes Rails to log
```
Failed enqueuing TestJob to (default), a before_enqueue callback halted the enqueuing execution.
```

which is more accurate, although I think `Failed` is scary language there. So I also added a log like
```
Aborting enqueue of TestJob (Job ID: 39afef71-ad97-4a53-adad-2ba05636d440) because the concurrency key 'Alice' has reached its limit of 1 job
```

to help make it more clear what's going on.

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -25,40 +25,13 @@ module GoodJob
         class_attribute :good_job_concurrency_config, instance_accessor: false, default: {}
         attr_writer :good_job_concurrency_key
 
-        around_enqueue do |job, block|
-          # Don't attempt to enforce concurrency limits with other queue adapters.
-          next(block.call) unless job.class.queue_adapter.is_a?(GoodJob::Adapter)
-
-          # Always allow jobs to be retried because the current job's execution will complete momentarily
-          next(block.call) if CurrentThread.active_job_id == job.job_id
-
-          # Only generate the concurrency key on the initial enqueue in case it is dynamic
-          job.good_job_concurrency_key ||= job._good_job_concurrency_key
-          key = job.good_job_concurrency_key
-          next(block.call) if key.blank?
-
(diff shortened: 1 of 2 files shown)


TAGraves:
@bensheldon That all makes sense to me. I think I should have some time tonight to work on this, so I'll give it a try then!

TAGraves:
@bensheldon I took a look at using `ActiveJob::EnqueueError`, but it looks to me like that only gets rescued from the job's `enqueue` method, not from the enqueue callback hooks. (At least when I raised it where I'm currently throwing an exception, it got bubbled up all the way to the spec.) It's a shame though, since that would allow us to provide the reason without having to log out a message ourselves. I'm going to move forward with throwing for >= 6.1 and otherwise not throwing.

TAGraves:
@bensheldon I've updated this PR to switch the strategy based on the ActiveJob -- let me know if anything more needs to be done here! Thanks!

bensheldon:
@TAGraves thank you! 🙌🏻 This looks fantastic. I especially like the arguments handling both `before_` and `around_` callback signatures; very nice! When CI is green I'll merge and release.

PR #824, opened 2023-02-05
Ensure batch is reloaded before updating on multiple enqueues

### LINKED ISSUE #822: Can't batch.enqueue the callback after retrying a job within the batch
Awesome work with the batches release!

Happy to solve this problem myself, just wondering if you can point me to where in the codebase `batch.enqueue` doesn't re-initiate the batch callback if you retry a single job then call `batch.enqueue` again.

eg:
`batch.enqueue` works fine when all jobs are done, i can just make the callback happen over and over
`job_inside_batch.retry_job` + `batch.enqueue` doesn't work, it thinks i'm in the case where we've already enqueued the callback

Why would i do this? Well sometimes I want to retry all the failed jobs on a custom basis, and still have the callback happen again if i ask via `batch.enqueue` when the retries are done. Think of it as a special case of `batch.add` where i'm just retrying existing jobs instead of adding new ones.

It seems like the callback enqueue isn't actually checking if a callback has already been enqueued, but rather checking the status of the jobs. I would like to put up a pr modifying that. Let me know if this wasn't enough context. Thanks again!

bensheldon:
Fixes #822.

I'm not sure if the new lock is absolutely necessary, but I think so.

### DIFF
--- a/app/models/good_job/batch.rb
+++ b/app/models/good_job/batch.rb
@@ -86,18 +86,23 @@ module GoodJob
     # @return [Array<ActiveJob::Base>] Active jobs added to the batch
     def enqueue(active_jobs = [], **properties, &block)
       assign_properties(properties)
-      record.save!
+      if record.new_record?
+        record.save!
+      else
+        record.with_advisory_lock(function: "pg_advisory_lock") do
+          record.enqueued_at_will_change!
+          record.finished_at_will_change!
+          record.update!(enqueued_at: nil, finished_at: nil)
+        end
+      end
 
       active_jobs = add(active_jobs, &block)
 
       record.with_advisory_lock(function: "pg_advisory_lock") do
-        record.reload
-        record.finished_at = nil
-        record.enqueued_at = Time.current if enqueued_at.nil?
-        record.save!
+        record.update!(enqueued_at: Time.current)
+        record._continue_discard_or_finish(lock: false)
       end
 
-      record._continue_discard_or_finish
       active_jobs
     end
 
--- a/app/models/good_job/batch.rb
+++ b/app/models/good_job/batch.rb
@@ -90,9 +90,12 @@ module GoodJob
 
       active_jobs = add(active_jobs, &block)
 
-      record.finished_at = nil
-      record.enqueued_at = Time.current if enqueued_at.nil?
-      record.save!
+      record.with_advisory_lock(function: "pg_advisory_lock") do
+        record.reload
+        record.finished_at = nil
+        record.enqueued_at = Time.current if enqueued_at.nil?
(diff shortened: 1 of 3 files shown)


app/models/good_job/batch.rb
```diff
@@ -86,15 +86,23 @@ def initialize(_record: nil, **properties) # rubocop:disable Lint/UnderscorePref
...
 
       active_jobs = add(active_jobs, &block)
 
-      record.finished_at = nil
-      record.enqueued_at = Time.current if enqueued_at.nil?
-      record.save!
+      record.with_advisory_lock(function: "pg_advisory_lock") do
+        record.update!(enqueued_at: Time.current)
+        record._continue_discard_or_finish(lock: false)
+      end
```
v2kovac, line 104: Looks good! i'm agnostic about the enqueued_at reset, I think you're doing that in lieu of just putting the `add` in the advisory lock (if that's even possible)? Perfectly fine solution but i would consider enqueued_at to mean when the first jobs started coming in, doesn't matter that much though total batch statistics can be derived from the jobs.

v2kovac, line 104: Looks good! i'm agnostic about the enqueued_at reset, I think you're doing that in lieu of just putting the `add` in the advisory lock (if that's even possible)? Perfectly fine solution but i would consider enqueued_at to mean when the first jobs started coming in, doesn't matter that much though total batch statistics can be derived from the jobs.

v2kovac: if `add` can raise an exception you might be able to get in a weird spot where the first callback from the first enqueue doesn't even fire

bensheldon: > Perfectly fine solution but i would consider enqueued_at to mean when the first jobs started coming in...

Yes, that was my hope too but I realize it has an operational meaning of like "have the jobs been enqueued so that the finish callback should be triggerable".

>  I think you're doing that in lieu of just putting the add in the advisory lock

Exactly. I think it would get weird, like blocking any jobs from fully performing until all jobs had been inserted (right now that happens as a single Bulk insert, but I could see that not necessarily being the case in the future).

> if `add` can raise an exception you might be able to get in a weird spot where the first callback from the first enqueue doesn't even fire, and no callback is ever fired.

You're right! Hmm. I guess if the `enqueue` method had an exception raised through it, I'd expect it to be in a weird state. Hopefully I can not be bothered by that edge case for now 🤞🏻

v2kovac:
Really good stuff here, I think sidekiq batches just sidesteps this issue by saying something like "enqueueing while a batch is in progress is totally unsafe don't do that", they may have a point but you've gone above and beyond.

PR #845, opened 2023-02-13
Fix typo of "class" in README

mthadley:
Just a quick fix for a typo I noticed while reading through the new section on "Interrupts".

Thanks!


bensheldon:
Thank you 🙏

PR #853, opened 2023-02-19
Use heartbeat instead of advisory lock to track processes

bensheldon:
Connects to #831.

Capsule extracted in #861


bensheldon:
Closing for now because the Capsule was extracted in #861 and I'll do a clean redo of the Process record part.

PR #857, opened 2023-02-21
Fix bug where notification check does not use configuration

mitchellhenke:
Thanks again for working with me on this feature. I was testing it just now and realized the early `return true` was causing me to send notifications with the following set in my Rails application:

```ruby
    config.good_job.enable_listen_notify = false
```

### DIFF
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -234,8 +234,8 @@ module GoodJob
     end
 
     def send_notify?(active_job)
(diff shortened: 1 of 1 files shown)


bensheldon:
@mitchellhenke oops! 😓

bensheldon:
Released: https://github.com/bensheldon/good_job/releases/tag/v3.12.3

Thank you!

mitchellhenke:
Thank you for the quick merge and patch release!

Issue #490, opened 2022-01-22
Running CLI under foreman doesn't display log output until exit


bensheldon:
When running the `good_job` CLI under foreman, log output is not displayed until exit. 

<img width="745" alt="Screen Shot 2022-01-21 at 5 16 41 PM" src="https://user-images.githubusercontent.com/47554/150619643-532a9cf6-e394-459e-895f-77f82abbe103.png">

From lightly investigating, it seems like setting `STDOUT.sync = true` in an initializer is a quick workaround, but it would be nice if that wasn't necessary.

Related discussion: https://github.com/puma/puma/issues/2545

Reported by @kobaltz

bensheldon:
Done in #882

PR #879, opened 2023-03-04
Flush STDOUT buffer after writing logs to STDOUT

blafri:
Hey I took a stab at fixing #490. I was also having the issue of no log output when running the CLI in docker and did some digging and found out the following

When running good_job CLI from the console and logging to STDOUT the log messages will be displayed right away since ruby sets `$stdout.sync = true` by default, when run from an interactive shell.

When the good_job CLI is started by foreman or in a docker container the shell is non interactive and therefore ruby sets `$stdout.sync = false` by default, causing the log messages to be buffered and not shown until $stdout flushes the buffer (on exit for example).

See https://github.com/ddollar/foreman/wiki/Missing-Output

In the #490 issue you said it would be nice if the users didnt need to set `$stdout.sync = true` to get the logging to not buffer so we would need to manually flush `$stdout` to display the the logs right away.

In this patch we check to see if any of the log subscribers are writing to $stdout and if they are and $stdout.sync == false we call flush on $stdout to show the log message right away.

Thanks for all your amazing work on this gem and hope this patch helps out.

Let me know if anything looks wrong or if what I did is just a terible idea. :grimacing:

### DIFF
--- a/lib/good_job/log_subscriber.rb
+++ b/lib/good_job/log_subscriber.rb
@@ -168,6 +168,12 @@ module GoodJob
       GoodJob::LogSubscriber.logger
     end
 
+    # Are any of the loggers associated with this {LogSubscriber} instance writing to STDOUT
+    # @return [Boolean]
+    def logging_to_stdout?
+      GoodJob::LogSubscriber.logging_to_stdout?
+    end
+
     class << self
       # Tracks all loggers that {LogSubscriber} is writing to. You can write to
       # multiple logs by appending to this array. After updating it, you should
@@ -192,10 +198,18 @@ module GoodJob
         @_logger ||= begin
           logger = Logger.new(StringIO.new)
           loggers.each do |each_logger|
+            @_logging_to_stdout = true if ActiveSupport::Logger.logger_outputs_to?(each_logger, $stdout)
             logger.extend(ActiveSupport::Logger.broadcast(each_logger))
           end
           logger
         end
+      end
+
+      # Are any of the loggers attached to {LogSubscriber} writing to STDOUT.
+      # @return [Boolean] true if any loggers are writing to STDOUT and false otherwise
+      def logging_to_stdout?
+        logger if @_logger.nil?
+        @_logging_to_stdout.equal?(true)
       end
 
       # Reset {LogSubscriber.logger} and force it to rebuild a new shortcut to
@@ -204,6 +218,7 @@ module GoodJob
       # @return [void]
       def reset_logger
         @_logger = nil
+        @_logging_to_stdout = nil
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/good_job/log_subscriber.rb
```diff
@@ -243,13 +258,15 @@ def tag_logger(*tags, &block)
...
-          end                                            #   end
-        end                                              #
+        def #{level}(progname = nil, tags: [], &block)          # def info(progname = nil, tags: [], &block)
+          return unless logger                                  #   return unless logger
+                                                                #
+          tag_logger(*tags) do                                  #   tag_logger(*tags) do
+            logger.#{level}(progname, &block)                   #     logger.info(progname, &block)
+          end                                                   #   end
+                                                                #
+          $stdout.flush if logging_to_stdout? && !$stdout.sync  #   $stdout.flush if logging_to_stdout? && !$stdout.sync
```
blafri, line 268: Not sure if this is the best place to flush the $stdout buffer. Let me know what you think.

bensheldon:
@blafri this looks great (ignore the flakey test). Thank you! 

I was revisiting the history on this. It looks like Puma did go back to setting `$stdout.sync = true`: https://github.com/puma/puma/pull/2557

Given that this was a problem with the CLI specifically, I wonder if it should be simpler to set `$stdout.sync = true` at the place in the CLI where the stdout logger is set up: 

https://github.com/bensheldon/good_job/blob/29bd19707ab08e8e28cd6e966070a8d3f2892b16/lib/good_job/cli.rb#L151-L154

What do you think of that instead? I don't really know what the performance/operational differences are between flush or sync.

blafri:
I think setting `$stdout.sync = true` where you suggested would be better.

In terms of performance I can't see how calling flush after every `$stdout` write will be any slower than just setting `$stdout.sync = true`. Also when running the CLI app directly from a console the logs show up immediately specifically because ruby sets `$stdout.sync = true` for interactive consoles, so in terms of performance it would be exactly the same as it is right now when run from a console.

I could see why the puma devs did not want to mutate `$stdout` because their application server would be running other developers code, but as you pointed out, this is related to the good_job CLI application specifically, so I think ensuring the same setting for `$stdout.sync` for both interactive and non interactive terminals is actually the better thing to do to have consistent behaviour.

Would you want me to make the change in the pull request, or would you want to do it?

bensheldon:
Would you mind opening another PR with that change? I'd love to keep this PR (though closed) in case we do decide to return to it in the future.

blafri:
Sounds good.

bensheldon:
Closing this because #882 has been merged. Thank you again for the help on this!

PR #881, opened 2023-03-06
Add `raise: false` to FrontendsController after action skip

simi:
- it was removed by accident at 7a36f2676347a21db2bf7afee39828033aa80f69


bensheldon:
Thank you! Released: https://github.com/bensheldon/good_job/releases/tag/v3.12.8

PR #882, opened 2023-03-06
If logging to STDOUT in CLI set $stdout.sync = true

blafri:
When running the CLI from non-interactive terminals such as foreman or docker, the logs were not being displayed immediatly because ruby sets `$stdout.sync = false` for these types of terminals by default which causes the log messages to be buffered. This leads to inconsistent behaviour between interactive and non-interactive terminals for the CLI application. By explicitly setting $stdout.sync = true when launching the CLI application we can ensure that STDOUT logging is consistent in both types of terminals.

See #879 and #490 for further disscussion.

### DIFF
--- a/lib/good_job/cli.rb
+++ b/lib/good_job/cli.rb
@@ -148,7 +148,10 @@ module GoodJob
       # Rails or from the application can be set up here.
       def set_up_application!
         require RAILS_ENVIRONMENT_RB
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
@blafri Thank you this is great! 🎉 

I will merge this.

PR #883, opened 2023-03-08
Deprecate definition of job priority, change to "smaller number is higher priority" to align with Active Job definition

bensheldon:
Reported in https://github.com/bensheldon/good_job/discussions/524#discussioncomment-5233753.

[Active Job describes priority](https://github.com/rails/rails/blob/e17faead4f2aff28da079d50f02ea5b015322d5b/activejob/lib/active_job/core.rb#L22) as "lower is more priority".

### DIFF
--- a/app/models/good_job/execution.rb
+++ b/app/models/good_job/execution.rb
@@ -108,7 +108,13 @@ module GoodJob
     # @!method priority_ordered
     # @!scope class
     # @return [ActiveRecord::Relation]
-    scope :priority_ordered, -> { order('priority DESC NULLS LAST') }
+    scope :priority_ordered, (lambda do
+      if GoodJob.configuration.smaller_number_is_higher_priority
+        order('priority ASC NULLS LAST')
+      else
+        order('priority DESC NULLS LAST')
+      end
+    end)
 
     # Order executions by created_at, for first-in first-out
     # @!method creation_ordered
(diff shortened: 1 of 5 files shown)


app/models/good_job/execution.rb
```diff
@@ -206,6 +212,14 @@ def self.queue_parser(string)
 
     # Construct a GoodJob::Execution from an ActiveJob instance.
     def self.build_for_enqueue(active_job, overrides = {})
+      if active_job.priority && !GoodJob.configuration.smaller_number_is_higher_priority
+        ActiveSupport::Deprecation.warn(<<~DEPRECATION)
+          The next major version of GoodJob (v4.0) will change job `priority` to give smaller numbers higher priority, in accordance with Active Job's definition of priority.
+            To opt-in to this behavior now, set `config.good_job.smaller_number_is_higher_priority = true` in your GoodJob initializer or application.rb.
+            To not opt-in yet, but silence this deprecation warning, set `config.good_job.smaller_number_is_higher_priority = false`.
```
segiddins, line 219: I think the conditional will always resolve to true when the config value is false, maybe you want to check if it's `nil?`

segiddins (commented):
Thank you so much for the swift follow up!!

app/models/good_job/execution.rb
bensheldon: Whoops! yep, that should be a `nil?` check 😓

bensheldon:
Released in https://github.com/bensheldon/good_job/releases/tag/v3.14.0

zarqman:
@bensheldon, does this also affect the index on `:priority`?

In an existing app (with migrations from prior to this change), I have this index:
```ruby
add_index :good_jobs, [:priority, :created_at], order: { priority: "DESC NULLS LAST", created_at: :asc },
```

Does that need to become `priority: :asc` (which I believe is equivalent to `ASC NULLS LAST`)? Even if so, no idea how that should be handled with regard to the deprecation cycle. 😬 

Apologies if this is a dupe and this was already on your radar.

PR #885, opened 2023-03-13
Add execution_result to event payload for perform_job.good_job

segiddins:
This will enable listeners to log different things depending on the result of execution (in particular, I will use it to log the result as a statsd tag)

### DIFF
--- a/app/models/good_job/execution.rb
+++ b/app/models/good_job/execution.rb
@@ -336,7 +336,7 @@ module GoodJob
           current_thread.execution_interrupted = performed_at if performed_at
           update!(performed_at: Time.current)
 
-          ActiveSupport::Notifications.instrument("perform_job.good_job", { execution: self, process_id: current_thread.process_id, thread_name: current_thread.thread_name }) do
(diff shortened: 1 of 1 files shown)


bensheldon:
Adding to the payload is good, but I'm reluctant to expose the ExecutionResult as part of the public interface. 

Would it be ok to add the discrete params to the payload instead?

segiddins:
Something like `instrument_payload.merge!(execution_result.to_h)` ?

app/models/good_job/execution.rb
```diff
@@ -345,9 +345,9 @@ def perform
             end
             handled_error ||= current_thread.error_on_retry || current_thread.error_on_discard
 
-            ExecutionResult.new(value: value, handled_error: handled_error, retried: current_thread.error_on_retry.present?)
+            instrument_payload[:execution_result] = ExecutionResult.new(value: value, handled_error: handled_error, retried: current_thread.error_on_retry.present?)
```
bensheldon, line 348: As part of the public interface, I want to be explicit

```suggestion
            instrument_payload.merge!(
              value: value, 
              handled_error: handled_error, 
              retried: current_thread.error_on_retry.present?
            )
```

Issue #876, opened 2023-03-03
Avoid querying through `serialized_params`


ollym:
As we've scaled up with good_jobs we've started to feel significant pain particularly when the arguments of a job are large, the performance of queries that query through serialized_params can slow down dramatically.

I think you should add `job_class` and `executions` as columns and refactor any code that queries their values to not use `serialized_params`.

Is there any reason why you don't?

ollym:
A quick monkey patch for anyone interested and having the same issues:

```ruby
# db/migrate/....rb
add_column :good_jobs, :job_class, :string
add_column :good_jobs, :executions, :integer, default: 0
```

```ruby
# config/initializers/good_job.rb
Rails.application.reloader.to_prepare do
  Rails.application.configure do
    # ... good job configuration here
  end

  module GoodJob::JobExtensions
    extend ActiveSupport::Concern

    prepended do
      scope :job_class, ->(job_class) { where(job_class: job_class) }

      # First execution will run in the future
      scope :scheduled, -> { where(finished_at: nil).where('COALESCE(scheduled_at, created_at) > ?', DateTime.current).where(arel_table[:executions].lt(2)) }

      # Execution errored, will run in the future
      scope :retried, -> { where(finished_at: nil).where('COALESCE(scheduled_at, created_at) > ?', DateTime.current).where(arel_table[:executions].gt(1)) }
    end

    def job_class
      read_attribute(:job_class)
    end

    def serialized_params=(serialized_params)
      self.job_class = serialized_params['job_class']
      self.executions = serialized_params.fetch('executions', 0)
      super
    end
  end
  GoodJob::Job.prepend(GoodJob::JobExtensions)


  module GoodJob::ExecutionExtensions
    extend ActiveSupport::Concern

    prepended do
      scope :job_class, ->(job_class) { where(job_class: job_class) }
    end

    def serialized_params=(serialized_params)
      self.job_class = serialized_params['job_class']
      self.executions = serialized_params.fetch('executions', 0)
      super
    end
  end
  GoodJob::Execution.prepend(GoodJob::ExecutionExtensions)


  module GoodJob::BaseFilterExtensions
    def job_classes
       filtered_query(params.slice(:queue_name)).unscope(:select)
                                                .group(:job_class).count
                                                .sort_by { |name, _count| name.to_s }
                                                .to_h
     end
  end
  GoodJob::BaseFilter.prepend(GoodJob::BaseFilterExtensions)
end
```

[...]

Issue #438, opened 2021-10-29
Improve Dashboard Charts (feedback welcome)


bensheldon:
I would love feedback on these charts:

- Have a chart that is simply number of jobs performed successfully, and errored. Would be nice to have:
  - Last 1 hour, and next 15 minutes of upcoming queued/scheduled jobs, in 1 minute increments
  - Longer lookback of 72 hours hourly and maybe 30 days daily\
- Queuing Latency. This is the big operational one. Duration between when a job is queued/scheduled, and when execution starts.  
- Have some kind of min/max/avg (wish postgres made medians easier) [banded-line chart](https://github.com/chartist-js/chartist/issues/283#issuecomment-96365199) of job duration

Other charts?

I'd like to make these charts load asynchronously (SJR) so that the dashboard doesn't load too slowly.

bensheldon:
Continuing to think and work on this. For context, Heroku has these time bucket options:

<img width="1232" alt="Screenshot 2022-10-29 at 5 24 11 PM" src="https://user-images.githubusercontent.com/47554/198857274-b34b97ec-60e8-46ad-a741-8dad31915700.png">

I _really_ would like to have a cumulative flow diagram.

Some thoughts on what to show:

- Queue Latency (`performed_at - COALESCE(scheduled_at - performed_at)`). Bucketed based on when an execution is Performed. This should be split out by queue. Or Min, Avg, P95, Max. 
- Perform Latency (`finished_at - performed_at`). Could be split by queue or job class. Or Min, Avg, P95, Max.
- Total Latency (`finished_at - performed_at`). Could be split by queue or job class. Or Min, Avg, P95, Max.

I dunno quite how to offer both a (Min, Avg, P95, Max)-breakdown, and Queue (or Job) -based breakdowns

PR #886, opened 2023-03-13
Allow joining executions to jobs scoped by state

segiddins:
To do this, we need arel to be able to track the current table name for jobs (otherwise we get ambiguous column errors), so we construct arel nodes instead of using raw sql strings

This allows me to do something like:

```ruby
Filter.new({}).filtered_query(state:)
.joins(:executions).then do |rel|
      rel.pluck(
        *rel.group_values,
        Arel::Nodes::Max.new(
          [Arel::Nodes.build_quoted(now, rel.arel_table[:created_at]) -
            rel.arel_table.coalesce(*rel.send(:arel_columns, %w[executions_good_jobs.performed_at executions_good_jobs.finished_at executions_good_jobs.scheduled_at executions_good_jobs.created_at]))]
        )
      )
        .to_h { |*a, v| [a, v] }
    end
```

### DIFF
--- a/app/models/good_job/job.rb
+++ b/app/models/good_job/job.rb
@@ -23,6 +23,26 @@ module GoodJob
       def table_name=(_value)
         raise NotImplementedError, 'Assign GoodJob::Execution.table_name directly'
       end
+
+      def json_string(json, attr)
+        Arel::Nodes::Grouping.new(Arel::Nodes::InfixOperation.new('->>', json, Arel::Nodes.build_quoted(attr)))
+      end
+
+      def job_class
+        json_string(arel_table['serialized_params'], 'job_class')
+      end
+
+      def scheduled_at_created_at
+        arel_table.coalesce(arel_table['scheduled_at'], arel_table['created_at'])
+      end
+
+      def execution_count
+        Arel::Nodes::InfixOperation.new(
+          '::',
+          json_string(arel_table['serialized_params'], 'executions'),
+          Arel.sql('integer')
+        )
+      end
     end
 
     self.primary_key = 'active_job_id'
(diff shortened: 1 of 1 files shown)


bensheldon (approved):
This is great! Thank you! I just pushed up some fixes for some naming collisions and I'll merge.

segiddins:
Ah nice I was just about to fix that :)

In case it interests you, this is going towards https://github.com/rubygems/rubygems.org/pull/3582, which adds some statsd instrumentation that is aimed at a more job-centric latency than an individual execution one (basically, we had a few days of retries pile up when we switched one job over to GoodJob and our dashboard just showed the queue depth, but not the fast that some of them shouldve been run days ago)

bensheldon:
> In case it interests you, this is going towards https://github.com/rubygems/rubygems.org/pull/3582, which adds some statsd instrumentation that is aimed at a more job-centric latency than an individual execution one

Yikes! Sorry that the dashboard gave a false sense that everything was fine 😬 I want to improve those charts in #438 (I have some bad SQL in #888). For queue latency I'm thinking the query needs to be something like `COALESCE(performed_at, NOW()) - COALESCE(scheduled_at, created_at)` because that will allow jobs that haven't begun to run at all to accumulate latency in the calculation by using `NOW()`.

segiddins:
Ah no it wasn't your dashboard :) it was the one I built off of our statsd metrics, and I knew with was wrong

bensheldon:
> Ah no it wasn't your dashboard :)

Well, my Dashboard also has that problem 😭  

This and #885 was released in https://github.com/bensheldon/good_job/releases/tag/v3.14.1

ollym:
@bensheldon @segiddins hang about, can you make this change also for `GoodJob::Execution` seems inconsistent to do it for Job and not Execution. 

I'm asking as I need to update my patch here:
https://github.com/bensheldon/good_job/issues/876#issuecomment-1453682432

And would be far easier if i can just override the `params_execution_count` and `params_job_class` methods in both classes.

bensheldon:
@ollym oops, sorry about that! I'll make that change today.

PR #894, opened 2023-03-16
Extract shared methods/scopes into a BaseExecution to share between Execution and Job models

bensheldon:
Connects to https://github.com/bensheldon/good_job/pull/886#issuecomment-1469715363


ollym:
@bensheldon don't forget to replace `GoodJob:: BaseFilter#job_classes` which makes reference to `serialized_params->>'job_class'` which instead should fetch the column from here ^

bensheldon:
@ollym thank you for catching that 🙏🏻

PR #899, opened 2023-03-21
Extract more Dashboard strings into localization strings

patriciomacadden:
Hi! I've migrated our app that's been running with delayed_job for 7 years to good_job and I immediately fell in love with it. I'm a native spanish speaker so it was nice to see good_job has i18n support (though I would use the english version). I noticed only a bunch of strings were translated so wanted to see if a PR like this one would be helpful. 

It's far for complete, but again, wanted to see if a PR like this one would be helpful.

* most of the UI is translated (there are some hidden strings I'd translate before merging)
* yaml structure is open for discussion 
* it's got some controller notices translated (the rest will be translated)
* I haven't translated other locales, but can give it a go with the help of our friend google translate

I might be forgetting something, so if you come up with something let me know.

Cheers!

PS: Here are some screenshots with the before and after:

image
image


bensheldon:
❤️ This is fantastic ✨ 

My approach with I18n is trying to balance it as something that I think communicates care, and also unfortunately trails adding new features. So I really appreciate you helping close the gap. 

I can use i18n-tasks to pull down translations from Google Translate in an automated way. Don't sweat that (lints will fail until I do). 

Could you take a look at the RSpec test failures?

patriciomacadden:
Absolutely! I honestly haven't run those yet. Now that I know this is worth continuing I'll finish all the translations (controllers, hidden strings in views, etc) and will take care of the tests. Just gimme a couple of days!

Cheers!

patriciomacadden:
hey there!

Finished all view translations and looked at the tests. Ended up running `i18n-tasks add-missing` as otherwise there's one test that would fail. Can revert those files if needed.

Cheers!

bensheldon:
Something isn't right with GH Actions; they appear to be running against the wrong commit. Locally, this looks fantastic to me 👍🏻

patriciomacadden:
Awesome, thanks for merging! Let me know if there's something to fix re i18n. Happy to help!

PR #910, opened 2023-04-03
Switches from Chrome to Chromium to support Apple M1/Arm CPUs

ckdake:
Attempting to build the current devcontainer on `main` yields:

`Unable to locate package google-chrome-stable`.

Switching from `google-chrome-stable` to `chromium` helps to resolve this, but it leaves the chromedriver installation installing chromedriver_linux64.zip which bails with:

`qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`.   

This PR swaps both over!  This should still be fine on x86, but there may be consequences I'm not aware of from making this swap.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,7 @@
 FROM mcr.microsoft.com/devcontainers/ruby:2.7-bullseye
 
 # Default value to allow debug server to serve content over GitHub Codespace's port forwarding service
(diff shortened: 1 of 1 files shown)


bensheldon:
Thank you for fixing that! I mainly develop locally on an M1 Mac so this probably isn't well exercised.

ckdake:
👍🏻 Hopefully I'll have a more broadly useful PR later :)

PR #936, opened 2023-04-24
Refine french translations

francois-ferrandis:
Following #899 (thank you very much @patriciomacadden!), I took some time to refine the french translations. Many of them were incorrect or weird, so I tried to use the most usual terms, and also be careful of consistency.

These changes are a big improvement, but as always they is still room for more contributions!


bensheldon:
@francois-ferrandis Thank you for improving these! 🙏🏻

PR #953, opened 2023-05-14
Update README.md

namiwang:


namiwang:
The comment is not matching the real option `args` in L479

bensheldon:
Oh! Totally missed that. Thank you!

PR #955, opened 2023-05-17
Improve the migration guide for testing

shouichi:
Always setting `self.queue_adapter = :good_job` is problematic because assertions such as `assert_enqueued_jobs` won't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1108,7 +1108,9 @@ If your application is already using an ActiveJob backend, you will need to inst
     ```ruby
     # jobs/specific_job.rb
     class SpecificJob < ApplicationJob
-      self.queue_adapter = :good_job
+      # Don't set queue_adapter when testing, otherwise assertions such as
+      # assert_enqueued_jobs won't work.
+      self.queue_adapter = :good_job unless Rails.env.test?
       # ...
     end
     ```


bensheldon:
I'm reluctant to give a blanket statement like that; I've found that people interpret that sort of comment to mean "GoodJob doesn't work in the test environment", whereas this is more "if you use the ActiveJob Test helpers...". People have all sorts of different testing approaches 😓  For example, I use the TestAdapter/Helpers in unit tests, but use GoodJob Inline for my system/integration tests. I want to be clear that GoodJob is flexible.

I recently learned about `diff`-highlighted codeblocks. What would you think about making this more generic? e.g.

```diff
# Replace config/ references with GoodJob
- config.active_job.queue_adapter = :other_adapter
+ config.active_job.queue_adapter = :good_job

# Replace app/job/ references with GoodJob
- self.queue_adapter = :other_adapter 
+ self.queue_adapter = :good_job
```

shouichi:
I understand that you don't want to give the impression that GoodJob is inflexible. The point I want to make is that explicitly setting queue_adapter in the individual classes has an unintended side effect if the queue_adapter is originally configured globally. Thus, the following diff does not express the point.

```diff
# Replace app/job/ references with GoodJob
- self.queue_adapter = :other_adapter 
+ self.queue_adapter = :good_job
```

But you are correct, testing approaches vary and it might be difficult to communicate well in short sentences. I'm going to close this PR.

By the way, we started migrating the job backend of our monolith to GoodJob and found the web UI was handy. Well, good job!

PR #964, opened 2023-05-23
Use generic error reporter in Readme examples

shouichi:


bensheldon:
Sorry, I've been waffling on this because I'm worried that someone will try this on Rails < 7.0 and then be confused, but I think I'm committed to it. Thank you 🙏🏻

Issue #532, opened 2022-02-27
Add debugging and telemetry stats


bensheldon:
Have been getting GH Issues recently that would benefit from some basic stats and system info. It should be easy for someone to dump them to help debug issues. 

Imagining a method like `GoodJob.debug` to be run from a console, imagining the following fields to start:

- GoodJob version
- Ruby Version
- Rails Version
- Rails Env
- Execution mode
- Queues
- Polling Interval
- Notifier listening?
- Processes table (consider: maybe this is all data that should be going into the Processes table)

Seems like it would also be useful to have some counters on a `Scheduler.stats` too:

- Successfully executed jobs
- Failed executed jobs
- Job query misses (e.g. queried for a job but there were no executable jobs)

bensheldon:
Thanks for opening up #911. That is really helpful! 

Also, I feel bad because now that I'm seeing what I wrote in this issue I'm like "hmm, what was I thinking?" 😓 

Here are some thoughts that I want to bounce off of you. Sorry for the stream of consciousness:

- Should these be incrementers or gauges? If they're gauges then there needs to be a mechanism to "flush/reset" the values every interval.
- What should be triggering the stats dump? Should this be invoked from the command line, or should it be logged by GoodJob itself. If it's by GoodJob itself, should that log come from somewhere that already exists (for example, it gets logged every time the Scheduler tries to pick up a job) or would a de veloper overload an existing mechanism e.g. an ActiveJob extension, or via cron (which would need to take a Proc; currently it only allows jobs) or some new self-contained, threaded TimerSet object that would wake up to log stats. I don't feel very strongly either way, but each has its tradeoffs.
- Then there is the content of the info of being dumped:
   - Would this only be intended for processes that have GoodJob and are _executing_ jobs (e.g. the CLI or async web processes), or any process that has the GoodJob constant?
   - `Scheduler`s are the smallest unit to track. A process that is executing jobs could have multiple schedulers. Though maybe the [recently created `Capsule`](https://github.com/bensheldon/good_job/pull/861) is the correct unit to track. Though a capsule can have multiple schedulers, and a process _could_ have multiple capsules if someone was initializing them manually. Maybe the Capsule is the thing that has stats on it, and multiple schedulers increment a shared Stats object owned by the Capsule. (I'm imagining in this case the stats are like "number of jobs finished/errored/discarded/retried since creation"
[...]

PR #984, opened 2023-06-19
Added metrics to Scheduler and track in Process state

AndersGM:
https://github.com/bensheldon/good_job/issues/532


lib/good_job/metrics.rb
```diff
@@ -0,0 +1,19 @@
...
+  class Metrics
+    attr_reader :failed_count, :succeeded_count
+
+    def initialize
+      @failed_count = 0
+      @succeeded_count = 0
+    end
+
+    def increment_failed
+      @failed_count += 1
```
bensheldon, line 12: `+=` isn't threadsafe. This will need to be Concurrent types (`AtomicFixnum`?)

bensheldon (commented):
I'm happy to accept this. Briefly, it needs:

- threadsafety
- have the value show up when the Process serializes the Schedulers so that it can be queried
- I think the value should be reset when the the Scheduler is restarted (otherwise it might get weird when forking, especially with Puma's worker_fork

lib/good_job/metrics.rb
AndersGM: Indeed - fixed in 3111a289b91f67e57077599a4fd66293de64c4ed. Actually, I just used ```CleanupTracker``` as a start - suppose it should be threadsafe as well?

AndersGM:
@bensheldon Thank you for taking your time to look into this! Regarding the serialization - as far as I can see process uses name when serializing, so I've added the metrics to the name here in 78d8aba. Not sure if that is the best place though. What do you think?

bensheldon:
@AndersGM Good things to flag! What do you think about making a method like `Scheduler#statistics` that contained the name and those values and then dropping that into the Process state? (I also think they should be `succeeded_executions_count` for clarity).

Ok, and then I have a totally random thought that I want to share, but isn't telling you to do instead: What if we tracked `executed_by_process_id` on the Execution/DiscreteExecution object in the database? Would that be equivalent? (it could be joined in the database against the Process record)

AndersGM:
@bensheldon That would be better, but maybe not ```statistics``` since there already is a method named ```stats``` - what about ```name_with_metrics``` or just adding a parameter ```include_metrics``` to the name method? Agree about the more longer more explicit variable names.

Sound interesting. However, as far as I know, the executions can be destroyed through the UI (and cleanup?). Consequently, the resulting counter could actually decrement, without being reset. That is the reason I implemented it this way, to avoid that. Did you think of adding a new column or have I misunderstood something? :-)

app/models/good_job/process.rb
```diff
@@ -50,8 +50,10 @@ def self.current_state
         proctitle: $PROGRAM_NAME,
         preserve_job_records: GoodJob.preserve_job_records,
         retry_on_unhandled_error: GoodJob.retry_on_unhandled_error,
-        schedulers: GoodJob::Scheduler.instances.map(&:name),
+        schedulers: GoodJob::Scheduler.instances.map(&:stats),
```
AndersGM, line 53: @bensheldon What do you think about simply including stats? Might be problematic for backwards compatibility, but not sure how it is currently being used

bensheldon: 👍🏻

bensheldon:
@AndersGM I made a few tweaks and pushed this back up. It's looking great! 

I just merged in #977 which will update the Process record in the database every 30 seconds.

One thing I noticed that I'd like your feedback on. Currently this PR only records a job failure if it bubbles up to the thread-handler. This will cover only a small number of execution failures. We want to cover all the circumstances where a job's execution does not succeed, right? (e.g. if an execution fails and is retried, we'd still count that). That seem right? If you agree, I can make that tweak.

AndersGM:
@bensheldon indeed - thank you! I wasn't aware of that - let me know if I can be of any help!

lib/good_job/metrics.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+module GoodJob # :nodoc:
+  # Metrics for the scheduler.
+  class Metrics
+    def initialize
```
AndersGM, line 5: @bensheldon What do you think about making metrics more generic? something like
```suggestion
    def initialize(*keys)
      @counters = keys.map { |key| [key, Concurrent::AtomicFixnum.new] }
```
then increment could be something like
```
def increment(key)
@keys.fetch(key).increment
end
```

app/models/good_job/execution_result.rb
```diff
@@ -9,17 +9,24 @@ class ExecutionResult
...
+    attr_reader :unlocked_error
+    # @return [Exception, nil]
     attr_reader :retried
     alias retried? retried
 
     # @param value [Object, nil]
     # @param handled_error [Exception, nil]
     # @param unhandled_error [Exception, nil]
-    def initialize(value:, handled_error: nil, unhandled_error: nil, retried: false)
+    def initialize(value:, handled_error: nil, unhandled_error: nil, unlocked_error: nil, retried: false)
```
AndersGM, line 20: just curious, but why not just case on the type of the error?

lib/good_job/metrics.rb
bensheldon: My preference (at the moment) is to keep them hardcoded. They are still fairly discrete and I also worry about fat-fingering them when typing (having explicit methods will at least raise an exception)

app/models/good_job/execution_result.rb
bensheldon: oh! `unlocked_error` (I just named it to `unexecutable`) and `retried` are both boolean flags. It was in the spirit of what's in #995.

Just to explain, this is all a bit messy because it involves extracting state out of Active Job by storing it in a thread-local variable and then using it later.

PR #992, opened 2023-06-30
Change `is_discrete` to default to `false`

vfonic:
I got this RuboCop error when I ran `bin/rails g good_job:install` and thought how I could contribute with a simple fix.

Few hours later and I still cannot get the tests to green. :)

Here's the last error:

```
Selenium::WebDriver::Error::SessionNotCreatedError:
  session not created: This version of ChromeDriver only supports Chrome version 94
  Current browser version is 114.0.5735.198 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
```

I opened Chrome to check the browser version and it just updated from 114.0.5735.133 to 114.0.5735.198.
I went to download the chromedriver but, believe it or not, that version is 404-ing. :)
https://googlechromelabs.github.io/chrome-for-testing/#stable

Screenshot 2023-06-29 at 18 04 23

At least I have Ruby 3.2.2 installed now. That took at least an hour... :/

I hope there's some value in this. I decided to create another migration so that people who already have the gem installed can run `bin/rails g good_job:install` again and get Rails will only add the new migration to their project.

### DIFF
--- a/lib/generators/good_job/templates/install/migrations/add_default_false_to_is_discrete_on_good_jobs.rb.erb
+++ b/lib/generators/good_job/templates/install/migrations/add_default_false_to_is_discrete_on_good_jobs.rb.erb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultFalseToIsDiscreteOnGoodJobs < ActiveRecord::Migration<%= migration_version %>
+  def change
+    change_column_default :good_jobs, :is_discrete, from: nil, to: false
+  end
+end
--- a/lib/generators/good_job/templates/install/migrations/create_good_jobs.rb.erb
+++ b/lib/generators/good_job/templates/install/migrations/create_good_jobs.rb.erb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class CreateGoodJobs < ActiveRecord::Migration<%= migration_version %>
   def change
     enable_extension 'pgcrypto'


bensheldon:
Sorry, I missed the explanation for why add the default column value?

vfonic:
It might also be a good idea to perhaps introduce an index since `unfinished_undiscrete` scope uses it in `where` query, but it's tough to say whether this will be better or worse for overall performance.

Why add default value?

It seems like a good practice.

Also, so that you don't have to do:
`scope :unfinished_undiscrete, -> { where(finished_at: nil, retried_good_job_id: nil, is_discrete: [nil, false]) }`

Plus SQL `NULL` acts weird sometimes.
https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsthreestatebooleancolumn
https://rails.rubystyle.guide/#three-state-boolean

This column seems to only support two values "is discrete" and "is not discrete". There's no "discretion not defined", which could be the case for some other columns. From the top of my head a random example: Order `is_high_risk`" could be NULL for draft orders, which changes to `true` or `false` after the customer finishes the checkout and we check the customer details.

vfonic:
It might also be a good idea to perhaps introduce an index since `unfinished_undiscrete` scope uses it in `where` query, but it's tough to say whether this will be better or worse for overall performance.

Why add default value?

It seems like a good practice.

Also, so that you don't have to do:
`scope :unfinished_undiscrete, -> { where(finished_at: nil, retried_good_job_id: nil, is_discrete: [nil, false]) }`

Plus SQL `NULL` acts weird sometimes.
https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsthreestatebooleancolumn
https://rails.rubystyle.guide/#three-state-boolean

This column seems to only support two values "is discrete" and "is not discrete". There's no "discretion not defined", which could be the case for some other columns. From the top of my head a random example: Order `is_high_risk`" could be NULL for draft orders, which changes to `true` or `false` after the customer finishes the checkout and we check the customer details.

vfonic:
Perhaps `null: false` could also be added. It's probably a good idea to `update_all` from `nil` to `false` before adding `null: false` to prevent errors from being raised.

bensheldon:
@vfonic I appreciate you opening this PR, but I don't think I'm going to merge it. 

The `is_discrete` column is intended to be temporary until the next major release of GoodJob (v4) when it will be removed because all v4 jobs will be "discrete". If you're seeing a specific performance problem, I'd consider handling that now, but otherwise I expect this to be temporary.

vfonic:
Sounds good! Thanks for looking into this!

Quick question, sorry for hijacking the thread, where should I post questions if something in the docs is not clear to me? Should I open an issue here, SO, something else? Thanks! 🙏

bensheldon:
@vfonic no worries! You can open an Issue or Discussion here. I'm very happy to help.

PR #1003, opened 2023-07-09
Unify `frozen_string_literal` comment style

dixpac:
Some file used

```ruby
# frozen_string_literal: true
class ...
```

other used
```ruby
# frozen_string_literal: true

class ...
```

This commit makes the style same in every file,
I've opeted to same style that Rails uses, and majority of other projects.


bensheldon:
Works for me 👍 Is there a Rubocop rule we could adopt to keep this consistent? I imagine it getting inconsistent again about 5 minutes after merging this 😅

bensheldon:
Aha, found it: https://docs.rubocop.org/rubocop/cops_layout.html#layoutemptylineaftermagiccomment

dixpac:
@bensheldon awesome call on the rubocop rule ❤️

PR #1005, opened 2023-07-10
[Docs] Explain how `priority` works in GoodJob

### LINKED ISSUE #991: Explicitly write that jobs with higher priority jobs run first (unlike Que or DelayedJob) at readme.
Hi,

I am switching from [DelayedJob](https://github.com/collectiveidea/delayed_job).
To my surprise, totally by chance, I found out that unlike [DelayedJob's execution order](https://github.com/collectiveidea/delayed_job#parameters) which is:
`:priority (number): lower numbers run first`
image

[Que Gem](https://github.com/que-rb/que#usage) also uses `lower priority runs first`
```ruby
  # taken from Que doc
  # We use the Linux priority scale - a lower number is more important.
```

At `GoodJob` it's the other way around, meaning higher numbers run first (at least according to my tests).

The only place which hints about it is https://github.com/bensheldon/good_job#queue-performance-with-queue-select-limit
image

I think it would be a good idea to add somewhere in `readme`- (I gave it a look, not sure exactly where),
so it would be obvious without a doubt, explicitly written.

`good_job (3.15.13)`

coreyaus:
## PR Notes

- Resolves #991
- Explains the changes introduced by #883 
- Obviously feel free to make any changes/removals you see fit to update the wording or layout of these documentation updates (or close this PR if this doesn't suit). No dramas with whatever makes the most sense for the project 👍 
- Massive thanks for all the work on this excellent gem! 🙏 

## Question (priority order on DB indexes)

I also want to draw attention to the question raised by @zarqman [here](https://github.com/bensheldon/good_job/pull/883#issuecomment-1496382603), which I think is an important consideration: 

> @bensheldon, does this also affect the index on `:priority`?
> 
> In an existing app (with migrations from prior to this change), I have this index:
> 
> ```ruby
> add_index :good_jobs, [:priority, :created_at], order: { priority: "DESC NULLS LAST", created_at: :asc },
> ```
> 
> Does that need to become `priority: :asc` (which I believe is equivalent to `ASC NULLS LAST`)? Even if so, no idea how that should be handled with regard to the deprecation cycle. 😬
> 
> Apologies if this is a dupe and this was already on your radar.


bensheldon:
@coreyaus thank you for opening this! 🙌🏻  I'll make a few tweaks but this is really helpful! Also, thanks for surfacing that question about the index. I can make a migration to address that.

bensheldon:
@coreyaus sorry, I'm just now getting this cleaned up and merged.

I'm going to remove the section about named queues and priority. I don't think that people should use priority that way: priority is for ordering jobs _within_ a queue. I don't understand why there would be a 1:1 correspondence between queue and a priority (is this maybe a result of coming from a different queue backend that doesn't have the concept of both queues and priorities?)

For a named priority one could use a constant:

```ruby
HIGH_PRIORITY = -10
self.priority = HIGH_PRIORITY
```

PR #1008, opened 2023-07-12
Add `GoodJob::Job#display_name` to allow customizing dashboard job display

### LINKED ISSUE #956: Support for customized job display name
In our app, we make extensive use of the Command pattern, and have a meta-job that allows us to enqueue any Command as a job:

```ruby
class CommandJob
  def perform(command_class_name:, args: [])
    command_class_name.constantize.new.call(*args)
  end
end
```

This means that in our GoodJob dashboard, we just have a bunch of "CommandJob", without any details. The dashboard gets the "job name" from the `GoodJob::BaseExecution#job_class` model method. 

I can monkey-patch GoodJob::BaseExecution to provide a better name:

```ruby
  def job_class
    name = super
    if name == "CommandJob"
      cmd_name = serialized_params.dig("arguments", 0, "command_class_name")
      name += "[#{cmd_name}]"
    end
    name
  end
```

This renders as something like "CommandJob[ImportUsers]" in the dashboard, which is much more useful. However, I'm concerned that replacing "job_class" that doesn't return exactly the job class name would break the operation of GoodJob, if not now then possibly with a future update.

Would you be open to a PR that allows the name of the job on the dashboard to be customized? I can think of a few approaches:

1. Add `BaseExecution#job_display_name`, which can be monkey-patched
2. Add `GoodJob::ApplicationHelper#job_display_name(job)`, which can be monkey-patched
3. Add an optional method to the jobs themselves, like `MyJob.good_job_display_name(job)`, so no monkey-patching is required.

Let me know which of these would be preferred, and I'll open up a PR.

paul:
Fixes #956

The Dashboard displays jobs by default using the name of the Job class. If you want to customize this name, for example you have a generic job and want to include some part of the job arguments within the name, you can override `GoodJob::BaseExecution#job_display_name`:

```ruby
module GoodJobCustomDisplayName
  def job_display_name
    super + serialized_params.dig("arguments", 0, "my_specific_name")
  end
end
# In a Rails initializer:
  # :after_initialize, so GoodJob has had a chance to load its models
  ActiveSupport.on_load(:after_initialize) do
    GoodJob::BaseExecution.prepend(GoodJobCustomDisplayName)
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,6 +413,26 @@ end
 
 The Dashboard can be set to automatically refresh by checking "Live Poll" in the Dashboard header, or by setting `?poll=10` with the interval in seconds (default 30 seconds).
 
+#### Customizing Job Display Name
+
+The Dashboard displays jobs by default using the name of the Job class. If you want to customize this name, for example you have a generic job and want to include some part of the job arguments within the name, you can override `GoodJob::BaseExecution#job_display_name`:
+
+```ruby
+module GoodJobCustomDisplayName
+  def job_display_name
+    super + serialized_params.dig("arguments", 0, "my_specific_name")
+  end
+end
+
+# In a Rails initializer:
+Rails.application.configure do
+  # :after_initialize, so GoodJob has had a chance to load its models
+  ActiveSupport.on_load(:after_initialize) do
+    GoodJob::BaseExecution.prepend(GoodJobCustomDisplayName)
+  end
+end
+```
+
 ### Concurrency controls
 
 GoodJob can extend ActiveJob to provide limits on concurrently running jobs, either at time of _enqueue_ or at _perform_. Limiting concurrency can help prevent duplicate, double or unnecessary jobs from being enqueued, or race conditions when performing, for example when interacting with 3rd-party APIs.
--- a/README.md
+++ b/README.md
@@ -413,26 +413,6 @@ end
 
 The Dashboard can be set to automatically refresh by checking "Live Poll" in the Dashboard header, or by setting `?poll=10` with the interval in seconds (default 30 seconds).
 
-#### Customizing Job Display Name
-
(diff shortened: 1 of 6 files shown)


bensheldon:
@paul thank you for opening this! I made a few tweaks:

- I narrowed it so the method is only on `GoodJob::Job.display_name`. I expect `BaseExecution` to become deprecated in the next major release, and removed in the following one, whereupon `Job` and `Execution` models will point at separate tables.
- I removed the documentation from the Readme. I don't want to imply that this is "officially supported", even if I want to make it possible.

For monkeypatching, I think the best option is this:

```ruby
# config/initializers/good_job.rb

module GoodJobCustomDisplayName
  def display_name
    super + serialized_params.dig("arguments", 0, "my_specific_name")
  end
end

Rails.application.autoloaders.each do |zeitwerk|
  zeitwerk.on_load("GoodJob::Job") do
    GoodJob::Job.prepend(GoodJobCustomDisplayName)
  end
end
```

bensheldon:
This has been released in https://github.com/bensheldon/good_job/releases/tag/v3.16.2

Let know how it works for you and I'm open to additional tweaks/improvements 👍🏻

paul:
@bensheldon Thanks for the super quick turnaround! We're running it in production now, and its looking great!

image

mroach:
I wanted to chime in and thank you guys for this! We also use the command pattern and had the same issue with the dashboard, so this is a great enhancement for us as well.

I took a pretty similar approach, except inverting control to let the job classes set their own display names. 

```ruby
GoodJob::Job.prepend(Module.new do
  def display_name
    klass = job_class.constantize

    if klass.respond_to?(:good_job_display_name)
      super + "[" + klass.good_job_display_name(self).to_s + "]"
    else
      super
    end
  end
end)
```

```ruby
class RunCommandJob < ApplicationJob
  class << self
    def good_job_display_name(job)
      job.serialized_params.dig("arguments", 0, "command_class", "value")
    end
  end
end
```

PR #1009, opened 2023-07-15
Move shared `BaseExecution` concerns into the base class.

dixpac:
Parent/base class for `Job and Execution` is `BaseExecution`, both of these child classes include same concerns: `Lockable, Filterable, Reportable`.

This commit moves inclusion of these concerns to the base class `BaseExecution`

### DIFF
--- a/app/models/good_job/base_execution.rb
+++ b/app/models/good_job/base_execution.rb
@@ -5,6 +5,9 @@ module GoodJob
   # which both read out of the same table.
   class BaseExecution < BaseRecord
     include ErrorEvents
+    include Filterable
+    include Lockable
+    include Reportable
 
     self.table_name = 'good_jobs'
 
--- a/app/models/good_job/execution.rb
+++ b/app/models/good_job/execution.rb
@@ -3,10 +3,6 @@
 module GoodJob
   # ActiveRecord model that represents an +ActiveJob+ job.
   class Execution < BaseExecution
-    include Lockable
-    include Filterable
-    include Reportable
-
     # Raised if something attempts to execute a previously completed Execution again.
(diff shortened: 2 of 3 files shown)


app/models/good_job/base_execution.rb
```diff
@@ -5,6 +5,9 @@ module GoodJob
   # which both read out of the same table.
   class BaseExecution < BaseRecord
     include ErrorEvents
+    include Filterable
+    include Lockable
+    include Reportable
```
dixpac, line 10: Would love if I could group this; `include Filterable, Lockable, Reportable`
but rubocop is not allowing it :)

bensheldon: Seems like a standard style; Rails uses it too. My digging surfaced this, but not the best explanation :-) 

https://github.com/rubocop/ruby-style-guide/pull/622

bensheldon:
This looks good. Thank you! btw, there's going to be some major refactoring in GoodJob v4 that will do away with the BaseExecution. There will just be `Job` and `Execution` (currently called `DiscreteExecution`).

app/models/good_job/base_execution.rb
dixpac: I agree definitely not worth changing this rule, it just looked nicer in this particular situation to me :)

I think this rule is on by default because of: 
1. Git diffing
2. Error tracing - line number, if there are multiple mixins on 1 line could be harder to trace

dixpac:
>  there's going to be some major refactoring in GoodJob v4 that will do away with the BaseExecution. There will just be Job and Execution (currently called DiscreteExecution)

I'm willing to work on this, if you haven't done it already :)

Issue #1013, opened 2023-07-19
Understanding Database Connections and Cron


blumhardts:
The [docs for database connections](https://github.com/bensheldon/good_job#database-connections) make the development setup clear. Below is my development configuration.

```yaml
# config/database.yml

development:
  <<: *default
  database: myapp_development
  # Good Job runs in :async mode in development and requires a few extra
  # database connections.
  # See https://github.com/bensheldon/good_job#database-connections for
  # more information.
  # - the number of threads for the web server (default is 5)
  # - 1 connection per query pool thread (default pool thread size is 5)
  # - 1 connection dedicated to the scheduler, a.k.a. LISTEN/NOTIFY
  # - 2 connections for cron, if it's running
  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5).to_i +
    ENV.fetch("GOOD_JOB_MAX_THREADS", 5).to_i +
    1 +
    (if ENV["GOOD_JOB_ENABLE_CRON"] == "true" then 2 else 0 end) %>
```

Things are less clear in test and production.

In test, GoodJob runs in `:inline` mode. I assume, because there is no mention of the connections required, I can leave them at the following.

```yaml
# config/database.yml

test:
  <<: *default
  database: myapp_test
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
```

 In production, GoodJob runs in `:external` mode. I plan on having dedicated worker instances that process jobs, separate from my web servers, which are capable of enqueuing jobs.

1. Does the "scheduler, a.k.a. LISTEN/NOTIFY" connection need to be accounted for on workers and web servers, or just workers?
2. Do the 2 cron connections need to be accounted for on workers and web servers, or just workers?
3. Can each of my worker processes have cron enabled? [The docs seem to state it's okay to run it on all worker processes, but the example implies it should only be run on one worker](https://github.com/bensheldon/good_job#cron-style-repeatingrecurring-jobs):

    > Cron-style jobs are run on every GoodJob process (e.g. CLI or async execution mode) when `config.good_job.enable_cron = true`.

    ```ruby
    # config/environments/application.rb or a specific environment e.g. production.rb

    # Enable cron in this process; e.g. only run on the first Heroku worker process
    config.good_job.enable_cron = ENV['DYNO'] == 'worker.1' # or `true` or via $GOOD_JOB_ENABLE_CRON
    ```

So far, I have been operating on the assumption that the answers to these questions are:

1. Just workers
[...]

PR #1015, opened 2023-07-19
Clarify database connections and recurring processes in README.md

blumhardts:
See https://github.com/bensheldon/good_job/issues/1013 for more context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -956,7 +956,7 @@ Keep in mind, queue operations and management is an advanced discipline. This st
 
 GoodJob job executor processes require the following database connections:
 
-- 1 connection for the job listener, a.k.a. `LISTEN/NOTIFY`.
+- 1 connection for the job reader.
 - 1 connection per execution pool thread. E.g., `--queues=mice:2;elephants:1` is 3 threads. Pool size defaults to `--max-threads`.
 - 2 connections for the cron scheduler and executor, if cron is enabled.
 - 1 connection per subthread, if your application makes multithreaded database queries within a job.
--- a/README.md
+++ b/README.md
@@ -958,7 +958,7 @@ GoodJob job executor processes require the following database connections:
 
 - 1 connection for the job listener, a.k.a. `LISTEN/NOTIFY`.
 - 1 connection per execution pool thread. E.g., `--queues=mice:2;elephants:1` is 3 threads. Pool size defaults to `--max-threads`.
-- 2 connections for the cron scheduler, if it's enabled.
+- 2 connections for the cron scheduler and executor, if cron is enabled.
 - 1 connection per subthread, if your application makes multithreaded database queries within a job.
 
 The executor process will not crash if the connections pool is exhausted, instead it will report an exception (eg. `ActiveRecord::ConnectionTimeoutError`).
@@ -974,8 +974,8 @@ pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
 When GoodJob runs in `:async` mode (in Rails's development environment, by default), the following database pool configuration works, where:
 
 - `ENV.fetch("RAILS_MAX_THREADS", 5)` is the number of threads used by the web server
-- `1` is the job enqueuer
-- `2` is the cron scheduler
+- `1` is the job listener
+- `2` is the cron scheduler and executor
 - `ENV.fetch("GOOD_JOB_MAX_THREADS", 5)` is the number of threads used to perform jobs
 
 ```yaml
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,8 +59,8 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
     - [Timeouts](#timeouts)
     - [Optimize queues, threads, and processes](#optimize-queues-threads-and-processes)
     - [Database connections](#database-connections)
-        - [Production setup](#production-setup)
-        - [Queue performance with Queue Select Limit](#queue-performance-with-queue-select-limit)
+    - [Production setup](#production-setup)
+    - [Queue performance with Queue Select Limit](#queue-performance-with-queue-select-limit)
```
blumhardts, line 63: These don't seem to make sense under `Database connections`, so I moved them up a level.

README.md
```diff
@@ -954,24 +954,51 @@ Keep in mind, queue operations and management is an advanced discipline. This st
 
 ### Database connections
 
-Each GoodJob execution thread requires its own database connection that is automatically checked out from Rails’ connection pool. For example:
+GoodJob job executor processes require the following database connections:
```
blumhardts, line 957: "job executor process" is maybe just a fancy way to say "worker", but I saw "executor" a couple other times in the README and wasn't sure what the correct language for GoodJob was, so I just went with it.

bensheldon: I'm trying to avoid using "worker"  because it can be  ambiguous: https://github.com/sidekiq/sidekiq/discussions/4971#discussioncomment-1216694

The meaning here is "any process where GoodJob is executing jobs (eg the executable or async-mode)". 

I agree with you that the current wording is clunky and not necessarily clearer.

README.md
```diff
@@ -954,24 +954,51 @@ Keep in mind, queue operations and management is an advanced discipline. This st
...
+
+- 1 connection for the job listener, a.k.a. `LISTEN/NOTIFY`.
```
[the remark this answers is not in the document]
bensheldon: Only the LISTEN part runs in the GoodJob processes that execute jobs (cli or async). LISTEN requires its own dedicated database connection which is used in a longrunning background thread. 

The NOTIFY part is a short query that is emitted when a job is enqueued and uses the same database connection that is used to enqueue the job in the same context (eg in a controller in a  puma thread).

README.md
bensheldon, line 957: I can think the extra detail in this section overall is useful. 

And I think the level of detail implies something different than my general advice: 

Be generous with your database connection pool size. Better to configure a slightly larger pool than you think you need (which will only consume as many connections as it actually needs) than a smaller one (which will lead to connection pool timeouts and exceptions).

bensheldon:
Thank you for this! Feel free to keep tweaking and I'll also make a few tweaks and merge it 💖

README.md
blumhardts: "job executor process" is kind of clunky, but it is at least less ambiguous than "worker," in my opinion. Mike Perham also refers to the same concept as a "process" in [the Sidekiq docs](https://github.com/sidekiq/sidekiq/wiki/Best-Practices#4-use-precise-terminology), but using that word alone is again ambiguous (a web server that is enqueuing jobs but not executing them is also running as a process), so sticking a qualifier on the front like "job executor" hopefully disambiguates the term.

README.md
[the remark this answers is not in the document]
blumhardts: Great, thank you for the clarification! With that in mind, I have removed the ", a.k.a. `LISTEN/NOTIFY`" portion from this line as I think it makes things less clear, not more. I have also updated the wording from "1 connection for the job listener" to "1 connection for the job reader" as I now see you have used the phrasing "read jobs" elsewhere in the docs and I'd like to remain consistent.

I debated adding your above comment basically verbatim into the docs as an "Enqueueing and reading jobs" subheader under this "Database connections" section. I think it makes the connection usage and requirements clearer, but might be letting the implementation details leak into the docs a little too much.

README.md
blumhardts: > Be generous with your database connection pool size. Better to configure a slightly larger pool than you think you need (which will only consume as many connections as it actually needs) than a smaller one (which will lead to connection pool timeouts and exceptions).

I think this is good advice. My immediate follow-up question to that as a user of the gem is, "okay, well, then what's the lower bound where I'll start experiencing timeouts and exceptions, so I can make sure I always avoid that?"

Maybe that could be answered more concisely than what I am currently proposing.

The original docs answered that for `:async` mode, but lost me when I started thinking about `:external` mode. I think it was the lack of clarity between what a "job executor" process and its threads need and what a "job enqueuer" process and its threads need. Because those are the same process in `:async` mode (albeit with different thread pools for serving web requests and performing jobs), the requirements get jumbled together.

Issue #988, opened 2023-06-25
Add systemd example at documentation


Roko131:
First of all Thank you for an amazing project. We are transitioning from DelayedJob.

I would like to add a systemd service entry for GoodJob. I saw it's mentioned several times in various issues but never encountered an actual example or template.

Sidekiq have a [wonderful example](https://github.com/sidekiq/sidekiq/blob/main/examples/systemd/sidekiq.service) of systemd service.

Is it possible to add something like that here as well?

bensheldon:
> Is it possible to add something like that here as well?

Yep, definitely.

> Since goodjob is thread based as well - would you say we will benefit from adding MALLOC_ARENA_MAX=2 as well?

I'm not the expert on Ruby tuning, but seems legit: https://devcenter.heroku.com/changelog-items/1683

> Does good_job support systemd's Type=notify?

No, but I'm open if anyone wants to implement that.

Roko131:
Great!

This is what I am currently using,  (using `capistrano` and `rbenv`, and adding `user service` not `system service`):
I'm not vastly familiar with the `good_job` CLI tool- any improvement ideas will be more than welcome!

```sh
[Unit]
Description=GoodJob Background Job Processor
After=syslog.target network.target

[Service]
Type=simple

WorkingDirectory=/home/USER/domains/DOMAIN/app/current
ExecStart=/bin/bash -lc '/home/USER/.rbenv/shims/bundle exec good_job start'
Environment=RAILS_ENV=edge
Environment=MALLOC_ARENA_MAX=2

RestartSec=1
Restart=always

StandardOutput=syslog
StandardError=syslog

SyslogIdentifier=good_job

[Install]
WantedBy=multi-user.target
```

bensheldon:
That looks like a good example for me 👍🏻 

Would you make a PR and put this in a `examples/systemd` directory? Thank you!

Roko131:
Just to clarify, I meant that a more knowledgeable `systemd` person will build upon [sidekiq's example](https://github.com/sidekiq/sidekiq/blob/main/examples/systemd/sidekiq.service)  and add that as an example that is more tailored  to `good_job`'s case.

I'm not confident that my service is up to standards.
That being said, I wouldn't mind adding my basic example in a PR.

bensheldon:
Please make a PR! 🙌 We're all winging it.

PR #1029, opened 2023-08-05
Automatically send status notifications to systemd

### LINKED ISSUE #1027: Add systemd/sd_notify support to CLI
Since #988, it would be nice to have better systemd support (ideally GoodJob is run in a containerized environment; I think systemd is the best alternative to that). 

Two examples of this:

- Puma: https://github.com/puma/puma/pull/2438
- Sidekiq: https://github.com/sidekiq/sidekiq/pull/4489

I'm partial to Sidekiq's decision to simply vendor the `SdNotify`.

Mr0grog:
This adds built-in support for sending service status information to systemd as well as systemd’s watchdog (where it expects you to ping it every N seconds to let it know your process isn’t stuck and in need of a restart). It also adds an example systemd configuration file in `examples/systemd/goodjob.service`. Fixes #1027.

This works, but I’m not sure whether it’s totally ready. Some questions I was left with:

- In #1027, you asked to vendor the sd_notify gem, which I did. It looks like this is the first vendored dependency, so I looked around and saw that Rubocop was already configured to expect a top-level `vendor` directory, so I put it there. However, that meant I had to add that directory to the `.gemspec` file, and also add a `require_relative` call in a `lib` file, which seems non-ideal. Should this go somewhere else?
    - Nope, this is good
    - In `lib/good_job/vendor/<filename>`
    - In `lib/good_job/<filename>`, undifferentiated from other library files
    - Somewhere else?

- `sorbet/tapioca/require.rb` is missing some things that `bin/tapioca require` wants to add. I added the minimum needed to make Sorbet happy, but not sure it should be further updated to reflect what the command-line tools would otherwise add to it.

- Logging: is using ActiveSupport notifications and adding methods to `GoodJob::LogSubscriber` for logging correct? I originally tried to use `GoodJob.logger` (based on `configuration.rb` and `probe_server.rb`) but realized that neither outputs to STDOUT nor tags the log lines. I feel a little awkward about stuff that should be very narrowly scoped to the CLI executable leaking out into a place for handling more generic events across the library.

- The way watchdog support works is a little naive — it just starts a thread that sends notifications on a timer without checking anything. I think this is ok for a start (it’s the same as Sidekiq and Puma), but ideally, it would either:
    1. Check the status of various threads (I see the Probe Server kind of does this, but I wasn’t sure which things we should expect to always be good from a systemd perspective) or
    2. Should have other parts of the system proactively check in with it (to detect deadlocked threads or something).

    But really if you’re already confident that the GoodJob CLI crashes or gracefully exits anytime you’d need systemd to restart it, then the current behavior is fine.

- Relatedly, I kind of wondered if the CLI should use ActiveSupport notifications to signal status things, and both this and the Probe Server should subscribe to those instead of being managed directly in `GoodJob::CLI#start`. But that seemed like growing the scope of this change unnecessarily.

    (One thing this immediately would get is notification about the `scheduler_restart_pools` event. Systemd has a `RELOADING` notification GoodJob would ideally send it when this happens. But GoodJob also already write a log about that, so it’s probably not a big deal.)

- I cribbed a lot from Sidekiq and from @Roko131’s example in #988; I don’t know if I should credit those (particularly Sidekiq, as a separate project) somehow. I don’t think that’s super important for an example config file like this, but figure it’s worth calling out.

    (Also, some stuff in both of those seemed out of date and I cleaned it up — on Ubuntu 22, systemd screamed at me about using syslog, which is apparently no longer supported. You’re supposed to use journal now.)

- I named the example service `goodjob` to match the example rc.d configuration. But I kind of intuitively though maybe it should be `good_job`? Not really a big deal.


examples/systemd/goodjob.service
```diff
@@ -0,0 +1,92 @@
...
+#
+# THOSE PAGES ARE CRITICAL FOR ANY LINUX DEVOPS WORK; read them multiple
+# times! systemd is a critical tool for all developers to know and understand.
+#
+[Service]
+# Type=notify is supported as of GoodJob v3.17.0. In earlier versions, use
+# Type=simple and remove the WatchdogSec line.
+Type=notify
+# If systemd doesn't get pinged by GoodJob at least this often, restart GoodJob.
+WatchdogSec=5s
```
Mr0grog, line 48: Not sure if you have a preferred example interval here. I just threw in 5 seconds.

Mr0grog: Actually, it just occurred to me to look at the liveness probe example in the readme, which uses 10 seconds. Should I change it to match that?

examples/systemd/goodjob.service
```diff
@@ -0,0 +1,92 @@
...
+ExecStart=/usr/local/bin/bundle exec good_job start
+# If you use rbenv:
+# ExecStart=/bin/bash -lc 'exec /home/<USERNAME>/.rbenv/shims/bundle exec good_job start'
+# If you use rvm in production without gemset and your ruby version is 2.6.5
+# ExecStart=/home/<USERNAME>/.rvm/gems/ruby-2.6.5/wrappers/bundle exec good_job start
+# If you use rvm in production with gemset and your ruby version is 2.6.5
+# ExecStart=/home/<USERNAME>/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec good_job start
+# If you use rvm in production with gemset and ruby version/gemset is specified in .ruby-version,
+# .ruby-gemsetor or .rvmrc file in the working directory:
+# ExecStart=/home/<USERNAME>/.rvm/bin/rvm in <PATH_TO_YOUR_RAILS_APP> do bundle exec good_job start
```
Mr0grog, line 62: These are completely stolen from Sidekiq. With the exception of system ruby and rbenv, I haven’t tested them, and am not an expert. If you have opinions about the right way to execute in other environments, let me know how to update this.

lib/good_job/systemd_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      @watchdog.add_observer(self.class, :task_observer)
+
+      true
+    end
+
+    def watchdog_interval
+      return 0.0 unless GoodJob::SdNotify.watchdog?
+
+      Integer(ENV.fetch('WATCHDOG_USEC')) / 1_000_000.0
+    end
```
Mr0grog, line 67: Surprisingly, sd_notify does not handle getting the actual interval time for you (the C package it is a port of does). I filed an issue suggesting adding it: agis/ruby-sdnotify#6

spec/lib/good_job/systemd_service_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    @socket.close
+    File.unlink(@path) if @path
+  end
+
+  def read(maxlen = 16)
+    @socket.recvfrom(maxlen)[0]
+  end
+end
+
+RSpec.describe GoodJob::SystemdService, skip_if_java: true do
```
Mr0grog, line 29: JRuby seems to have trouble binding to the socket (this only matters for tests, not the actual implementation, I think).

I noticed Puma also skips tests on JRuby, presumably for this reason. Not sure if switching to `UNIXSocket` would help, but I don’t really have more time to play around with this at the moment.

examples/systemd/goodjob.service
bensheldon: 10 seconds was also arbitrary. I don't think it necessitates updating to be consistent; it's an example value.

examples/systemd/goodjob.service
bensheldon: I have no opinions 🤷🏻 I can wait to see if someone runs into trouble and hopefully opens an issue.

lib/good_job/systemd_service.rb
bensheldon: Thanks for opening that issue! I noticed it had `#watchdog?` where it did inspect `WATCHDOG_USEC`, but doesn't do the math here. This looks good.

spec/lib/good_job/systemd_service_spec.rb
bensheldon: This is good 👍🏻

bensheldon:
To your other questions:

- `lib/good_job/<filename>` is good. The file has enough boilerplate license at the top of the file that I think it's obvious that it's special.
- Sorbet change looks good 👍🏻 
- Instrumentation + `LogSubscriber` methods is perfect. I agree that it isn't tightly scoped, but it's nice having a central place to format log messages. 
- I think directly adding the behavior to the CLI is good for now. It would be nice to have hookable events/callbacks for the behavior... but we can consider it YAGNI until there are maybe a few more hooks.
- For `RESTARTING` that's good to note, but not necessary right now as there isn't a way to trigger a GoodJob restart externally (there's no `HUP` signal at the moment).
- If you wanted to rename both the service names to be `good_job`, I wouldn't object; it's also fine to leave them as is.

Mr0grog:
OK, I moved the `sd_notify.rb` module into `lib`, so assuming this passes tests, it should be good to go.

PR #1030, opened 2023-08-05
 Replace Webrick with custom simple http server

### LINKED ISSUE #1017: Replace webrick with a small/simple custom rack-compatible http server
I would like to drop the runtime dependency on webrick. Webrick is only used for the [probe server functionality](https://github.com/bensheldon/good_job#cli-http-health-check-probes) which is intended to be a private interface.

I think webrick could be replaced with <100 lines of Ruby to implement a bare minimum http server to be used for the probe server, thus it would only need to implement simple GET and string paths, and not do other query-parsing/etc.

dixpac:
This commit removes Webrick and replaces it with a (simple) custom HttpServer.

This is a single-process/singe-thread server that behaves similarly to webrick.

Initially I've build this implementation to comfort to `Rack::Handler` interface, but looks like rack team is removing `Rack::Handler` from the rack gem into separate gem and I'm not sure about the future of `Rack::Handler`, so i reverted to this implementation.

Resolves #1017


.rubocop.yml
```diff
@@ -46,6 +46,9 @@ Layout/ParameterAlignment:
 Lint/AmbiguousBlockAssociation:
   Enabled: false
 
+Lint/IncompatibleIoSelectWithFiberScheduler:
```
dixpac, line 49: Fiber-aware scheduling, instead of ths
`ready_sockets, = IO.select([@server])`
this:
`@server.wait_readable`

But afaik Ruby < 3, doesn't have fiber-aware scheduling or `wait_readable`, so I've disabled this linter.

PR #1037, opened 2023-08-10
Fix NoMethodError: undefined method `silence' for nil:NilClass

### LINKED ISSUE #1036: NoMethodError in ProcessHeartbeat if ActiveRecord logger is nil
We have ActiveRecord logging disabled in our application (`config.active_record.logger = nil`), which results in a NoMethodError in GoodJob::Notifier::ProcessHeartBeat#refresh_process [here](https://github.com/bensheldon/good_job/blob/d26b94835b87d6b8514c4578dabcc498775aa9f3/lib/good_job/notifier/process_heartbeat.rb#L26-L28):

```ruby
            GoodJob::Process.logger.silence do
              @process&.refresh_if_stale(cleanup: true)
            end
```

The error is:
```
Notifier errored: NoMethodError: undefined method `silence' for nil:NilClass
```

I'll follow up with a pull request shortly.

afn:
Don't attempt to call `logger.silence` if ActiveRecord::Base.logger is
nil.

Fixes #1036.

### DIFF
--- a/app/models/good_job/base_record.rb
+++ b/app/models/good_job/base_record.rb
@@ -28,6 +28,18 @@ module GoodJob
       false
     end
 
+    # Runs the block with self.logger silenced.
+    # If self.logger is nil, simply runs the block.
+    def self.with_logger_silenced(&block)
+      # Assign to a local variable, just in case it's modified in another thread concurrently
+      logger = self.logger
+      if logger
+        logger.silence(&block)
+      else
+        block.call
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


bensheldon (commented):
@afn for some reason I'm having trouble pushing into your branch, but to solve Sorbet, run:

`bin/tapioca todo` and then run `bin/lint` again and you should be golden. Thank you!

afn:
Done - thanks for the quick reply!

PR #1042, opened 2023-08-14
[Doc] for cron schedule add info about Dashboard view

steveroot:
Hi, 
Suggested minor update to the docs.

I noticed the Dashboard has a cron menu and that allows scheduled tasks to be called on demand or paused, which is fabulous! For testing I set myself a cron task to run every minute but I could have easily called it every time I was ready instead :-)
I didn't see this mentioned anywhere else in the read me... so this pull request suggest a one line addition in the cron section, linking back to the dashboard section for people skim reading.

PS Thanks for sharing your work on good_job!

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,6 +481,8 @@ GoodJob's cron uses unique indexes to ensure that only a single job is enqueued 
 
 Cron-format is parsed by the [`fugit`](https://github.com/floraison/fugit) gem, which has support for seconds-level resolution (e.g. `* * * * * *`) and natural language parsing (e.g. `every second`).
 
+If you use the [Dashboard](#dashboard) the scheduled tasks can be viewed in the 'cron' menu. In this view you can also disable a task or run/enqueue a task immediately.
+
 ```ruby
 # config/environments/application.rb or a specific environment e.g. production.rb
 


bensheldon:
Thanks @steveroot! 

Totally random thought, but this also makes me wonder: Is there a fugit value that could be  used in `cron:` config so that the job _never_ runs automatically but still could be triggered manually through the dashboard? (Something like either "run once at 0-0-0 0:0;0" or Infinity or something). Your PR made me think "hey, I wonder if this could be done without changing the GoodJob code?"

steveroot:
hehe, I was wondering the same thing as I experimented and saw the manual run option in the dashboard. It could be lazy way for me of having commands ready to run via web interface rather than via command line.

So, I thought I could be clever and find a crontab day/month/dayofweek that wouldn't run until decades away but then found: https://stackoverflow.com/questions/8324306/a-cron-job-that-will-never-execute and the comments pointing out crontab handles that possibility and makes sure it runs anyway,

So my next thought is I could pick a single day of the year and write my script(s) to prevent them running on that day. EG: Christmas Day https://crontab.guru/#0_12_25_12_* ... which is just asking for me to make a mistake in my code and bring my hobby servers down on Christmas Day, probably not the gift I'm hoping for :-)

PR #1043, opened 2023-08-16
New probe server fixes

stas:
A couple of suggested fixes for the new probe http server. Commits should be self-explanatory, still some more context below.

---

Passing `exceptions: false` no longer raises `IO::WaitReadable` leaving the server in an unknown state.

> By specifying a keyword argument exception to false, you can indicate that [accept_nonblock](https://ruby-doc.org/3.2.2/exts/socket/TCPServer.html#method-i-accept_nonblock) should not raise an IO::WaitReadable exception, but return the symbol :wait_readable instead.

---

An `IO.gets` returns either a string or a nil. A nil response from a socket was not handled, suggesting we just skip those.

---

Appreciate taking a look as we're affected by these issues.

### DIFF
--- a/lib/good_job/http_server.rb
+++ b/lib/good_job/http_server.rb
@@ -49,8 +49,10 @@ module GoodJob
           client = @server.accept_nonblock
           request = client.gets
 
-          status, headers, body = @app.call(parse_request(request))
-          respond(client, status, headers, body)
+          if request
+            status, headers, body = @app.call(parse_request(request))
+            respond(client, status, headers, body)
+          end
 
           client.close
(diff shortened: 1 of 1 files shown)


bensheldon:
@stas thank you! This change makes sense. I guess we were using `exception: false` but not checking if the return value == `:wait_readable`. From reading the docs, it wasn't entirely clear if there were other classes of exceptions that could be raised. But I'm ok accepting this and we'll see what we learn next 👍🏻

bensheldon:
@stas I just released this in https://github.com/bensheldon/good_job/releases/tag/v3.17.3

Issue #1035, opened 2023-08-09
JRuby deprecation policy


bensheldon:
Got a reply from [`@headius` on Reddit](https://www.reddit.com/r/ruby/comments/143qnqx/comment/jvi9601/?utm_source=reddit&utm_medium=web2x&context=3) that I want to preserve for the future:

> We support two major releases at a time, so in this case 9.3 and 9.4 are supported.
> 
> We generally support EOL releases for another year; 9.3 will probably stop getting maintenance toward the end of this year (especially if we push 9.5).
> 
> We strongly recommend everyone run latest JRuby, and 9.4 is very good and very stable.
>
> We got a bit behind on the major Ruby version compatibility due to the pandemic and other issues catching up, but we plan to stay as current as possible now and 9.4 has been out for almost a year now.

PR #1047, opened 2023-08-22
Upgrade Puma to the latest version

dixpac:


bensheldon:
I think we're blocked here until JRuby 9.3 (Ruby 2.6 compatible) is EOLed #1035

dixpac:
Yup

Issue #1050, opened 2023-08-25
[Feature Request] Hook to extend cron schedules


DanielHeath:
Context:

I have a job that should run every 17 hours. 

If I specify `every 17 hours`, the job actually runs at 5pm every day, because fugit interprets it as a cron string of `0 */17 * * *`.

I'd argue that's somewhat a bug in fugit (it should raise an error if given an 'every X hours' greater than 12, since it can't represent it).

I plan to work around this for now by subclassing `CronEntry` to override `next_at` and `original`, and then overriding `GoodJob.configuration.cron_entries` to return instances of my subclass for the relevant jobs.

Suggestions:

`CronEntry#fugit` is only used within `CronEntry` and is an implementation detail - could be made private.

Extending the cron configuration syntax to take a block and a description parameter, eg:

```
config.good_job.cron = { example: { 
  description: 'every 17 hours', 
  next_time: -> (last_run_at) { (last_run_at || Time.now) + 17.hours}, 
  class: 'ExampleJob'  
} }
```

DanielHeath:
I'd be happy to write a PR for this feature, if it would be accepted.

bensheldon:
I really like this! I think my preference would be to overload the existing `cron` key rather than introduce a new key. Hopefully it wouldn't get too messy. Totally take a swing at it and I'm happy to help finish it up if it gets hairy.

bensheldon:
I guess I'm imagining that the proc could return a string, in which case it would be passed to fugit, or a time/datetime in which case it would be used. I like your design of passing in `last_run_at`.

DanielHeath:
My reasoning for a new description key was that it can't necessarily be derived from the time proc.

Happy to re-use the `cron` key for the proc argument.

bensheldon:
Implemented in #1051

PR #1051, opened 2023-08-25
Support configuring cron schedule dynamically with a block

DanielHeath:
Implements https://github.com/bensheldon/good_job/issues/1050#issuecomment-1692682116

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ config.good_job.cron = {
   },
   complex_schedule: {
     class: "ComplexScheduleJob",
-    cron: -> (last_ran) { last_ran.blank? ? Time.now : last_ran + 14.hours}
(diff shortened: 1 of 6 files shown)


bensheldon:
@DanielHeath I just pushed up some changes.

- Now uses the last in-memory time the job was scheduled _or_ the time in the database. This is necessary because the next task is scheduled _before_ the previous task is enqueued.
- I wish the database query could be removed altogether because it happens during the Rails initialization process and I've seen that cause autoloading deadlocks in the past... but for safety I wrapped it in a `Rails.application.executor`.
- Added a few more tests

DanielHeath:
Anything else required for this to be merged?

bensheldon:
Released 🎉 https://github.com/bensheldon/good_job/releases/tag/v3.18.0

Issue #890, opened 2023-03-14
How can I avoid a job being discarded when retries are exhausted?


TAGraves:
We have a job configured like:
```
    retry_on SomeError, wait: 5.seconds, attempts: 4 do |job, _exception|
      Rails.logger.warn("SomeError still occurred after retries were exhausted")
      # do something to handle this
    end
```

When the retries are exhausted, the block we pass to `retry_on` is called, but SomeError is still propagated up to GoodJob, causing the job to get discarded with SomeError. We'd instead like to treat the job as successful in this case.

The docs seem to suggest this should already be happening by default, but it isn't for us.
> When using retry_on with a limited number of retries, the final exception will not be rescued and will raise to GoodJob's error handler. To avoid this, pass a block to retry_on to handle the final exception instead of raising it to GoodJob

bensheldon:
I think what you're describing is the intended behavior, which is how I understand "Discarded" to mean: a job that has errored, but not been retried, is discarded. 

Active Job is vague on this so far as I can tell. Though it does has instrumentation for "retry_stopped" which is different than the instrumentation for "discard". 

Can you tell me a bit more about how you think of a job being "discarded" is different that retires being exhausted? You mention you want to marked as "succeeded" but maybe there's a third state to expose?

The other thing I could suggest, more manually, is for you not to use `rescue_from` and instead use `around_perform`and manually count executions and call `retry_job`.

TAGraves:
I think from a practical perspective, what I want is:
1. The job should show up under the "Succeeded" tab rather than "Discarded" tab on the dashboard (and shouldn't be retriable anymore)
2. `good_job.on_thread_error` shouldn't be called.

> a job that has errored, but not been retried, is discarded

I think based on my reading of the documentation, it doesn't really make sense to consider this a job that has errored. I'm passing a block to _handle_ the error, right?

I guess my original post wasn't totally clear on this, but I specifically mean when a block is passed to `retry_on`. It definitely seems like the correct behavior for the job to be discarded when retries are exhausted and a block is _not_ passed to `retry_on`. But if I pass a block, I'm handling the error in the block and I want to treat it as rescued and handled. (I can always re-raise the error if I do want to explicitly give it back to GoodJob!)

> maybe there's a third state to expose

[...]

PR #1073, opened 2023-09-13
"Force" discard jobs that are already running/runaway to prevent retry

### LINKED ISSUE #625: What's the best way to stop and discard a running job?
We recently had an issue where a couple of our jobs would get stuck in the `running` state due to an unrelated bug. These jobs would take up workers which would cause other jobs not to run.

We needed a way to kill and discard these jobs so they wouldn't get retried but couldn't find it. The `discard` button is disabled when they are in the `running` and trying to discard them via the rails console using `discard_job` would raise a `GoodJob::Lockable::RecordAlreadyAdvisoryLockedError`.

As a workaround we ended up commenting the code in these jobs out and deploying that which would trigger a retry which would then no-op. However, we do need to find a reliable way to kill these jobs in the future in case something like this happens again. Do you have any suggestions on how to do that?

jgrau:
Allows jobs to be marked as finished, with an error, even if they are stuck in a `running` state. Should only be used in exceptional circumstances. Naming is to distinguish from regular discard, which will not operate on a running job.

"Force Discard" will not stop the running job, but will prevent it from retrying.

Closes #625. Connects to #890.

This is just a draft as I wanted to get some early feedback on whether this is the right approach. To be honest: I am fumbling in the dark a little bit :)

If I understand this correctly, then we're marking a running job as discarded (finished but with an error) even though it is still running and it will continue to run until it's somehow killed eg. the worker restarts) but at least now the job is not being picked up again. If this is correct, then I feel like it needs a little more explanation.

Anyways, all suggestions and advise is welcome :)


bensheldon:
@jgrau this looks great! 👍🏻 

I like that you extracted the common logic and the only change between `discard_job` and `force_discard_job` is that the former is wrapped in an advisory lock and the latter isn't. That seems right.

>  we're marking a running job as discarded (finished but with an error) even though it is still running and it will continue to run until it's somehow killed eg. the worker restarts) but at least now the job is not being picked up again.

That is correct. I think the only remaining work here is to:

- Clearly communicate that behavior (or difference between a regular discard) to the User; e.g. copywriting and figuring out how to fit it into the Dashboard UI (maybe in the browser confirmation modal?)
- Maybe change the error message (or have a separate `ForceDiscardError`) so that it's searchable in the database
- Add a test for `force_discard_job` that does something simple like take an advisory lock and verify that it still works. Something like this:

```ruby
job = GoodJob::Job.new ...

locked_event = Concurrent::Event.new
done_event = Concurrent::Event.new

promise = Concurrent::Promises.future do
  job.advisory_lock do
    locked_event.set
    done_event.wait(10)
  end
end

locked_event.wait(10)

job.force_discard_job
expect(job.finished_at).to be_present

done_event.set
promise.join!
```

bensheldon:
Also, I'm happy to take over the PR at any point, so please don't feel any pressure if any of those beyond you right now.

jgrau:
Hi @bensheldon . I've updated the PR with

* Better communication about what "force discard" does. It would be cool if we could communicate this before the user actually clicks the button (you have to be pretty brave to click a "force discard" button without knowing what that means) but it's better than nothing.
* Copywriting and translations (using i18n-tasks translate-missing)
* Differentiated error message on discarded and force discarded jobs

As far as I can tell the only thing missing is updating the jobs_spec but that is beyond me so I will leave that to you :) Let me know if you need me to do anything :)

bensheldon:
Thank you! This looks excellent! 🙌 

I'll take a look at the tests and get this merged 👍

Also, I agree about the need for less-scary affordances. I won't block this, but two thoughts:

- Having an info icon on the button with a hover/popover. 
- Building our native modals with more details (I guess we'd still need an affordable to indicate "don't worry, this has an additional confirmation step")

bensheldon:
Released in https://github.com/bensheldon/good_job/releases/tag/v3.19.0

Thank you for implementing this 🎉

PR #1075, opened 2023-09-15
Allow Probe Server's `/connect` to handle a certain number of reconnects before statusing

### LINKED ISSUE #1068: Probe failures on heavy usage of dashboard(?)
Hi @bensheldon 

This is more of a question than an issue. We're using good_job in a kubernetes environment with lots of success. We experiencing that the health check fails a couple of times a day. A few days ago I was browsing the good job dashboard and noticed that at the same time, the healthcheck failed and kubernetes restarted the workers. The dashboard does seem to perform some very slow queries but on the other hand the dashboard is served by the webserver pods and not the worker pods. 

Anyways: I just wanted to quickly check in to see if this made any sense to you before I go on a wild-goose chase... :)

bensheldon:
Fixes #1068.

Once GoodJob has successfully connected, Probe Server's `/connected` endpoint will continue returning 200-success if the database connection is lost and reconnections are attempted. This is now aligned with the existing thread-error reporting behavior:
 https://github.com/bensheldon/good_job/blob/3c45787add63c72f6c365275f518121664faa126/lib/good_job/notifier.rb#L152-L160

I also increased the number of reconnect attempts, so it will now be 6 attempts, once every 5 seconds = 30 seconds of reconnecting before the `Notifier#connected?` will become false and the Probe Server will start returning a 503 and the `GoodJob.on_thread_error` callback will be triggered.

Also some slight refactoring so that the utility methods `listen?` and `wait?` are now condition variables and have a timeout.

### DIFF
--- a/lib/good_job/notifier.rb
+++ b/lib/good_job/notifier.rb
@@ -24,6 +24,9 @@ module GoodJob # :nodoc:
     WAIT_INTERVAL = 1
     # Seconds to wait if database cannot be connected to
     RECONNECT_INTERVAL = 5
+    # Number of consecutive connection errors before reporting an error
+    CONNECTION_ERRORS_REPORTING_THRESHOLD = 6
+
     # Connection errors that will wait {RECONNECT_INTERVAL} before reconnecting
     CONNECTION_ERRORS = %w[
       ActiveRecord::ConnectionNotEstablished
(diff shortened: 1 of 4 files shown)


jgrau (approved):
FWIW I feel the design and code is good :)

bensheldon:
@jgrau This has been released! https://github.com/bensheldon/good_job/releases/tag/v3.18.3

PR #1097, opened 2023-09-29
Address rails 7.1 deprecation when using `ActiveJobExtensions::Concurrency`

### LINKED ISSUE #1096: Including GoodJob::ActiveJobExtensions::Concurrency triggers rails deprecation notice
Rails is renaming `exponentially_longer` to `polynomially_longer`, see here: https://github.com/rails/rails/pull/49292.

Running GoodJob with the 7.1 release canidate and including the concurrency extension triggers the following deprecation warning:
> DEPRECATION WARNING: `wait: :exponentially_longer` will actually wait polynomially longer and is therefore deprecated. Prefer `wait: :polynomially_longer` to avoid confusion and keep the same behavior.

I don't think GoodJob needs to go through a deprecation period itself since I believe it's all just passed to rails, so just tweaking the internal usage should be good.

Earlopain:
Closes #1096, it's a simple fix so why not do it myself.

There are still a few references to exponentially_longer in the readme but it didn't seem wise to change those while it's not actually usable. I did change one instance in the readme where it explains what it does, since it's not actually exponetially longer but if you think that's just going to cause confusion I'll just drop that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -816,7 +816,7 @@ GoodJob.on_thread_error = -> (exception) { Rails.error.report(exception) }
 
 By default, GoodJob relies on ActiveJob's retry functionality.
 
-ActiveJob can be configured to retry an infinite number of times, with an exponential backoff. Using ActiveJob's `retry_on` prevents exceptions from reaching GoodJob:
+ActiveJob can be configured to retry an infinite number of times, with a polynomial backoff. Using ActiveJob's `retry_on` prevents exceptions from reaching GoodJob:
 
 ```ruby
 class ApplicationJob < ActiveJob::Base
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -36,10 +36,16 @@ module GoodJob
           end
         end
 
+        wait_key = if ActiveJob.gem_version >= Gem::Version.new("7.1.0")
+                     :polynomially_longer
+                   else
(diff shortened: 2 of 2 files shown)


mhenrixon (approved):
Looks good to me, just came here to create a PR, this is how I would have solved it.

bkeepers (approved):
👍 same here as @mhenrixon. Thanks @Earlopain

bensheldon:
thank you!

Earlopain:
Just a note, rc (and probably beta) versions don't match like this. The deprecation warning will stop triggering once a non-preview version of rails is released.

bensheldon:
@Earlopain oh, we should fix that. There's a version matcher that will include RC versions, I just need to find/remember what it is.

Earlopain:
My bad, I wasn't aware about that functionality. As per the docs appending `.a` to the version check should do the trick. https://ruby-doc.org/3.2.2/stdlibs/rubygems/Gem/Version.html. A quick test seems to confirm this.

bensheldon:
Thanks! I'll make that change on `main` 👍🏻

edouard:
Hey @bensheldon I see you’ve released 3.19.4 but [the latest version released](https://rubygems.org/gems/good_job) on rubygems.org is 3.19.3. Could you push 3.19.4 to rubygems.org?

bensheldon:
Oops! Sorry about that! I just pushed the gem.

PR #1127, opened 2023-10-24
Add "cron enabled" column to processes index page

bforma:
Screenshot 2023-10-24 at 10 23 15

I've also added translations (from Google Translate).


bensheldon:
@bforma thank you! This looks great! And double-thanks for even doing the translations 🙇🏻 

I'll fix the `bin/lint` errors and then will merge 🙌🏻

PR #1134, opened 2023-11-01
Readme querying and dequeuing

ManickYoj:
In working with GoodJob through incidents, I and the other devs at my company weren't able to easily find any instructions on how to dequeue jobs. We've since written our own procedure down in a 'runbook', but for the benefit of other GoodJob users, I wanted to include this information in the README itself, as that's where we always look first!

While I wrote down what I know, I'm far from an expert on GoodJob. I welcome corrections; I'll learn something in the process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -961,10 +961,19 @@ jobs = GoodJob::Job
   .where("serialized_params ->> 'job_class' = 'DoSomeThingJob'")
 ```
 
-GoodJob provides a scope for finding jobs by job class, as this is a common operation:
+GoodJob provides some scopes for finding jobs by common identifiers. Eg. to find a job by job-class without manually querying serialized parameters, you can instead query via the `job_class` scope:
 ```
 jobs = GoodJob::Job.job_class('DoSomeThingJob')
 ```
+
+Other scopes include:
+- `running`
+- `finished` - jobs that have been completed (with or without an error) and will not be re-run
+- `scheduled` - jobs that have not run yet and will run in the future
+- `retried` - jobs that have attempted to run at least once, did not complete, and will run again in the future
+- `queued` - jobs whose scheduled time to run has passed and will run when a thread to run them becomes available
+- `succeeded` - jobs that have run without an error
+- `discarded` - jobs that encountered an error while running and wil not be retried
 
 ### Dequeuing Jobs
 
@@ -973,8 +982,8 @@ Jobs can be dequeued by deleting them out of the Jobs table:
 ```
 GoodJob::Job
   .job_class("<JobNameHere>") # Filter to a particular job type
-  .where(finished_at: nil) # Only query jobs that haven't run
-  .delete_all # Or `.each(&:destroy)` for callback support
+  .where(finished_at: nil) # Filter to enqueued jobs, ignoring those that have previously completed
+  .delete_all # Or `.each(&:destroy)` for Rails callback support (at the expense of speed)
 ```
 
 ### Optimize queues, threads, and processes
--- a/README.md
+++ b/README.md
@@ -914,6 +914,69 @@ class ApplicationJob < ActiveJob::Base
 end
 ```
 
+### Querying Jobs
+
+Jobs are contained in the mounted GoodJob::Job table. The schema is:
+
+```
+GoodJob::Job < GoodJob::Execution {
+                     :id => :uuid,
+             :queue_name => :text,
+               :priority => :integer,
+      :serialized_params => :jsonb,
+           :scheduled_at => :datetime,
+           :performed_at => :datetime,
+            :finished_at => :datetime,
+                  :error => :text,
+             :created_at => :datetime,
+             :updated_at => :datetime,
+          :active_job_id => :uuid,
+        :concurrency_key => :text,
+               :cron_key => :text,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -914,6 +914,78 @@ class ApplicationJob < ActiveJob::Base
...
+            :finished_at => :datetime,
+                  :error => :text,
+             :created_at => :datetime,
+             :updated_at => :datetime,
+          :active_job_id => :uuid,
+        :concurrency_key => :text,
+               :cron_key => :text,
+    :retried_good_job_id => :uuid,
+                :cron_at => :datetime
+}
```
ManickYoj, line 938: We're a few versions behind on GoodJob. If I didn't get the schema quite right with the current branch, please either let me know or put in the proper, up-to-date schema.

README.md
```diff
@@ -914,6 +914,78 @@ class ApplicationJob < ActiveJob::Base
...
+         "provider_job_id" => nil,
+    "exception_executions" => {}
+}
+```
+
+You can query via parameters in the `serialized_parameters` field using normal Postgres JSON operations:
+```
+jobs = GoodJob::Job
+  .where("serialized_params ->> 'job_class' = 'DoSomeThingJob'")
+```
```
ManickYoj, line 962: This is 'general' Postgres knowledge, but at least in my experience, it's pretty rare for folks to query into JSON, so I figured including the command here would be helpful.

README.md
```diff
@@ -914,6 +914,78 @@ class ApplicationJob < ActiveJob::Base
...
+
+Other scopes include:
+- `running`
+- `finished` - jobs that have been completed (with or without an error) and will not be re-run
+- `scheduled` - jobs that have not run yet and will run in the future
+- `retried` - jobs that have attempted to run at least once, did not complete, and will run again in the future
+- `queued` - jobs whose scheduled time to run has passed and will run when a thread to run them becomes available
+- `succeeded` - jobs that have run without an error
+- `discarded` - jobs that encountered an error while running and wil not be retried
```
ManickYoj, line 976: I did my best :shrug:. If I didn't quite nail the nuances of each of these, please do correct me.

README.md
```diff
@@ -914,6 +914,78 @@ class ApplicationJob < ActiveJob::Base
...
+
+### Dequeuing Jobs
+
+Jobs can be dequeued by deleting them out of the Jobs table:
+
+```
+GoodJob::Job
+  .job_class("<JobNameHere>") # Filter to a particular job type
+  .where(finished_at: nil) # Filter to enqueued jobs, ignoring those that have previously completed
+  .delete_all # Or `.each(&:destroy)` for Rails callback support (at the expense of speed)
```
ManickYoj, line 986: This is what we have been doing, but please correct me if there's a more 'correct' way to go about this.

bensheldon:
@ManickYoj I appreciate you documenting these, but I'm uncomfortable including them in the Readme. I don't consider the schema or model methods "public" in the sense that they're guaranteed to follow SemVer with clear deprecation cycles.

In terms of writing the Readme, I want to try to start with a specific need/use-case, and then document how to accomplish that. I've learned that extraneous detail can be interpreted as _necessary_ to operating GoodJob, rather than as like "extra extra context if you really need it". 

And on that, I think I want to back up to building a proper interface to address the need which could be a stable API (e.g. `GoodJob.delete_jobs(job_class:, queue_name:, job_arguments:)`). What would have solved your problem?

ManickYoj:
Hi there @bensheldon!

I understand the discomfort around documenting the schema and I'm happy to remove that. The use-case that made it really difficult for us to deal with issues that arise vis-a-vis Goodjob is that - other than reading the source code - we could not find any instructions in the README or elsewhere on how to find and delete jobs. 

If you'd prefer to provide this through a 'blessed' method like `GoodJob.delete_jobs`, then that would work as well if it's okay to document that interface.

Would you like me to attempt to write that method?

If so, I think my main concern would be that I'm not sure how best to match on your suggested `job_arguments` kwarg. The `job_class` and `queue_name` kwargs seem straightforward enough though.

bensheldon:
> Would you like me to attempt to write that method?

Yes please. Don't worry about the job args or anything else that you didn't need during your incident (sorry, I thought you mentioned job args but I made that up).

I assume you're not using the Web Dashboard? As the Dashboard offers ample options for deleting/discarding jobs. That also makes me wonder if this should be an operation of the `good_job` CLI rather than expecting this to be used through the Rails console.

ManickYoj:
We did try to use the web dashboard during our incident but we're still on v2.11.3 which is quite far out of date. We had two issues with the web dashboard. During our incident (which turned out to be due to excessive memory use on our job-worker instances)  the web dashboard was regularly timing out (30s timeout) despite our web-server itself being in good condition. Second, in the version of GoodJob we were using the mass-delete feature was not yet introduced. We are planning to update as a result, of course, in order to get the mass-delete feature :upside_down_face:. I'm aiming for a method as a fallback should the dashboard end up less than responsive again when we need it!

But yes - I am happy to write that method, though it may take me a few days between other things!

bensheldon:
I made a new issue and will close this PR out: https://github.com/bensheldon/good_job/issues/1245

PR #1138, opened 2023-11-02
Update README's optimize queue exaplanations

maestromac:
While reading through the doc, I noticed that one of those explanation was not consistent with the rest.


bensheldon:
Thanks for catching this!

Issue #1140, opened 2023-11-04
Question about jobs


netikular:
I'm working on using fly.io to scale up and down a group of machines for running jobs.  I've got a really bursty job work load.

good_job is working really well to run these tasks. The work that I've done so far is to have an ActiveJob after_perform callback that checks to see if there are any things left in the queue, and if there is nothing left then I send a sig TERM to the current process effecting a shutdown of the fly.io VM.  This works really well to scale down But I'm running into issues where the queue is empty but the last job was pulled off by the last worker and then the worker shuts itself down. Leaving the job in a running state and nothing actually running it.

My question: Is there something that can be done so that the callback will not shutdown if the job runner is actually running a job.  I've been digging around the code base and I've not found a great way to query if a runner is actually processing something at any given time.

Thanks so much for all your work, this is a great too.

bensheldon:
Thanks for opening the issue! I wanted to unpack a few things for further exploration:


1. Fly is great! I'm really curious about how you have it configured for scaling, just cause that sounds cool. 
2. If you're sending a `TERM` to the GoodJob CLI, it should exit gracefully. If you're running GoodJob async in a Puma webserver, it's likely you don't have the worker_exit hooks set up to gracefully exit GoodJob. 
3. Even better than sending a TERM, you could call `GoodJob.shutdown`. This will cause the CLI to gracefully exit. 
4. This sounds like maybe a feature to build into the GoodJob CLI like `--exit-on-idle` (I think Puma just introduced something similar so I'd look at their naming).

Any of those directions seem most fruitful?

netikular:
I'll dig into the GoodJob.shutdown - testing this locally on my Mac the good_job cli never actually exited.  

I saw the loop in the CLI that should exit the application on sig TERM or INT or if GoodJob.shutdown is run. I'll start there to see why it is not exiting on my Mac.  I'll also test on linux to see if the problem is only on the Mac.

The exit on idle would be great feature, it is basically what I'm trying to build. I then use an after enqueue callback to ensure that there are severs running if a job is started.  

If there was going to be a feature like that I would like to investigate if it would be possible to run just the cron component of good job in say a puma process, so that all the job servers could shutdown but cron jobs could still be enqueued.  I'm currently getting around that by running a good_job cli with overmind on the same machine as my web application.  It works well, but I have to ensure that it doesn't monitor the queues that I'm using.

I'll report back here my findings.

netikular:
[...]

PR #1146, opened 2023-11-07
This is a naive implementation of an idle timeout.

netikular:
I've still got to clean up some formatting to put it back the way that it was - I'll update this PR when I get that done.

I've added a `--count-till-idle` command line argument that is the number of seconds a good_job process will stick around.  This is based on the fact that the main loop sleeps for a tenth of a second and the code checks for working threads each time through the loop, if it finds working threads it resets the idle time out.

This is a rough implementation, I've not extended any tests to check for this yet.  But wanted to push up the PR to further the conversation.

I'm currently running this change in production and is working as I expect.

- [x] Return formatting to original

Connects to #1140


bensheldon:
@netikular thanks for doing this! 🙇🏻 

What's the best way to make changes here? I normally would simply push up to your branch, but I don't want to break anything if you're running this particular branch in production.

(I could also leave a bunch of comments on the PR, but I tend to feel like the effort between trying to be specific about the change I want, and just making it and then explaining why after the fact is a better tradeoff).

bensheldon:
In terms of the things i'm thinking of:

- I'd like to track the true "last_job_executed_at" time. That will involve adding that to `GoodJob::Metric` (which is instantiated in a scheduler) and then exposing that through stats.
- Schedulers can also be MultiSchedulers, so the stats needs to be aggregated in a MultiScheduler too (I might even refactor some of that so it's _always_ a multischeduler rather than right now it's only sometimes).
- Rather than exposing all of stats from the capsule, I'd rather more narrowly expose two values: "active_execution_threads_count" and "last_job_executed_at"; I like to keep interfaces as specific as possible (`stats` is not a good example of that)
- I'd like the name to be `--shutdown-on-idle SECONDS`, with matching name in Configuration
- This makes me think that there needs to be a like `Capsule.started?(wait:)` option so that we make sure that the capsule is fully started before counting down (we can probably defer that for now, but I do imagine the lack of that will cause me some support grief at some point).

netikular:
@bensheldon Sorry for the delay. Had to step away from the computer for a few days.

Feel free to make any changes, I'll make the changes that you've recommended in your comment and push them up this morning.

netikular:
Instead of pushing the logic of idle all the way up to the CLI I thought
it would be better in the GoodJob::Capsule class.

I've created a `#idle?` method that does this check based on the last
time a scheduler in the GoodJob::Capsule has run a job.

I ran into an interesting thing that if I check the configuration object
in the initialization of the GoodJob::Capsule it causes a crash because
rails is not initialized yet.  The configuation object can't access the
rails config.

I instead changed it so that in the initialize of the GoodJob::Capsule
the `shutdown_on_idle_enabled` flag is set to false - meaning never
shutdown on idle. Then in `#start` the GoodJob::Capsule checks the
config and sets the flag so that #idle? method will work.

netikular:
@bensheldon the last PR that I put in should make this good to go.

Looking forward to your thoughts.

bensheldon:
@netikular thank you! I'm at RubyConf this week so I'll be a little slow in getting to this. But I really appreciate it!

netikular:
@bensheldon have fun at ruby conf.  I've not been in years!

bensheldon:
I wasn't able to push up to your branch (I think that's because it's the `main` branch), so I made a new PR: #1159

PR #1158, opened 2023-11-24
Skip `RecordAlreadyAdvisoryLockedError` during mass-update action

jmarsh24:
When running mass update discard, there can be a condition where the good_job tries to destroy a record that is locked. I have added a rescue for this error condition, so good_job skips the discard on this job but doesn't raise an error.


bensheldon:
Thank you! https://github.com/bensheldon/good_job/releases/tag/v3.21.2

PR #1159, opened 2023-11-26
Add `--idle-timeout` CLI option to create temporary processes that exit when inactive

bensheldon:
Replaces #1146. Connects to #1140. Naming inspired by https://github.com/puma/puma/pull/3209

The idea here is that, especially in a containerized environment, temporary instances of the GoodJob CLI can be spun up to execute jobs, and then they will automatically exit when no jobs remain to be executed.

Note that this feature is ignorant of the polling interval or future-scheduled jobs. So if the exit-on-idle duration is less than the polling interval and LISTEN/NOTIFY is disabled, it's possible for the process to exit even though jobs exist on the queue.


bensheldon:
Going to rename this from `--exit-on-idle` to `--idle-timeout` to match the name of Puma's option.

PR #1163, opened 2023-11-27
Recreate cron indexes to be conditional

### LINKED ISSUE #1161: Use partial indices for cron_key?
Hi @bensheldon

I observed constantly growing indices:

```
'good_jobs::index_good_jobs_on_cron_key_and_created_at' bloat 35372.5 waste 553 MB
'good_jobs::index_good_jobs_on_cron_key_and_cron_at' bloat 33274.0 waste 520 MB
```

The problem is that the application **does not use the CRON functionality** - thus cron_key and cron_at are always NULL.

Application uses PG version 14.7, good_jobs table has a lot of churn, but except occasional bursts - enqueued jobs count is low (< 100). VACUUMS are triggered regularly, but for some reason those indices' size stay intact. Non-production databases (staging/sandbox) manifest themselves in the same way.

I wonder if other users have also experienced this problem?

Problem can be mitigated with `VACUUM FULL good_jobs` or `REINDEX ...`, but I think the correct solution is to use partial indices:

```rb
add_index :good_jobs, [:cron_key, :created_at], name: :index_good_jobs_on_cron_key_and_created_at, where: 'cron_key IS NOT NULL'
add_index :good_jobs, [:cron_key, :cron_at], name: :index_good_jobs_on_cron_key_and_cron_at, where: 'cron_key IS NOT NULL', unique: true
```

Is there any downside to this approach? If not, I'll be happy to prepare a PR to fix it.

defkode:
fixes #1161

### DIFF
--- a/demo/db/migrate/20231128075428_recreate_good_job_cron_indexes.rb
+++ b/demo/db/migrate/20231128075428_recreate_good_job_cron_indexes.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class RecreateGoodJobCronIndexes < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    reversible do |dir|
+      dir.up do
+        # Ensure this incremental update migration is idempotent
+        # with monolithic install migration.
+        remove_index :good_jobs, [:cron_key, :created_at] if index_exists?(:good_jobs, [:cron_key, :created_at])
+        remove_index :good_jobs, [:cron_key, :cron_at] if index_exists?(:good_jobs, [:cron_key, :cron_at])
+
+        add_index :good_jobs, [:cron_key, :created_at], where: "(cron_key IS NOT NULL)",
+          name: :index_good_jobs_on_cron_key_and_created_at, algorithm: :concurrently
+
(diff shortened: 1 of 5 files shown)


bensheldon:
@defkode sorry, I realized I have so much experience doing this :D 

- I think this can be a single migration.
- Indexes are a little weird to do idempotently because each call needs to be wrapped with an `if index_exists?` check because the migration could fail out somewhere in the middle.
- You'll also need to update the `templates/install` migration
- Then you'll need to run the template generator on the demo app to generate the live database migration files. Then run _those_ which will update the demo app's schema.rb

...so ultimately this should be updates to 4 files:

- lib/generators/good_job/templates/update/migrations/06_recreate_good_job_cron_indexes.rb
- lib/generators/good_job/templates/install/migrations/06_recreate_good_job_cron_indexes.rb
- demo/db/migrate/xxxxx_recreate_good_job_cron_indexes.rb
- demo/db/schema.rb

defkode:
@bensheldon Thanks for hints. PR is updated.

bensheldon:
@defkode thank you! I just made a few tweaks. I'm going to let CI run through 💚 and then it should be done ⭐

bensheldon:
I'm pretty certain this is failing because of a change introduced today on Rails main: https://github.com/rails/rails/pull/50205#issuecomment-1835416318

PR #1171, opened 2023-12-08
[minor] Use symbol form of index name

andyatkinson:
All the other indexes that use an explicit (non-generated) name use symbols, except this one. This is for consistency. I've tested that a version of this gem that includes this change, generates the Good Jobs migration, and that the migration applies as expected.

```sql
-- From generated migration file,
--with this change included (using "path" bundler option in Gemfile)
\di index_good_jobs_on_scheduled_at*
                            List of relations
  Schema   |              Name               | Type  | Owner |   Table
-----------+---------------------------------+-------+-------+-----------
 rideshare | index_good_jobs_on_scheduled_at | index | owner | good_jobs
```

### DIFF
--- a/demo/db/migrate/20200224015928_create_good_jobs.rb
+++ b/demo/db/migrate/20200224015928_create_good_jobs.rb
@@ -15,7 +15,7 @@ class CreateGoodJobs < ActiveRecord::Migration[5.2]
       t.timestamps
     end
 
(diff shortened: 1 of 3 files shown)


bensheldon:
Thank you! I agree pretty minor but good to be consistent 👍🏻 

I think the _correct_ thing is that these should all be strings. I think the `schema.rb` is the correct representation, which is _very_ heavy on the strings. But overall not important enough to change imo. I squash down migrations every major version, so maybe I'll remember to be consistent next time 😅

Issue #1172, opened 2023-12-10
Running GoodJob in production with systemd throws an error due to a wrong communication with WatchDog


tagrudev:
Hello, I am trying to run good_job in production using the systemd service example that was added in https://github.com/bensheldon/good_job/pull/1029, unfortunately there's a problem with the communication between good_job and watchdog.

My setup

```
ruby 3.2.2
good_job 3.21.2 
rails 7.1.1
ubuntu 22.04
```

My goodjob service running as a user service 

```
Description=GoodJob Background Job Processor
After=postgresql.service network.target

[Service]
Type=notify
WatchdogSec=5s

WorkingDirectory=/my/working/directory

Environment=RAILS_ENV=production
ExecStart=/bin/bash -lc 'exec /usr/local/rbenv/shims/bundle exec good_job start'

RestartSec=1s
Restart=always

StandardOutput=journal
StandardError=journal
SyslogIdentifier=goodjob

[Install]
WantedBy=default.target
```

This results in

<img width="1059" alt="Screenshot 2023-12-10 at 22 46 09" src="https://github.com/bensheldon/good_job/assets/438405/b8e832f6-5ed4-4299-90c1-7525dce698c5">

Goodjob is running, but I think there's a problem with the communication with Watchdog.
I think the problem comes from https://github.com/bensheldon/good_job/pull/1029/files#diff-121a5f6ab045b7079f271a0b111a46097545d1d13550bc5b1394d3e7adb94bfbR178  

Looking at the other methods it's missing `return unless exception` is that the problem? adding this locally fixes the issue, but I am not sure if the watchdog communication is ok.

bensheldon:
@tagrudev Nice find! Can you make a PR for that?

PR #1173, opened 2023-12-11
Fix no method error in systemd service

tagrudev:
Related to https://github.com/bensheldon/good_job/issues/1172

While I was dealing with deploying `good_job` to production using `systemd`, I encountered an error in the logs. Overall it's just a missed return check in the `systemd_service`.

### DIFF
--- a/lib/good_job/systemd_service.rb
+++ b/lib/good_job/systemd_service.rb
@@ -11,10 +11,10 @@ module GoodJob # :nodoc:
   #
   class SystemdService
     def self.task_observer(_time, _output, thread_error) # :nodoc:
-      return if thread_error.is_a? Concurrent::CancelledOperationError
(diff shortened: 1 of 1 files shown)


lib/good_job/systemd_service.rb
```diff
@@ -11,10 +11,10 @@ module GoodJob # :nodoc:
   #
   class SystemdService
     def self.task_observer(_time, _output, thread_error) # :nodoc:
-      return if thread_error.is_a? Concurrent::CancelledOperationError
+      return if thread_error.is_a?(Concurrent::CancelledOperationError) || thread_error.blank?
```
bensheldon, line 14: Let's put the nil check first:

```suggestion
      return if !thread_error || thread_error.is_a?(Concurrent::CancelledOperationError)
```

bensheldon (approved):

PR #1175, opened 2023-12-11
Fix latest migration not affecting `GoodJob.migrated?`

Earlopain:
The latest migration was added but the test was sidestepped by just rolling back the latest 2 migrations.

This allows you to be notified in tests/CI about this latest migration.


bensheldon:
Thank you! I goofed on the purpose of `GoodJob.migrated?` being for CI-only; I didn't like that index_exists? has to make a DB query rather than being able to fetch that out of the schema cache.

PR #1180, opened 2023-12-13
Test migrations reverse cleanly

DanielHeath:
Initial cut at the check mentioned in 
https://github.com/bensheldon/good_job/issues/1161#issuecomment-1850281220

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,6 +135,8 @@ jobs:
       - run: chrome --version
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: Update .ruby-version with matrix value
         run: echo "${{ matrix.ruby }}" >| .ruby-version
       # Dependencies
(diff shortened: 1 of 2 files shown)


bensheldon:
Do you think you could fit this test into the existing generators spec? I worry about my ability to maintain a shell script. Here's an example: 

https://github.com/bensheldon/good_job/blob/ae7448254729520a83fedf2d23e75add3500b0e6/spec/generators/good_job/update_generator_spec.rb#L38

DanielHeath:
Shouldn't be too hard to port - I just grabbed the one I already had written.

DanielHeath:
Oh, actually - the test suite will be holding an open database connection, so you can't drop/recreate the database.

bensheldon:
hmmm, I think it might be possible to `disconnect!` the current connection(s):

https://github.com/bensheldon/good_job/blob/ae7448254729520a83fedf2d23e75add3500b0e6/spec/support/database_cleaner.rb#L9-L10

I dunno if that's enough though to fully drop the database 🤞🏻

DanielHeath:
Yeah, likely possible. I probably won't get to look at this in detail until... [checks calendar] early Feb

bensheldon:
I'm going to close because of age, but I do think it would be useful to have a generator spec for this.

Issue #1048, opened 2023-08-22
Fix documentation to always include space in "Active Job", "Active Record", etc.


bensheldon:
Took me a decade to learn that the Rails libraries' names are two separate words, "Active Record" whereas the class names are "`ActiveRecord`"

andyatkinson:
HI @bensheldon I used to write for "This Week in Rails", which was eventually folded into the official Ruby on Rails weblog. This was one of the first things that was taught to me, and that I taught to other new editors. 😁 

I will take a quick look on this one.

bensheldon:
I think we did it! Thank you @andyatkinson 🙇🏻

PR #1182, opened 2023-12-13
Active Record and Active Job name formatting

andyatkinson:
Initial exploration of some formatting changes for consistency with Rails Guides and official sources:
https://guides.rubyonrails.org/index.html

- Active Job over ActiveJob
- Active Record over ActiveRecord

This requires close inspection though because ActiveJob and ActiveRecord can be valid parts of a class name or constant in the GoodJob source code.

https://github.com/bensheldon/good_job/issues/1048


andyatkinson:
@bensheldon Low priority one, but let me know whether this is going in the correct direction. Getting early feedback here before spending too much time making changes that aren't correct or wanted.

app/models/good_job/execution.rb
```diff
@@ -150,9 +150,9 @@ def self.queue_parser(string)
     # @return [ActiveRecord::Relation]
     scope :schedule_ordered, -> { order(coalesce_scheduled_at_created_at.asc) }
 
-    # Get Jobs were completed before the given timestamp. If no timestamp is
-    # provided, get all jobs that have been completed. By default, GoodJob
-    # destroys jobs after they are completed and this will find no jobs.
+    # Get completed jobs before the given timestamp. If no timestamp is
+    # provided, get *all* completed jobs. By default, GoodJob
+    # destroys jobs after they're completed, meaning this returns no jobs.
```
andyatkinson, line 155: Can back this out, but this was a little funky. Do you use capital-j Job a lot in comments to refer to Job instances? Could these usages be lowercase j generic "jobs"?

app/models/good_job/job.rb
```diff
@@ -60,7 +60,7 @@ def table_name=(_value)
 
     scope :unfinished_undiscrete, -> { where(finished_at: nil, retried_good_job_id: nil, is_discrete: [nil, false]) }
 
-    # The job's ActiveJob UUID
+    # The job's Active Job UUID
```
andyatkinson, line 63: This is a UUID that's turned into a string?

bensheldon: Yep, that's how Ruby represents UUIDs:

```ruby
❯ bin/rails c
Loading development environment (Rails 7.1.1)
[1] pry(main)> SecureRandom.uuid.class
=> String
```

app/models/good_job/execution.rb
bensheldon: this is ok 👍🏻 Reading through this, I think there's some historical baggage here given this is in `GoodJob::Execution` but referring to jobs. I think the language is an improvement.

PR #1193, opened 2023-12-19
Fix namespace for `InterruptError` in README

padde:
The example otherwise crashes with:

```
NameError:
  uninitialized constant ApplicationJob::InterruptError
```


bensheldon:
Whoops! Thank you for the fix.

PR #1194, opened 2023-12-19
remove duplicate index

jmarsh24:
`index_good_jobs_on_active_job_id on [:active_job_id]`
`index_good_jobs_on_active_job_id_and_created_at on [:active_job_id, :created_at]`
The first index (index_good_jobs_on_active_job_id) is made redundant by the second index (index_good_jobs_on_active_job_id_and_created_at). This is because the second index can serve the same purpose as the first when queries are made solely on active_job_id. Additionally, the second index can also support queries that involve both active_job_id and created_at.

This removes the redundant index which should improve speed.


bensheldon:
Thank you! I have another PR to do this in #1181 so let me add you as an author on that commit and I'll close this out 🙏🏻

bensheldon:
Merged into main in https://github.com/bensheldon/good_job/commit/2dc07b789007e68966c3ce34ae07880a1670c3a5

Thank you!!

PR #1199, opened 2023-12-25
Fix new migrations to work with `strong_migrations`

dixpac:
I needed to fix new migrations in order for `strong_migrations` to let them pass.

Don't think there was anything wrong with the original approach, I think it is more issue on the `strong_migrations` side, but still there is no reason users should do this manually.

### DIFF
--- a/lib/generators/good_job/templates/update/migrations/02_create_good_job_settings.rb.erb
+++ b/lib/generators/good_job/templates/update/migrations/02_create_good_job_settings.rb.erb
@@ -2,19 +2,15 @@
 
 class CreateGoodJobSettings < ActiveRecord::Migration<%= migration_version %>
   def change
-    reversible do |dir|
-      dir.up do
-        # Ensure this incremental update migration is idempotent
-        # with monolithic install migration.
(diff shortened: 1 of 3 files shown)


bensheldon:
I think this looks good. Any ideas for how to make it fully reversible? (I realize it wasn't reversible before)

bensheldon:
Looking through this in the fullness of time, I notice:

- The migrations aren't reversible.
- They're in conflict now with some Rubocop rules. Like breaking out the `change_table` into individual column changes now conflicts with [Rubocop suggesting](https://www.rubydoc.info/gems/rubocop/0.61.1/RuboCop/Cop/Rails/BulkChangeTable) `change_table bulk: true`. It looks like `strong_migrations` isn't currently compatible with `change_table`; bummer!

bensheldon:
I'm closing this because with GoodJob v4 these migrations have gone away. 

I think generated files will always be tough because they live in the parent-application and thus will be targetable by any linting or code scanning rules. I'll try to do better with Strong Migrations in the future.

Issue #1197, opened 2023-12-20
Adding links/metadata in dashboard


grncdr:
Is there a supported way to extend GoodJob views with additional metadata?

The use case for me is linking to admin backends and observability tools. For example, we already have a centralized log collector and a log viewer that can handle searching by Job ID. I'd love to be able to add a link on the `good_job/jobs#show` view that takes me straight to those logs.

Similarly, we have an admin backend with a search that supports searching by GlobalID (e.g. `"gid://theapp/User/123"`) since job parameters are often serialized as Global IDs, it _could_ be trivial to show an admin link from the job details.

## Idea

I know I _can_ just override `app/views/good_job/jobs/show.html.erb` within my own app, but then I'm on the hook for pulling in upstream changes whenever we upgrade GoodJob. At the same time, I don't want to introduce a heavyweight solution. What do you think about adding an empty partial that is rendered within that view, that users can explicitly override within their applications?

In practice that would look like:

Update `show.html.erb`

```diff
+ <%= render 'custom_details' %>
  <div class="my-4">
    <h5>
      <%= t "good_job.models.job.arguments" %>
```

Add `_custom_details.html.erb`:

```
<%# This partial is intentionally blank. You can override it by adding app/views/good_job/jobs/_custom_details.html.erb in your app. %>
```

With that change, users can add what they need in a way that should impose relatively little maintenance burden for anyone.

bensheldon:
Oh I like this! I also like that it's trivial to override the custom template from the main app. 

For that custom template, I imagine that it should pass in the `GoodJob::Job` parameter. That sort of exposes a model that I don't consider public/stable, so I'm not sure what would be the recommendation for throwing a test around it, but I like the overall idea.

grncdr:
> For that custom template, I imagine that it should pass in the `GoodJob::Job` parameter. That sort of exposes a model that I don't consider public/stable, so I'm not sure what would be the recommendation for throwing a test around it, but I like the overall idea.

My initial thoughts on that:

1. Partials have access to instance variables, so a user can always poke around in `@job`, stable interface or not. The same applies to view helpers. Personally, I think simply documenting "this is experimental, no stable API" is sufficient to start.
2. For my examples, I mostly use `@job.active_job` anyways, so a view helper to return that (strawman: `current_active_job`) would provide most of the necessary API without exposing GoodJob models.
3. However, I also want to extend the view of individual executions, which puts me right back to depending on a non-public model.

[...]

PR #1200, opened 2023-12-28
Add extension partials to good_job/jobs#show view

grncdr:
This is a minimal implementation of the idea from #1197.

I've opted not to address the issues of exposing internal classes in this first pass, but rather just document the potential for unstable APIs. I'm open to ideas, but my opinion at the moment is that this strikes the right balance between complexity/maintenance burden for GoodJob maintainers and users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
     - [Dashboard](#dashboard)
         - [API-only Rails applications](#api-only-rails-applications)
         - [Live polling](#live-polling)
+        - [Extending dashboard views](#extending-dashboard-views)
     - [Job priority](#job-priority)
     - [Concurrency controls](#concurrency-controls)
         - [How concurrency controls work](#how-concurrency-controls-work)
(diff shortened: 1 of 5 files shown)


bensheldon:
Thank you! This looks nice. Could you please add an example in to the `/demo` application and also add a system test for that?

grncdr:
Now updated with an example and a spec.

lib/generators/good_job/install_generator.rb
```diff
@@ -21,6 +21,12 @@ def create_migration_file
       migration_template 'migrations/create_good_jobs.rb.erb', File.join(db_migrate_path, "create_good_jobs.rb")
     end
 
+    def copy_empty_partials
+      %w[custom_job_details custom_execution_details].each do |partial|
+        copy_file "views/_#{partial}.html.erb", "app/views/good_job/jobs/_#{partial}.html.erb"
```
bensheldon, line 26: I think I'd prefer not to copy these over as part of the install generator. Initially I was excited about it, but I try to step lightly inside of other people's applications and I think I'd prefer to document it in the readme, but otherwise not force it on an application.

bensheldon: This also makes me think that the path should either be:

- `views/good_job/_custom_job_details.html.erb` OR
- `views/good_job/jobs/_job_details.html.erb`

...because I think the custom is useful if it's at a simple root path, but otherwise it should just be "override this existing vanilla-looking path"

grncdr:
Thanks for updating the PR @bensheldon, I'm dealing with a newborn here and won't have time to follow up for a while 😅

bensheldon:
@grncdr no worries at all 💖  I appreciate your help moving it this far forward 🙇🏻

gap777:
Woohoo! When do you expect to cut a new release with these changes?

Issue #1139, opened 2023-11-03
Should the table of jobs be a table element?


bensheldon:
When the Dashboard went through heavy redesign, the jobs table was rewritten to not use `<table>` elements. I've wondered if that was the wrong choice for markup and accessibility reasons.

I was just reading this and thought of it: https://adrianroselli.com/2023/11/dont-turn-a-table-into-an-aria-grid-just-for-a-clickable-row.html

PR #1205, opened 2024-01-03
Rewrite Jobs table using table element

sparshalc:
This pr addresses #1139. It rewrites the jobs table using table element.


bensheldon:
@sparshalc thanks for tackling this! It looks like the table currently has a single column which then `.row` and `.col` elements. Can you please replace those with `td` elements so that it is a table with multiple columns?

bensheldon:
Closing this for now. We can reopen when it picks up again.

Issue #1116, opened 2023-10-19
Show environment in dashboard (UI)


bforma:
I often have multiple browser tabs open with GoodJob Dashboard page from different Rails environments (ie. development and production). It would be great if I could somehow easily distinguish between pages from those different environments.

For example, show the environment label in the header (perhaps even differently colored per environment).

bensheldon:
I have been thinking about this, especially after https://github.com/bensheldon/good_job/discussions/1109 was recently suggested.

I'm thinking that, once that secondary navbar exists, the content could be configured in the goodjob configuration e.g.

```ruby
# eval'ed into the partial
config.good_job.dashboard_message = -> { 
  "<strong>Environment:</strong> #{Rails.env}</strong>"
}
```

bforma:
Yes, the secondary navbar would be a good place to put it 👍 . Why would this need extra custom configuration?

bensheldon:
> Why would this need extra custom configuration?

I guess I'm proposing _not_ putting the environment string in it by default, but making it simple for you to insert it for your uses.

PR #1206, opened 2024-01-04
Add environment label to navbar

sparshalc:
This pr addressed #1116 . I have updated the relevant code to ensure that the environment label is visible in the secondary navbar.

image
image


Pauloparakleto:
@bensheldon This is so usefull for a first version. Sometimes I place development and production together when monitoring the jobs. It can avoid lot of mistakes.
@sparshalc does it also play well in dark mode? I would love to know that.
Thanks!

bensheldon:
I like this in principle, but I think the visual design of it needs some 

- Could you explore a way to fit it into the same horizontal line as the other text in the secondary navbar? Vertical space is at a premium and I want to avoid making those navbars any taller.
- I'd like to avoid using alert/warning colors for things that are purely informational (even if they should be high-visibility)
- Another option would be to put this into the Primary Navbar with some sort of [icon](https://icons.getbootstrap.com/) and an abbreviated environment name (e.g. Prod/Dev/first(3))

sparshalc:
@bensheldon i have pushed the changes you requested.

Pauloparakleto:
@sparshalc may you please provide the pictures how it is after the updates, thank you a lot. I am eager to use the new version :D

sparshalc:
As per @bensheldon 's suggestion, I have incorporated the environment details into the primary navbar.

image

bensheldon:
I made a few tweaks so it's always visible regardless of viewport. I also swapped it to a globe because ChatGPT made a very convincing, though possibly ridiculous, argument 😆 

> An appropriate icon to represent the environment (development, staging, production) could be a globe or a world icon. This is because these environments are often associated with different stages of deployment, which can be thought of as different 'worlds'.

Screenshot 2024-01-19 at 4 15 21 PM

Screenshot 2024-01-19 at 4 15 30 PM

mroach:
I just saw this change come through in the dependabot changelog. Nice one!

Would you be open to having a way to set/override the source of this label? In our setup, we have 3 Rails envs: development, test, and *release*. The *release* env itself runs in different stages, namely: edge, demo, production. 
So in our case, `Rails.env` is always *release* outside of local dev and testing. 

For the one or two places in our app where we want to know what actual stage it's running in, we read the `APP_ENV` or `KUBERNETES_NAMESPACE` env var.

What about a config option to allow customising this that would continue to fall-back to `Rails.env`? Then we could do something like:

```ruby
config.good_job.environment_label = ->() { ENV.fetch("APP_ENV") { Rails.env } }
```

Just as I was about to submit this, I saw [this comment](https://github.com/bensheldon/good_job/issues/1116#issuecomment-1771272601) from the linked issue actually recommended making it customisable. This was a nice approach.

PR #1208, opened 2024-01-08
Fix default poll interval in documentation

yenshirak:
It seems seems that the default poll interval is 10: https://github.com/bensheldon/good_job/blob/main/lib/good_job/configuration.rb#L15


bensheldon:
Thank you for catching this!

Issue #585, opened 2022-04-27
Interacting with Dashboard should pause Live Poll


bensheldon:
Described by @reczy in https://github.com/bensheldon/good_job/pull/582#issuecomment-1110028956

> Thoughts on programmatically toggling off live poll if a user checks a checkbox or clicks on an action button? Maybe even if search bar is dirty?
> 
> In trying to weigh possible frustrations from unexpected things happening, I'd say live poll that disrupts a user in the middle of trying to perform some action is probably a bit more annoying than not having the dashboard live update automatically.

Implementation idea: add a listener on all form input changes, and use that to toggle off polling.

PR #1210, opened 2024-01-09
Fix: Interacting with input field should pause the live poll

sparshalc:
The pr adresses #585. This enhancement addresses the frustration users may experience when live polling disrupts their workflow on the dashboard. The update introduces a seamless solution by programmatically toggling off live polling when users makes the use of the search bar.

-  Key Features:
1. Search Bar Interaction: Live polling is now dynamically disabled when users initiate a search. Once the search request is complete, live polling resumes, ensuring a smooth and uninterrupted user experience.

### DIFF
--- a/app/frontend/good_job/modules/live_poll.js
+++ b/app/frontend/good_job/modules/live_poll.js
@@ -3,6 +3,23 @@ import renderCharts from "charts";
 
 const MINIMUM_POLL_INTERVAL = 1;
 const STORAGE_KEY = "live_poll";
+const input_box = document.getElementById('query');
+
+input_box.addEventListener('keyup', () => {
+  closePolling(input_box.value);
+})
+
+function closePolling(value) {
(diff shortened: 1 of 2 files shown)


bensheldon:
@sparshalc I really appreciate you making this PR. It led me to look more closely at the problem and I learned that it could be more simply solved than with adding more javascript:

The filter input was wrapped in a `data-live-poll-region` attribute, which meant that it was _being forced to refresh_. I removed that attribute and it is now safe to edit while the rest of the page is refreshing (i.e. rather than pausing the page refresh, it just doesn't refresh that part of the page).

I also realized that the Navigation Tabs below the filter interface weren't updating, so I fixed that too.

bensheldon:
hmm, I guess the downside of this change is that the Dropdown Selects are no longer updated. That's probably fine 😓 I think the main issue is typing into the searchbar, but I also think it's annoying to have a dropdown change out from under you.

PR #1211, opened 2024-01-09
Discard job with nonexistent job class

yenshirak:
Currently, a NameError is raised when a job with a nonexistent job class is discarded. Job classes may be deleted after the job is scheduled and currently, it's impossible to discard these scheduled jobs from the dashboard.


bensheldon:
Thank you! This is really nice 🙌🏻

PR #1212, opened 2024-01-11
Add Korean translation (ko)

hahwul:
First of all, thank you for creating this fantastic Gem 🥰

I've added the Korean translation, referring to other locale configurations. 
If there are any areas that may need improvement, I would appreciate your feedback! :D


bensheldon:
Thank you! This is fantastic 💖

PR #1220, opened 2024-01-19
Update README, add poll_interval defaults/recommendations

andynu:
These details were provided by @bensheldon himself. I've slightly reworded them and double-checked all the numbers, both defaults and thresholds.


bensheldon:
Thank you!

PR #1226, opened 2024-01-24
Brazilian Portuguese locale

hss-mateus:
A little addition of a pt-BR locale file into the collection.

Thanks for regularly maintaining and improving this gem!

### DIFF
--- a/config/locales/pt_br.yml
+++ b/config/locales/pt_br.yml
@@ -1,0 +1,241 @@
+---
+pt_br:
+  good_job:
+    actions:
+      destroy: Excluir
+      discard: Descartar
+      force_discard: Forçar descarte
+      inspect: Inspecionar
+      reschedule: Reagendar
+      retry: Tentar novamente
+    batches:
+      index:
+        older_batches: Lotes antigos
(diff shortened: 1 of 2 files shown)


jmonteiro (approved):
As a native Brazilian Portuguese speaker I reviewed those translations, and they look good :-)

config/locales/pt_br.yml
```diff
@@ -0,0 +1,243 @@
+---
+pt_br:
```
jmonteiro, line 2: Just as a small note, some apps might use `pt-BR` instead of `pt_BR`. Not sure if or how this can be handled though.

```suggestion
pt-br:
```

bensheldon: Thanks for flagging this! From a code search, it looks like it may not be strictly necessary: https://github.com/search?type=code&q=path%3Aconfig%2Flocales%2F*.yml+%2F%5Ept%5B_-%5Dbr%2F


And seems like both RubyGems.org and Mastodon use `pt-BR` so I'll go with that. I'll make that change.

PR #1229, opened 2024-02-01
Tweak docs for new concurrency default

Earlopain:
Followup to #1145. When I initially wrote this I expected all arguments to be part of the concurrency key by default. Now that that isn't the case anymore it doesn't really make sense to use transient arguments as an example since that is something the user has thought about already when setting a custom key.

I've instead chosen the (probably) common case of `def perform(my_model)` and added a note about what you need to do to make that work. model.to_s is something like `#<MyModel:0x00007f3f634a7320>` which won't work. On it's own this argument isn't allowed but when used in interpolation it will just be cast to a string so I added a note about that as well.


bensheldon:
Thank you!!

Issue #1057, opened 2023-08-30
Load Dashboard nav count badges asynchronously


bensheldon:
When loading the GoodJob dashboard, querying for navigation badge counts is the largest source of request latency. I think this could be improved in 2 ways:

1. Moving the badge counts into an asynchronous request and creating a lightweight stimulus controller that would fetch and insert the values into the HTML. The Stimulus controller should likely live at the `body` level so that it controls the entire page and can make a single request (or maybe 2, one for the global nav, and one specifically for the content).
2. Writing some gnarly Active Record to combine what is currently 13 (!) queries for individual counts into a single (or at least _less_) queries. Ideally it doesn't look like [this unmaintainable block of sql from a previous project](https://github.com/codeforamerica/drai/blob/a0442ee5689f09e6ede1b7eec6f29dd0f0c36f26/app/models/organization.rb#L23). I'm not sure if it would involve CTEs or Unions, but ideally as much as possible stays in Active Record and doesn't drop down to SQL.

![Image](https://github.com/bensheldon/good_job/assets/47554/60dac0fb-075f-40c2-806c-b744e85dbee5)

jeremyf:
First, thank you for writing up the succinct issue and for maintaining a kick-ass ActiveJob adapter.  With Good Jobs we've noticed a stability compared to other options; we don't lose jobs (as in evaporate with no trace) that were in process during a server upgrade or what have you.

My team has definitely noticed the slow-down; and @orangewolf asked that I hop over and find if there's a filed issue (or write one).

I've been tasked with looking into this along with how we would best clean-up our completed jobs.  We've had two issues, observing that if we don't clean-up completed jobs:

1. We see overall performance issues.
2. And the dashboard for good jobs becomes unresponse

jqr:
I haven't had a chance to do it yet, but I think some CTEs that limit the results to something reasonable is an easyish first step.

So instead of showing `10M`, what about `10K+`? Users could configure whatever that limit ought to be, but 10k seems like a very reasonable first guess that should usually be plenty fast.

Alternatively, Rails 7.1 can now do [async count queries](https://rubyonrails.org/2022/5/14/this-week-in-rails-more-async-queries-pattern-matching-and-no-more-autoclosed-prs-c37d3f15) to issue a lot of these in parallel, but this is only going to improve the situation on the margins, when egregiously over reasonable limits this is just going to hammer the database with many simultaneous queries.

bensheldon:
[...]

PR #1231, opened 2024-02-01
Load metrics for top nav asynchronously

binarygit:
PR addresses #1057.

I have only worked on the metrics in the top nav like you mention in the comment in that issue.  I am looking into loading the metrics in the tab navs async too, but I'll wait until you review this implementation.


app/controllers/good_job/metrics_controller.rb
```diff
@@ -0,0 +1,16 @@
+module GoodJob
+  class MetricsController < ApplicationController
+    def primary_nav
+      jobs_count = number_to_human(GoodJob::Job.count)
+      batches_count = number_to_human(GoodJob::BatchRecord.migrated? ? GoodJob::BatchRecord.all.size : 0)
+      cron_entries_count = GoodJob::CronEntry.all.size
+      render json: { jobsCount: jobs_count, batchesCount: batches_count, cronEntriesCount: cron_entries_count }
```
bensheldon, line 7: nice and simple 👍🏻 

This is a personal preference: I'd like these values to be snake_case instead of camelCase

app/frontend/good_job/modules/metrics_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "stimulus"
+export default class extends Controller {
```
bensheldon, line 2: I'd like this Stimulus controller to be written generically as like "AsyncValuesController". I'm imagining,

It would take a `url` value that corresponded to the value being fetched.

It would have `value` targets with a `key` attribute whose value corresponded to a key in the fetched json.

The resulting HTML would look something like this:

```html
<div data-controller="async-values" data-async-values-url="...">
  <div data-async-values-target="value" data-async-values-key="jobs_count"></div>
</div>
```

bensheldon (changes requested):
I like where this is going, but have some suggestions 👍🏻

binarygit:
@bensheldon I apologize for taking sooooo long to make these changes. Past month's been a little of a roller coaster. Could you look at the changes and let me know what you think?

bensheldon (approved):
I made a couple tweaks and this looks fantastic! Thank you so much for doing this.

Issue #1234, opened 2024-02-05
Disable Dashboard Live Polling in production


erick-tmr:
Is there a way to disable Live Polling button in production??

If not, will it be a feature that is desired?

Someone from my team just "forgot" about this button and it was ON it production for about 5 hours and had spiked our response times 5x.

PR #1235, opened 2024-02-07
Allow disabling of Dashboard Live Polling configuration

erick-tmr:
This PR adds a simple configuration to disable dashboard live polling depending on a configuration.

# Considerations
I didn’t make more efforts to hard enforce this rule, like checking this config in JS code or something like this, because Good Job is supposed to be an dev only environment and people should know the consequences of editing HTML and trying to skip some configuration "brute forcibly".
But I'm open to other suggestions.

Issue https://github.com/bensheldon/good_job/issues/1234


app/views/good_job/shared/_navbar.erb
```diff
@@ -60,7 +60,7 @@
 
         <li class="nav-item d-flex flex-column justify-content-center">
           <div class="form-check form-switch m-0">
-            <%= check_box_tag "live_poll", params.fetch("poll", 30), params[:poll].present?, role: "switch", class: "form-check-input" %>
+            <%= check_box_tag "live_poll", params.fetch("poll", 30), params[:poll].present?, role: "switch", class: "form-check-input", disabled: !GoodJob.configuration.dashboard_live_poll_enabled %>
```
bensheldon, line 63: I think we also want to make sure that it is always unchecked when disabled to:

```suggestion
            <%= check_box_tag "live_poll", params.fetch("poll", 30), (GoodJob.configuration.dashboard_live_poll_enabled && params[:poll].present?), role: "switch", class: "form-check-input", disabled: !GoodJob.configuration.dashboard_live_poll_enabled %>
```

bensheldon (commented):
This looks good! Would you mind adding a simple system spec for this? 

(I'm also happy to make that change too if you find it difficult)

app/views/good_job/shared/_navbar.erb
erick-tmr: nice, this looks good

erick-tmr:
> This looks good! Would you mind adding a simple system spec for this?
> 
> (I'm also happy to make that change too if you find it difficult)

Sure, I'll try to add those

erick-tmr:
hey @bensheldon just added the specs, can you take a look again?

erick-tmr:
Looks like some tests failed because of connection timeouts, I can’t retry those by the way :/

bensheldon:
No sweat! This is a new flake, unrelated to your changes.

PR #1256, opened 2024-02-19
[Docs] Bulk.enqueue takes an array of jobs

jpcamara:
This is a simple change to fix the `Bulk.enqueue` interface in the docs

Handing in jobs as arguments, as specified in the README, results in an argument error: `ArgumentError: wrong number of arguments (given 2, expected 0..1)`


bensheldon:
Thanks for fixing that 🙇🏻

PR #1258, opened 2024-02-20
Fix default engine cron value

hss-mateus:
The engine sets an default value for the `cron` configuration, causing it to ignore the ENV value, if it exists ([source](https://github.com/bensheldon/good_job/blob/e502ec024c2883ec4c29f9bde62d098a0264d0d8/lib/good_job/configuration.rb#L209)).

This PR just removes the default value and adds some test cases to ensure the ENV value is considered.


bensheldon (approved):
Thanks for catching this!

(btw, the CI failure is unrelated; I'm fixing that right now and then will accept this)

Issue #1198, opened 2023-12-20
Limiting by executed jobs count by period


Marcovecchio:
I have a situation where I need to limit job execution by a given period of time. Example: I want to limit a given job to 50 executions per minute. When the limit is reached, GoodJob would wait until the time (1 minute) expires, to resume picking new jobs for execution.

The reason for this is that the job will call an external API that rete-limits us.

In other words: is there a way to rate limit GoodJob? Or should I go for implementing this myself in the application?

Thanks!

Eduardo

bensheldon:
This sounds similar to #716, which is not implemented.

I think you could likely do something like this though:

```ruby
class YourJob < ApplicationJob
  LimitExceededError = Class.new(StandardError)
  LIMIT_PER_MINUTE = 50

  retry_on LimitExceededError, retry: :polynomially_longer

  before_perform do
    # It's a little tricky/racy to try to get the perfect count
    # GoodJob::DiscreteExecution is not a public API, so it could change
    recent_executions = GoodJob::DiscreteExecution.where(job_class: self.class.name)
                                                  .where("created_at > ?", 1.minute.ago)
                                                  .where.not(error: "YourJob::LimitExceededError")
                                                  .count
    raise LimitExceededError if recent_jobs > LIMIT_PER_MINUTE
  end
end
```

Marcovecchio:
Thanks a lot! That seems to be a good solution, at least while it's not officially implemented in Good job.

I have only a last question, regarding this implementation: it will raise an exception, to force a retry, right? Is there a way to make it not consume one of the retries? Thinking about an extreme scenario, where the limiter takes a long time to be released, the retries might go exhausted. If I could prevent the limiter to consume retries, that would be perfect! Is there a way to do that?

Thanks again,

Eduardo

bensheldon:
> I have only a last question, regarding this implementation: it will raise an exception, to force a retry, right?

Unfortunately not. the `before_perform` happens after the job has already been dequeued. 

I updated the example to do infinite retries. This is also how the non-throttling Concurrency Control extension works:

https://github.com/bensheldon/good_job/blob/b2f0344a449a2beb39761a0784de09612e5dc0de/lib/good_job/active_job_extensions/concurrency.rb#L45-L49

Marcovecchio:
Many thanks, that will solve our problem! And congratulations for the great job on good job :)

Eduardo

bensheldon:
[...]

Issue #315, opened 2021-07-29
Add an ActiveJob extension for throttling


bensheldon:
Suggested on [Reddit](https://www.reddit.com/r/rails/comments/osngwe/goodjob_has_builtin_cron_and_concurrency_controls/h6s5r9z?utm_source=share&utm_medium=web2x&context=3).

I think this could be done as a `before_perform` hook in which GoodJob queries the previous time-period for a count, and if the value is met/exceeded, then retries the job with an incremental backoff, similar to how the Concurrency extension does it:

https://github.com/bensheldon/good_job/blob/d365e3ef04f8580da40e65dc7c7e47feeb65e196/lib/good_job/active_job_extensions/concurrency.rb#L44-L46
 
I think this maybe should be part of the Concurrency extension so that they would re-use the same key. So maybe it would be:

```ruby
class MyJob < ApplicationJob
  include GoodJob::ActiveJobExtensions::Concurrency

  good_job_control_concurrency_with(
    # New stuff....
    enqueue_throttle: { limit: 1, period: 1.min },
    perform_throttle: { limit: 100, period: 1.hour },

    # Existing stuff...
    # Maximum number of jobs with the concurrency key to be concurrently enqueued
    enqueue_limit: 2,

    # Maximum number of jobs with the concurrency key to be concurrently performed
    perform_limit: 1,

    # A unique key to be globally locked against.
    # Can be String or Lambda/Proc that is invoked in the context of the job.
    # Note: Arguments passed to #perform_later must be accessed through `arguments` method.
    key: -> { "Unique-#{arguments.first}" } #  MyJob.perform_later("Alice") => "Unique-Alice"
  )

  def perform(first_name)
    # do work
  end
end

34code:
why not use activejob-traffic-control? 

https://github.com/nickelser/activejob-traffic_control

I guess it adds a redis/memcached dependency

bensheldon:
@34code thank for suggesting `ActiveJob-traffic_control`! I wasn't aware of that gem; and it's good to know there is an existing option until the feature exists in GoodJob.

I'd like to build it into GoodJob because throttling seems like a useful function of concurrency control, as well as it could be powered by Postgres.

aried3r:
Maybe not directly related, but in case you wanted to look into your own throttling algorithm, schneems wrote a nice article about it, although it not being about job scheduling.

Overview:
https://schneems.com/2020/06/25/rate-limiting-rate-throttling-and-how-they-work-together/

[...]

PR #1270, opened 2024-03-01
Implement throttling options in concurrency extension

marckohlbrugge:
Initial MVP for at throttling extension. I'm not sure I'm qualified to take it the whole way as my understanding of Good Job and some of its architecture is very limited. But creating this PR by @bensheldon's request so we have a place to discuss this particular code.

Credit goes to Ben for his gist where this code is based on: https://gist.github.com/bensheldon/89f772eb8f55058e87959b008e5081b9

Also see #315 #1198

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -96,7 +96,7 @@ module GoodJob
 
               query = DiscreteExecution.joins(:job)
                                        .where(GoodJob::Job.table_name => { concurrency_key: key })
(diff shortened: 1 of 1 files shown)


bensheldon:
@marckohlbrugge Thank you so much for opening this! I'd like to make some tweaks to the interface for throttling. I can push those up and also fixup any of the other lint warnings.

bensheldon:
I really like this! I combined your throttle extension into the existing Concurrency extension.

I also discovered that Rail's TimeHelpers [don't work the way I expected](https://engineering.freeagent.com/2021/03/25/timecop-vs-rails-timehelpers/), so I replaced them with Timecop:

> There is no way to just travel in time without also freezing it when using TimeHelpers.

PR #1287, opened 2024-03-18
Ensure job execution Advisory Lock query uses bind parameters

bensheldon:
- Do not call `to_sql` when generating materialized CTE; use Arel UnaryOperation instead
- Use a `Arel::Nodes::BindParam` instead of a Time value directly

### DIFF
--- a/spec/app/models/concerns/good_job/advisory_lockable_spec.rb
+++ b/spec/app/models/concerns/good_job/advisory_lockable_spec.rb
@@ -21,6 +21,15 @@ RSpec.describe GoodJob::AdvisoryLockable do
       end
     end
 
+    it 'generates bind parameters instead of serialized values' do
+      messages = []
+      allow(model_class.logger).to receive(:debug) do |message|
+        messages << message
+      end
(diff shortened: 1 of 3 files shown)


spec/app/models/concerns/good_job/advisory_lockable_spec.rb
```diff
@@ -21,6 +21,15 @@
       end
     end
 
+    it 'generates bind parameters instead of serialized values' do
+      messages = []
+      allow(model_class.logger).to receive(:debug) do |message|
+        messages << message
+      end
+      model_class.where(priority: 99).order(priority: :desc).advisory_lock.to_a
+      expect(messages.first).to match(/WHERE "good_jobs"."priority" = (\$1|\?)/)
```
bensheldon, line 30: This could be improved byh walking the Arel directly:

https://github.com/rails/rails/commit/ecec50da1d709f9b442c88c2f53a8065e2ed6548#diff-3b43a67ee0006a80ddf7c8b7eb5f6cdfb8b3cfa1a6ef2c65f2b5c70b75736a2bR23-R29

PR #1293, opened 2024-03-21
Set an implicit order on models

### LINKED ISSUE #1242: Set an implicit order on models
Rails 6.0 added a mechanism for setting the order column on models. Because GoodJob uses UUIDs for primary keys, we always need to add an explicit order, so this would make that easier:

https://github.com/rails/rails/pull/34480

mec:
As we use uuids as primary keys the order of objects can be unexpected
without an explicit sort order.

By setting the `implicit_order_column` to `created_at` we get back the
expected behaviour of sequential primary keys.

https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema.html#method-c-implicit_order_column

This work resolves #1242

### DIFF
--- a/app/models/good_job/batch_record.rb
+++ b/app/models/good_job/batch_record.rb
@@ -7,6 +7,7 @@ module GoodJob
     include AdvisoryLockable
 
     self.table_name = 'good_job_batches'
+    self.implicit_order_column = 'created_at'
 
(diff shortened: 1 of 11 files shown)


bensheldon (approved):
@mec Thank you!!! This is sooo nice 🥇

PR #1294, opened 2024-03-21
Add PGHero to the demo app

### LINKED ISSUE #1166: Install PgHero on the Demo app
The demo app tries to touch all the GoodJob features (anything missing?) so it would be nice to get any helpful performance telemetry from it.

mec:
We want to show any helpful performance data in the GoodJob demo app so
we are installing PGHero:

https://github.com/ankane/pghero

We have to add Sprockets to the Rails app so the PGHero assets get loaded
correctly, whilst we could have used a more modern asset pipeline,
Sprockets is well supported and a simple drop in to get PGHero running.

PGHero is available at `/pghero`.

We have not enabled:

- [Suggested indexes](https://github.com/ankane/pghero/blob/master/guides/Rails.md#suggested-indexes)
- [Historical query stats](https://github.com/ankane/pghero/blob/master/guides/Rails.md#query-stats)
- [Historical space stats](https://github.com/ankane/pghero/blob/master/guides/Rails.md#historical-space-stats)

As these may not be suitable for the demo application and can always be
added later.

This work will begin to resolve #1166.

@bensheldon couple of questions:

- are we happy with PGHero being publiclally accessible on the demo app, it's usually not something you would ever want open, but the demo app is a pretty unusual context?
- I wasn't sure about adding the histrorical stats as the docs say "The query stats table can grow large over time" which might be a pain on the demo app - we can always enable them later.


bensheldon:
Let's enable query stats and add a job to clean them up. I copied this from another app:

```ruby
# demo/app/jobs/pg_hero_capture_query_stats_job.rb

# frozen_string_literal: true
class PgHeroCaptureQueryStatsJob < ApplicationJob
  good_job_control_concurrency_with(
    key: "pg_hero_capture_query_stats",
    total_limit: 1
  )

  discard_on StandardError

  def perform
    PgHero.capture_query_stats
    PgHero.clean_query_stats
  end
end


# demo/initializers/good_job.rb
# in the cron config...
pg_hero_capture_query_stats: {
  cron: "*/10 * * * *", # Every 10 minutes
  class: "PgHeroCaptureQueryStatsJob",
  description: "Runs PG Hero maintenance",
}, #...

```

It looks like I can configure BasicAuth on the production demo site:

https://github.com/ankane/pghero/blob/master/guides/Rails.md#authentication

It would be rad if you could add a simple system test that stubbed ENV to enable it, and then verified that basic auth is triggered. But that also sounds like it could be complicated to get working, so not required of this PR.

mec:
Yep, basic auth seems good - let me give it a whirl and see on here.

I am assuming we can trigger the `PgHero.capture_query_stats` with GoodJob as well - let me have a look.

bensheldon:
Thank you!!! 🙇🏻 

> I am assuming we can trigger the PgHero.capture_query_stats with GoodJob as well 

Yep, it's in the job example (in addition to the cleanup)

mec:
So as best I can tell, as soon as we load the PgHero gem it does something that GoodJob does not like at all!

If you watch the demo app started with `RAILS_ENV=development` without PgHero you can see that lots is happening, whereas once PgHero is just available the behaviour is completely different, it just sits there doing nothing (as best I can tell) - this only seems to be the case with `RAILS_ENV=development` as `RAILS_ENV=demo` looks to be working as expected.

The following spec fails: https://github.com/bensheldon/good_job/blob/1acb7218a35b0ff868be8f26cf9407672b6a4c5d/spec/integration/server_spec.rb#L11 because the jobs never launch like they do usually, so 'Enqueued ExampleJob' is not present, again the `RAILS_ENV` has been set to `development` in this spec. The new PgHero specs also fails when not run in isolation returning 500?

Right now I don't have enough context to understand what is happening in development that is so different to cause this, I'll keep looking, but if you have any ideas I am all ears 👂

bensheldon:
@mec thank you for getting so far. I can take a look at it. I have PgHero running successfully on other applications so I imagine there's just something odd somewhere.

spec/system/pghero_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    context 'with the appropriate environment variables set' do
+      before do
+        allow(ENV).to receive(:[]).with('PGHERO_USERNAME').and_return('link')
+        allow(ENV).to receive(:[]).with('PGHERO_PASSWORD').and_return('hyrule')
+      end
+
+      it 'fails without the basic auth credentials' do
+        visit pg_hero_path
+
+        expect(status_code).to eq 401
```
bensheldon, line 16: Unfortunately it looks like this isn't testable because of how basic auth is implemented in PGHero: 

https://github.com/ankane/pghero/blob/6d1c688a0917a4a240c0599b5aecb9b4b8c0c15a/app/controllers/pg_hero/home_controller.rb#L7

Another option would be to enforce a password in all environments _except_ Development (which I think can be done through pghero's config), but I'm also ok just operating on trust and removing this test.

PR #1296, opened 2024-03-23
fix(ui-dropdown): use dropdown-end on locales dropdown

WailanTirajoh:
**Description:**

Add class `dropdown-menu-end` to fix overflow on header locales dropdown open state

**Before:**

image

**After:**

image

**Other options:**

- Set min width to auto based on child element, it would look like this:

image

### DIFF
--- a/app/frontend/good_job/style.css
+++ b/app/frontend/good_job/style.css
@@ -38,3 +38,7 @@
 .btn-outline-secondary {
   border-color: #ced4da; /* $gray-400 */
 }
+
+.min-w-auto {
+  min-width: auto;
+}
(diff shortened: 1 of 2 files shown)


bensheldon:
@WailanTirajoh thank you! This looks great. How are you able to apply the `min-width`? It didn't seem to have an effect when I tried it.

WailanTirajoh:
Cool! updated to use min width instead. Bootstrap has `min-width: 10rem` set to dropdown menu by default, so we need to create custom class `min-w-auto` to set it auto since bs has no min with auto class (1 level with `dropdown-menu` class).

PR #1298, opened 2024-03-25
Invoke inline Bootstrap Color Mode javascript above CSS link tag to better avoid flash of un-themed content

### LINKED ISSUE #1224: Buggy theme animation while switching tabs
flaky-theme-animation

**Link**: [goodjob-demo.herokuapp.com/](https://goodjob-demo.herokuapp.com/)
**Arc Browser**: Version 1.25.1 (45028) Chromium Engine Version 120.0.6099.216
**OS**: macOS 14.2.1 (23C71) M1

bensheldon:
Fixes #1224

Big thank you to @carlosmintfan for pointing out the change.

### DIFF
--- a/app/views/layouts/good_job/application.html.erb
+++ b/app/views/layouts/good_job/application.html.erb
@@ -6,6 +6,20 @@
     <meta content="width=device-width, initial-scale=1, shrink-to-fit=no" name="viewport">
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
+
+    <%# Bootstrap Color Modes
+      "It is suggested to include the JavaScript at the top of your page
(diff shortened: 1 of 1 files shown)


carlosmintfan:
I should however have pointed out that this is only my thought on this and I haven't actually tested it – please test it and don't be angry if it doesn't work :sweat_smile:

bensheldon:
💖 I think this will be an improvement. Though I think there will be no way to entirely prevent a flash: https://github.com/twbs/bootstrap/issues/37774

bensheldon:
I guess the way to entirely avoid this would be to store the value in a server-readable cookie so that the value could be inserted into the html server-side, rather than done entirely in JS in the frontend.

Issue #1305, opened 2024-03-29
PG::SyntaxError: ERROR from timestamp


bananatron:
I couldn't go to /jobs in the admin view, getting:

```ActionView::Template::Error (PG::SyntaxError: ERROR:  syntax error at or near ":"
LINE 1: ... ( good_jobs.scheduled_at ,good_jobs.created_at ) : : TIMEST...
                                                             ^
):
    1: <%= render 'good_job/shared/filter', title: t("good_job.shared.navbar.jobs"), filter: @filter %>
    2: <%= render 'good_job/shared/chart', chart_data: GoodJob::ScheduledByQueueChart.new(@filter).data %>
    3: 
    4: <div data-live-poll-region="jobs-table">
    5:   <%= render 'good_job/jobs/table', jobs: @filter.records, filter: @filter %>
```

[This line](https://github.com/bensheldon/good_job/blob/main/app/charts/good_job/scheduled_by_queue_chart.rb#L27) was the culprit producing the following string:

```
COALESCE (
  good_jobs.scheduled_at
  ,good_jobs.created_at
) : : TIMESTAMP AS scheduled_at
```
I'm not sure what's generating the whitespace between : : TIMESTAMP

- good_job v3.26.1
- rails v7.1.3.2

I upgraded everything without a lot of precision which seems to have fixed the issue, just wanted to past it here for searchability.

bananatron:
Closing this, but here is a hint at a fix if anyone lands here: https://github.com/bensheldon/good_job/pull/1306

jgrau:
Sorry to revive such an old thread but I've had this problem even after the fix in #1306 and still see it in v4.19. I found that the culprit was `pp_sql`. It freedompatches `ActiveRecord::Relation#to_sql` (https://github.com/kvokka/pp_sql/blob/10f03b25d35cc1ab1c4242bb2177de2370967eef/lib/pp_sql.rb#L29-L37) (https://github.com/kvokka/pp_sql/blob/10f03b25d35cc1ab1c4242bb2177de2370967eef/lib/pp_sql.rb#L62-L65)

and that is what made the final good_job query invalid syntax (at least for me)

Fix if you want to keep the gem:

```
PpSql.rewrite_to_sql_method = false if defined?(PpSql)
```

PR #1306, opened 2024-03-29
Remove usage of COALESCE from dashboard chart

bananatron:
- Fixes https://github.com/bensheldon/good_job/issues/1305 which may actually be a non-issue.
- The verbosity may not be worth the advantage given the reproducibility of this issue.


bensheldon:
This is lovely! Any chance you could try putting the entire COALESCE into Arel?

bananatron:
@bensheldon I almost never use Arel so lmk if this is better/worse.

bensheldon:
Sorry, I missed entirely that the Arel was then getting `to_sql`'ed into a larger block of text 🤦🏻 

I'm still unsure of why those spaces got inserted into the SQL previously. I'll accept this and maybe someday it will be fully Arel-ized.

bananatron:
@bensheldon I am also unsure :[ but my theory is some weird level rails dependency issue 🤷

bananatron:
This is coming back with a bundler update, I wonder if the ruby version is somehow involved 🤔 

`ActiveRecord::StatementInvalid in GoodJob::Jobs#index` This prevents me from viewing my jobs page.

```Showing /Users/_/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/good_job-3.28.2/app/views/good_job/jobs/index.html.erb where line #2 raised:

PG::SyntaxError: ERROR:  syntax error at or near ":"
LINE 1: ... ( good_jobs.scheduled_at ,good_jobs.created_at ) : : TIMEST...
```

PR #1311, opened 2024-04-03
Implement enqueue_after_transaction_commit?

### LINKED ISSUE #1310: Adapter to implement `enqueue_after_transaction_commit`
hey Ben 👋🏻 

I'm running an app on edge and noticed my jobs started failing after I updated Rails this afternoon. I tracked it down to [this commit](https://github.com/rails/rails/commit/a134d9166e4d34ab0d97fb82f3e4f0092b3b98ee) and I think [Adapter](https://github.com/bensheldon/good_job/blob/main/lib/good_job/adapter.rb) should implement `enqueue_after_transaction_commit?` too

luizkowalski:
Closes #1310

It defines whether the job should be enqueued implicitly after committing an ActiveRecord transaction. More details [here](https://github.com/rails/rails/commit/a134d9166e4d34ab0d97fb82f3e4f0092b3b98ee)

### DIFF
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -25,7 +25,7 @@ module GoodJob
     #   -+test+: +:inline+
     #  - +production+ and all other environments: +:external+
     #
-    def initialize(execution_mode: nil, _capsule: GoodJob.capsule, enqueue_after_transaction_commit: true) # rubocop:disable Lint/UnderscorePrefixedVariableName
+    def initialize(execution_mode: nil, _capsule: GoodJob.capsule, enqueue_after_transaction_commit: false) # rubocop:disable Lint/UnderscorePrefixedVariableName
       @_execution_mode_override = execution_mode
       @enqueue_after_transaction_commit = enqueue_after_transaction_commit
       GoodJob::Configuration.validate_execution_mode(@_execution_mode_override) if @_execution_mode_override
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -25,9 +25,8 @@ module GoodJob
     #   -+test+: +:inline+
     #  - +production+ and all other environments: +:external+
     #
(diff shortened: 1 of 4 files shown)


luizkowalski:
I'm just not sure how to test it here since this is just an option passed to AJ, and it is being tested there already. Do you have any idea?

bensheldon (commented):
Let's set the default to `false` because we do have the option of transactional integrity. Otherwise this looks good 👍

bensheldon:
And don't worry about tests. We'll trust Active Job.

luizkowalski:
it is worth mentioning that I tested locally with my app, and it worked

luizkowalski:
I am scratching my head trying to fix the failure: using `instance_double` to double `ActiveJob::QueueAdapters::InlineAdapter` doesn't work on Rails edge but works for GA versions. Adding `:enqueue_after_transaction_commit? false` then works on edge but fails on GA since the commit was not backported (nor do I think it will?). I could change to `double` instead. It's not ideal, but it gives us the flexibility to handle different versions of InlineAdapter.

Have you seen such a problem? what's your take on this one? There's only one other place where `double` is used in tests, so I know it is not the preferred way to go

bensheldon:
hmm. It's not pretty but what about something like this?

```ruby
optional_adapter_kwargs = ActiveJob::QueueAdapters::InlineAdapter.method_defined?(:enqueue_after_transaction_commit?) ? { enqueue_after_transaction_commit?: false } : {}
adapter = instance_double(ActiveJob::QueueAdapters::InlineAdapter, enqueue: nil, enqueue_at: nil, **optional_adapter_kwargs)
```

luizkowalski:
that should do it

bensheldon:
Thanks for all the help on this! I decided to lift it up to a global configuration because the reason why someone would use this---they're using a separate database for jobs---would be a global decision rather than per-Adapter. I'll get this merged and cut a release and unblock you 🚀

luizkowalski:
that’s actually a great idea! thanks for the quick follow up, ben! ❤️

bensheldon:
Released! https://github.com/bensheldon/good_job/releases/tag/v3.27.4

PR #1317, opened 2024-04-08
Change ApplicationRecord to ApplicationJob for label documentation.

frans-k:
Maybe I've misunderstood some detail, but I think this should be correct.


frans-k:
Happy to help!

PR #1318, opened 2024-04-08
Add "best practices" section to Readme

bensheldon:
Kind of goofy, but I've had some amazing private conversations with folks lately and wanted to capture it.

cc @drale2k

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,11 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
     - [Write tests](#write-tests)
     - [PgBouncer compatibility](#pgbouncer-compatibility)
     - [CLI HTTP health check probes](#cli-http-health-check-probes)
+- [Doing your best job with GoodJob](#doing-your-best-job-with-goodjob)
+    - [Sizing jobs: mice and elephants](#sizing-jobs-mice-and-elephants)
+    - [Isolating by total latency](#isolating-by-total-latency)
+    - [Configuring your queues](#configuring-your-queues)
+    - [Additional observations](#additional-observations)
 - [Contribute](#contribute)
     - [Gem development](#gem-development)
         - [Development setup](#development-setup)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1527,6 +1528,87 @@ gem 'webrick'
...
+Priority will not help when an elephant is in the doorway. Yes, you could say mice have a higher priority than elephants and always allow any mouse to go _before_ any elephant in queue will start. But once an elephant *has started* going through the door, any subsequent mouse who arrives must wait for the elephant to egress regardless of their priority. In Active Job and Ruby, it’s really hard to stop or cancel or preempt a running job (unless you’ve already architected that into your jobs, like with the [`job-iteration`](https://github.com/Shopify/job-iteration) library)
+
+The best solution is to have a 2nd door, but only sized for mice, so an elephant can’t ever block it. With a mouse-sized doorway _and_ an elephant-sized doorway, mice can still go through the big elephant door when an elephant isn’t using it. Each door has a _maximum_ size (or “latency”) we want it to accept, and smaller is ok, just not larger.
+
+### Configuring your queues
+
+If we wanted to capture the previous 2-door scenario in GoodJob, we’d configure the queues like this;
+
+```ruby
+config.good_job.queues = "mice:1; elephant,mouse:1"
```
elkesrio, line 1572: Hey! 
Do you mean this instead ?
```ruby
config.good_job.queues = "mice:1; elephant,mice:1"
```

drale2k:
The mice and elephant analogy a good one and helps understanding how to properly think about and setup queues. Really like that one and after reading it, it gave me an even better understanding of queues than i already had after our Slack thread. 

To nitpick, this paragraph is a bit of a mouth full. Required me to read it couple of times. It includes a lot of jargon to process:

> Priority can’t fix a lack of capacity. Using priority (i.e. assigning a weight to which jobs or queues should execute first or in which specific order) doesn’t do anything when the capacity to execute jobs is completely utilized/full. Instead, create homogenous workloads (based on latency) and increase execution capacity when queuing latency causes the jobs to exceed their total latency target.

I think that it says is that priority can only be given on an a road with available free lanes. But when all lanes are occupied, assigning priority has no effect. And that the best way to circumvent this problem is to have dedicated lanes for mice and elephants?

README.md
bensheldon: whoops! fixed and thank you! 🙇🏻

bensheldon:
@drale2k that's great feedback! I'll take another look at that sentence.  And maybe steal your wording because this is exactly correct:

> I think that it says is that priority can only be given on an a road with available free lanes. But when all lanes are occupied, assigning priority has no effect.

bensheldon:
Thank you all for the help 💖

PR #1323, opened 2024-04-12
WIP: features to pause batch execution

pgvsalamander:
Pausing jobs works as far as enqueueing/runners, unpausing does work but is done by hand.  Opening this now so I have the diff to look at and to make it easy to discuss implementation.


bensheldon:
I think this looks good so far. I think some Readme-driven-development might be in order here. 

What do you think about an interface like this:

```ruby
my_job = MyJob.set(good_job_paused: true).perform_later(1234)
my_job.good_job_unpause

# or
job_id = my_job.job_id
MyJob.good_job_unpause(job_id) # and an array of IDs, or an array of ActiveJob::Base instances
```

A `GoodJob::Batch` is the most "public" internal API that GoodJob offers, so that might be a good target to implement `#pause`, `#unpause` and `#paused?` methods. I guess the semantics of a batch that is paused is if _any_ jobs in that batch are paused. 

Another thing we'll want to think about is under what job-states a job could be paused (after job creation). Those get a little messy because we'll probably want to take an advisory lock (to prevent pausing a job that is actively executing). But we can defer that too if we just want to focus on creating jobs at create too (we don't have to land _everything_ all at once)

pgvsalamander:
The snippet in my original post was just a proof of concept, I started with pausing/unpausing single jobs just to get a hand on the mechanics.  I'm of the same mind about using the batch for the pause/unpause methods.

I'm keeping the 'pausing an arbitrary batch/jobs' feature in the back of my mind while I'm working, but for the first pass I'm just targeting pausing a batch at creation and adding jobs to it:

```ruby
# I've only skimmed the batch code, my understanding is that the options are automatically distributed to all jobs
# If they aren't, then I'd just use `paused:` for the option as the prefix is a bit redundant in this context
batch = GoodJob::Batch.new(good_job_paused: true)
batch.add do
  10.times { MyJob.perform_later(1234) } # All jobs added to the batch at this point are paused
end
batch.unpause # Jobs will now start running
```

bensheldon:
These don't get distributed as options to a job, they are "properties" of the batch. Do this won't work:

`batch = GoodJob::Batch.new(good_job_paused: true)`

pgvsalamander:
OK!  I still need to document everything and add/update tests (I think the failures are caused by seed data containing NULL values), but I'd appreciate a once-over before that in case there's anything you'd like changed.

app/models/good_job/batch.rb
```diff
@@ -130,6 +130,44 @@ def add(active_jobs = nil, &block)
       buffer.active_jobs
     end
 
+    # TODO: document
+    def paused?
+      # TODO: consider querying to see if any jobs within the batch are paused, and if/how that should be represented if that result does not match properties[:paused]
+      # I think there are probably going to need to be separate methods for "is the batch set to pause all jobs within it" and "does the batch contain any paused jobs" as those cases aren't always lined up
+      properties[:paused] || false
```
bensheldon, line 137: I don't really want to support (at least right now, initially) having a special property of a batch ("paused") have a side effect or meaning to the library; I consider them arbitrary application data.

I realize it makes it a little less optimal to say: "To create a paused batch, enqueue any job that is paused" but I think that's a fine workaround for now. I sort of see the order here being:

1. Design and implement the functionality to allow jobs to be paused/unpaused
  - I think you have all the changes for allowing `scheduled_at` to be nil
  - I think we need to handle the unpausing interface, especially in the dashboard
3. Handle the situation in which a batch has paused jobs in it

And honestly, other than that, this looks great. I can help polish stuff up if whenever you're ok with where things are at.

pgvsalamander:
I noticed there are some situations where jobs will be enqueued with `scheduled_at: nil`, particularly using `GoodJob::Execution.enqueue()` and `GoodJob::Adapter#enqueue_at()`.  I recall you mentioning that GoodJob has been setting `scheduled_at: Time.current` by default since a very old version.  Is the intent that setting those defaults occurs at higher levels and the enqueue methods should accept whatever is handed to them, or have I bumped into an edge case that hasn't mattered previously?

bensheldon:
> Is the intent that setting those defaults occurs at higher levels and the enqueue methods should accept whatever is handed to them, or have I bumped into an edge case that hasn't mattered previously?

I don't consider those public methods and I would expect that job enqueues would go through the Adapter. Though you bring up a good point that I should doublecheck that `scheduled_at` does get set on _everything_ (especially bulk enqueue).

Though reading the code now, maybe I'm wrong and it's changed. Though I do think it's intended given breadcrumbs like this:

https://github.com/bensheldon/good_job/blob/4fd5ac4396a7475976fe76fab0917b5132976fa8/app/models/good_job/execution.rb#L514-L523

We should do a separate PR if there are places that we discover are not reliably setting `scheduled_at`

pgvsalamander:
It looks like `GoodJob::Adapter#enqueue()` sets `scheduled_at: nil` by default and `enqueue_at()` tolerates nil values as well, I'll extract the updates I made here and submit them separately.

app/models/good_job/batch.rb
bensheldon: > Would you be open to storing a batch's paused state in a separate column/hash/etc to avoid the interference, or would you prefer to remove that entirely?

I'd like to defer that. So not one way or another in perpetuity, but for right now it looks like we have a path to allow for a pausing feature without changing any existing interfaces or database columns. This is mostly about my personal ability to handle complexity, not the technical stuff. 

> What functionality did you have in mind for the dashboard? I could see a use for unpausing jobs by hand if we also had the capability to pause jobs by hand, but if they're only paused programmatically I don't see as much benefit.

We'd need to show that the jobs are in a paused state on the dashboard, to answer the question of "why isn't this job running?" And it would be nice to also add options (simple ones) to pause and unpause through the UI.

On the programmitic side, I think the `pause`/`unpause` methods should be class methods on the job. e.g. `OtherJob.unpause(jobs_or_ids)`

pgvsalamander:
There are a handful of tests that assert that `scheduled_at` and `created_at` have the exact same or nearly the same value.  Do you recall if those tests needed to be that strict, or if it was just a way of ensuring that `scheduled_at` was set to a sane value?  Some of the updates to always set `scheduled_at` are setting the value sooner than before, leading to test failures because the values are a few milliseconds off where they were before.

As far as my understanding goes I don't think this should cause any issues, but a spot-check to make sure my understanding is right would be appreciated.

pgvsalamander:
OK, should I skip the ["always set `scheduled_at`" PR](https://github.com/bensheldon/good_job/pull/1323#issuecomment-2067157116) then?  Looking over things again it looks like most if not all of the places I set `scheduled_at` coincide with `is_discrete` logic.

PR #1325, opened 2024-04-17
Store error backtraces on discrete executions

### LINKED ISSUE #1162: Storing backtrace in database?
Loving GoodJob so far, thanks so much for this!

I was wondering if there's an option to store the backtrace (or at least a few lines of it) in the good_job_executions table. It'd be handy to have that information in the database and UI in order to determine the specific cause of a failure, and so that once I fix a bug, I can easily identify all jobs caused by that same bug and retry them. The `error` column often isn't specific enough for me to pinpoint this.

Am I correct in my read that the best way to get the backtrace right now is to set `on_thread_error` and report the error to Rails, which then would have it printed to the logs and/or sent to a tool like Sentry?

Thanks!

Intrepidd:
Closes #1162 

Screenshot 2024-04-18 at 5 43 32 PM

### DIFF
--- a/app/views/good_job/jobs/_table.erb
+++ b/app/views/good_job/jobs/_table.erb
@@ -68,7 +68,7 @@
                 <%= tag.h5 tag.code(link_to(job.display_name, job_path(job), class: "text-reset text-decoration-none")), class: "text-reset mb-0" %>
                 <% if job.error %>
                   <div class="mt-1 small">
-                    <strong class="small"><%=t "good_job.shared.error" %>:</strong>
+                    <strong class="small text-danger"><%=t "good_job.shared.error" %>:</strong>
                     <code class="text-wrap text-break m-0 text-secondary-emphasis"><%= job.error %></code>
                   </div>
                 <% end %>


app/models/good_job/execution.rb
```diff
@@ -455,12 +459,14 @@ def perform
...
         if job_error
           error_string = self.class.format_error(job_error)
+          error_backtrace = self.class.format_backtrace(job_error.backtrace)
 
           job_attributes[:error] = error_string
           job_attributes[:error_event] = result.error_event if self.class.error_event_migrated?
           if discrete_execution
             discrete_execution.error = error_string
             discrete_execution.error_event = result.error_event
+            discrete_execution.error_backtrace = error_backtrace if discrete_execution.class.backtrace_migrated?
```
bensheldon, line 469: Just to save a little bit of unnecessary processing, let's only format the backtrace if it's been migrated:

```suggestion
            if discrete_execution.class.backtrace_migrated?
              error_backtrace = self.class.format_backtrace(job_error.backtrace)
              discrete_execution.error_backtrace = error_backtrace
            end
```

bensheldon (commented):
This is awesome and super clean 👍🏻 

Would you want to also render this out into the Dashboard view too in this PR? 

Otherwise I have one tiny suggestion, and the test that fails is a flake, so this looks awesome 🎉

Intrepidd:
Thanks for the swift review ! 

> Would you want to also render this out into the Dashboard view too in this PR?

You can already kinda see it in the details of an execution in the `_good_job_execution` key.

I figured this is good enough for this PR and could be released as-is to provide some value, but if you have pointers on how you would like it to be displayed as a first class citizen, I'm glad to make a PR for it

bensheldon:
No sweat! I was poking around to form an opinion and just did it 😁 

Thank you for working on this. It's great!

Earlopain:
This is pretty neat, thanks for working on this👍

If I am understanding this corretly, the cleaned backtrace will be saved and only the first frame displayed, is that correct? There have been several occasions where my application backtrace didn't include the full context, for one or another reason.

Would it make sense to store the full backtrace and format only on display? I'm thinking of something like how Rails displays errors in development.

PR #1327, opened 2024-04-19
add missing dropdown-item class

patriciomacadden:
hi! I've been using good_job a lot lately and noticed that when I wanted to mass destroy some jobs the destroy dropdown button didn't have the right style. So here's a small PR to fix that!

Thanks :) 

before:
image

after:
image


bensheldon:
Thanks for catching that! Much better 👍🏻

PR #1328, opened 2024-04-19
Store and optionally display the full execution backtrace

Earlopain:
See https://github.com/bensheldon/good_job/pull/1325#issuecomment-2066050703. This stores the full backtrace and later displays a filtered one by default. The user can optionally switch to a tab for viewing the complete backtrace.

grafik


bensheldon:
Thank you! Huge improvement 🙇🏻

PR #1338, opened 2024-04-25
Clarify PgBouncer Compatibility

isaac:
The compatibility story regarding PgBouncer in the README confused me so this is my attempt at clarification.

I would also like to clarify what the implications of the following paragraph are, but I am still unclear on this.

> GoodJob is not compatible with PgBouncer in _transaction_ mode, but is compatible with PgBouncer's _connection_ mode. GoodJob uses connection-based advisory locks and LISTEN/NOTIFY, both of which require full database connections.

For example: does this mean that the GoodJob dashboard will work fine with PgBouncer, but actually running jobs will not work?

Thanks again for all your work on this project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1383,7 +1383,7 @@ _Note: Rails `travel`/`travel_to` time helpers do not have millisecond precision
 
 GoodJob is not compatible with PgBouncer in _transaction_ mode, but is compatible with PgBouncer's _connection_ mode. GoodJob uses connection-based advisory locks and LISTEN/NOTIFY, both of which require full database connections.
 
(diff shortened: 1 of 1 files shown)


bensheldon:
Thanks for improving the documentation! 🙇🏻 

> does this mean that the GoodJob dashboard will work fine with PgBouncer, but actually running jobs will not work?

The Dashboard uses the same Active Record models as the jobs / executors, so it will/must use the same database connections which won't work with PgBouncer. The Dashboard also can perform some job operations that require taking advisory locks, and those wouldn't work correctly with PgBouncer.

PR #1345, opened 2024-04-30
Fix latest migration not affecting `GoodJob.migrated?`

Earlopain:
The test did not catch this because the migration in the generator migration folder is named incorrectly: it will not be the last migration generated and as such the test just checks that the previous migration was handled (which it is).

I also added a test that ensures that future migrations are named correctly: They must have a unique number prefix. This test fails when the renaming is undone since the prefix 10 is used two times.


bensheldon:
Thanks for catching this! And sorry about the delay; just got back from vacation 🌴

PR #1346, opened 2024-04-30
Don't abort CI jobs when a single one fails

Earlopain:
Some tests appear flaky (at least on CI, or just for JRuby): https://github.com/bensheldon/good_job/actions/runs/8892740760/job/24417466385?pr=1345

This prevents all jobs from aborting when that happens and gives a better idea for the scope of failures.


bensheldon:
Thank you!!

Issue #1053, opened 2023-08-29
Recording accurate job durations


DanielHeath:
Currently, good_job records the start and end time of a job. 

That seems like it should be enough to calculate the job runtime, except... during NTP adjustments it can result in a negative duration. 

Using `Process.clock_gettime(Process::CLOCK_MONOTONIC)` to calculate start/end times would also be sufficient if you didn't want the extra overhead of storing duration.

bensheldon:
This is really interesting! And totally makes sense that we should try to be as accurate as possible for calculating those durations) Some thoughts:

- This makes a lot of sense when calculating execution duration, but would be harder to calculate queue latency because the job could be enqueued and executed on different machines. Hopefully that queue latency is brief so theres less chance of an NTP adjustment 🤞🏻 
- I would like to keep storing timestamps. Would it be weird to say the "finish_time" is `ntp_start_time + monotonic_duration`? I think that would be fine 🤷🏻 
- For jobs that are executed inline or async (i.e. if the operation does take place on the same machine) I think that makes sense to calculate _all_ the timestamps (that we can) using the monotonic clock (though there also _shouldn't_ be any meaningful queue latency in that case)

If that all makes sense, I would accept a PR if you wanted to calculate the duration of a job using the monotonic clock and use that to calculate the finished_at/errored_at (and anywhere else you find; like batches too I think)

DanielHeath:
For latency, the problem is essentially not solvable; postgres lacks a "monotonic timestamp" function, and multi-machine clock consistency should be out of scope.

If all clock calls use `CLOCK_MONOTONIC`, then if they happen to occur on the same instance they will behave sensibly. Might make sense to record the host alongside each timestamp?

bensheldon:
> For latency, the problem is essentially not solvable; postgres lacks a "monotonic timestamp" function

That's a bummer, but also makes sense that would likely be very difficult to implement for a cluster. 

> Might make sense to record the host alongside each timestamp?

I'm generally reluctant to make schema changes. I think we should make best effort to be accurate, but I imagine storing the host alongside timestamps would have very niche usage.

bensheldon:
We have monotonic execution latency, and also can calculate queueing and total latency. It's sorta shown on the dashboard. I'm closing this but everything can always be better.

PR #1362, opened 2024-05-25
Add stats about job classes

arnaudlevy:
Capture d’écran 2024-05-25 à 09 31 58


bensheldon:
@arnaudlevy I'm excited where you're going with this! fyi, there is some discussion in #438 about such a feature. Let me know when you'd like me to review it and give you feedback.

arnaudlevy:
Hey @bensheldon, glad you like it! Maybe what I did there could be a first step. Tell me what you think about the names, so I can walk in your steps and adjust. I put all chart config back in ruby, is that ok?

arnaudlevy:
@bensheldon what do you think about denormalizing `runtime_latency`?
It would be easier to manipulate as a field in the table.

bensheldon:
@arnaudlevy there's a little about storing the latency in #1053.

I would say:
- Go for it 👍🏻 
- Please do it as a separate PR (nows the time because I'm working on the next major GoodJob upgrade and it would be ideal to get in any database migrations before that)
- Store the column as a BigInt as like `duration_ms` on the `good_job_executions` table. I don't think we want to store a float.
- Do it as a monotonic time (get the monotonic start as a Ruby variable and then calculate the total duration in milliseconds and store that)

bensheldon:
> Tell me what you think about the names, so I can walk in your steps and adjust. I put all chart config back in ruby, is that ok?

- Chart config is ok 👍🏻
- I'd like to call this page "Performance" rather than "Statistics" (🚧 Necessary)
- The calculations _must_ be done in the database, rather than by loading all the executions. (🚧 Necessary)
  - We should probably scope it by time ago (1 hour, 4 hours, 1 dday, etc.). When I've spiked on that in the past (but never made as much progress as you!) I had a drop-down that would set a paramter. But I would say just pick one for now and we can always go back and add more options in a later PR
- I'd ideally like this to (eventually, ok in a follow-on PR) be a top level list of queues, that then could be drilled down on to job class. 

Sorry if that's a lot. Only two changes are truly necessary I think (calculate in DB, rename to Performance)

arnaudlevy:
Hey @bensheldon ! 
Renaming to performance was easy, it's done.
I did some SQL limits (start / end), but i'm not too sure what you think about.

Here is the current graph (very suboptimal, as every execution is shown).

Capture d’écran 2024-05-29 à 07 37 10

The idea, for us, is to spot the "slow elephants", the ones that are really long.
What do you think?

arnaudlevy:
Don't know if you noticed @bensheldon but @SebouChu added a fix to the filters

bensheldon:
> @SebouChu added a fix to the filters

Nice! Could you break that out into its own PR? 

On that same note, I'd like to isolate this PR just down to the `PerformancesController#index` controller/view. That would help me really isolate down this feature and we can work on making it performant/possible. Once we get that done/shipped, then focus on additional features for drilling down (don't throw it away! I just want to make it easier to move forward the first piece)

SebouChu:
Hi @bensheldon!

The filters fix is now on this PR if you want: https://github.com/bensheldon/good_job/pull/1373 !

SebouChu:
@bensheldon Just finished adjustments here, mainly having one SQL request for the Performances#index, and focusing on the page only.

Right now, the request gets the execution duration with `EXTRACT(EPOCH FROM (finished_at - created_at))`, but with #1374, it will be simplified so feel free to merge this one so that I can adjust the SQL request here!

SebouChu:
@bensheldon I made the adjustment in the Performance Controller to use duration instead of extracting EPOCH, should be good now!

bensheldon:
I had to open a new PR in order to push up changes to it (it seems like when a PR is opened from a branch named `main` I'm unable to push changes to it 🤷🏻). It looks great and I'm about to merge it 🎉

arnaudlevy:
Thanks @bensheldon, that's really cool!

SebouChu:
Perfect, thanks!

PR #1373, opened 2024-06-13
Fix counts in tab links with filters

SebouChu:
Tab links ignored the current filters, this PR fixes this!


bensheldon (approved):
Thanks! Something is broken with JRuby test dependencies right now. I want to try this locally but looks great!!

SebouChu:
Perfect! I'm currently working on the monotonic duration, PR is coming soon!

PR #1380, opened 2024-06-20
Use newer syntax in documentation for `wait` config on `retry`

benoittgt:
good_job already handle both of course, but doc should provide code that do not produce deprecation warnings like:

```
W, [2024-06-20T23:10:09.948834 #81411]  WARN -- : DEPRECATION WARNING: `wait: :exponentially_longer` will actually wait polynomially longer and is therefore deprecated. Prefer `wait: :polynomially_longer` to avoid confusion and keep the same behavior. (called from <class:TaskWebhookJob> at /Users/benoit.tigeot/projects/tooling/k8s-tooling-gw/app/jobs/task_webhook_job.rb:3)
```

No problem if you think it's better to have doc that work on more version of Rails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,7 +848,7 @@ Active Job can be configured to retry an infinite number of times, with a polyno
 
 ```ruby
 class ApplicationJob < ActiveJob::Base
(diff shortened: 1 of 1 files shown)


bensheldon:
Thanks for catching that!

PR #1383, opened 2024-06-24
fix: require rack/handler for rack >= 3.1.x

shouichi:
From rack >= 3.1.x, `Rack::Handler` is not auto-loaded anymore. Though we should migrate to `Rackup::Handler`, we can't in the short term. This is because of the support of rails <= 7.0 that doesn't depend on rackup.

### DIFF
--- a/bin/tapioca
+++ b/bin/tapioca
@@ -24,4 +24,5 @@ end
 require "rubygems"
 require "bundler/setup"
 
+ENV['BOOTSNAP_CACHE_DIR'] ||= File.expand_path("../demo/tmp/cache/bootsnap", __dir__)
 load Gem.bin_path("tapioca", "tapioca")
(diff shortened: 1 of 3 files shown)


Earlopain:
How about something like rails does it? https://github.com/rails/rails/blob/de4d8744744acab2dd9db0683ccf784ea45810b2/railties/lib/rails/rackup/server.rb

bensheldon:
@shouichi @Earlopain Thank you both! I updated it based on the [Rails PR](https://github.com/rails/rails/pull/47885), which I think will produce the Rack deprecation warning too during the fallback. 

I'm reluctant to add `rackup` to the gemspec because I don't consider the healtcheck "core" functionality, so at a certain point we'll likely need to catch the loaderror and gracefully handle it with a message like "You'll need to add `rackup` to your gemfile."

bensheldon:
Released: https://github.com/bensheldon/good_job/releases/tag/v3.29.5

shouichi:
@Earlopain Thank you for the feedback!

@bensheldon Thank you for the quick release! I confirmed the release works as intended in rails 7.1.

PR #1392, opened 2024-07-06
Add missing newline after frozen string literal for migrations

Earlopain:
Very small nit. All the other current migrations follow the RuboCop rule. It's erb so RuboCop doesn't know about it.

### DIFF
--- a/demo/db/migrate/20240518175057_create_good_job_process_lock_ids.rb
+++ b/demo/db/migrate/20240518175057_create_good_job_process_lock_ids.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 class CreateGoodJobProcessLockIds < ActiveRecord::Migration[7.1]
   def change
     reversible do |dir|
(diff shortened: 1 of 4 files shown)


bensheldon:
Thank you!!

Probably years late to this party, but I just started adding [this](https://github.com/rubocop/rubocop-rails?tab=readme-ov-file#rails-configuration-tip) to my projects:

```ruby
# config/environments/development.rb
  # Run Rubocop after generating files
    config.generators.after_generate do |files|
      parsable_files = files.filter { |file| file.end_with?('.rb') }
      system("bundle exec rubocop -A --fail-level=E #{parsable_files.shelljoin}", exception: true) unless parsable_files.empty?
    end
  end
```

Earlopain:
https://github.com/rails/rails/pull/50506 will expose it nicely with a rails config option, you'll probably like that then. Though I wonder if doing unsafe autocorrect is a sane default :thinking:

bensheldon:
Oh nice! Thanks for sharing that Rails change!

> Though I wonder if doing unsafe autocorrect is a sane default 🤔

Unfortunately, the ones I'm most likely to forget, like Frozen String Literals, are unsafe 😭

PR #1401, opened 2024-07-08
Discrete cleanup

Earlopain:
* `v4_ready?` can always return true on v4. Deprecation?
* Nothing checks against `discrete_support?`/`is_discrete?` anymore
* Benchmark to compare the two doesn't do anything anymore
* Remove some duplicated tests now that they both do the same
* Ignore the `is_discrete`. I don't believe this can safely be dropped beforehand. Probably a v5 thing.


bensheldon:
Thanks for doing this! This is great!

> v4_ready? can always return true on v4. Deprecation?

Yes, I think returning true on v4, and adding a deprecation warning to get people to remove it would be good.

Earlopain:
Alright, added a deprecation for that.

Should I skip `ignored_columns` for `is_discrete` for now? I saw you added that as a task for 4.99 in #1407.

bensheldon:
This looks really good. Thank you for this!

> Should I skip `ignored_columns` for `is_discrete` for now?

I think let's leave it. 

I am a little reluctant to have `ignored_columns` because it changes the SQL queries from a `SELECT *` to explicit columns (but maybe that's good? but still different), but I don't see a safe way to only add it solely in a `x.99` transitional release (because that only allows for a code change, or a migration change, but not both at the same time).

Earlopain:
Yeah, I can see where you're coming from with that. All the additional features introduced previously were nicely guarded behind column checks and whatnot which should be sufficient.

Crossing my finders that you won't have to come back to this one for quite a while yet

PR #1402, opened 2024-07-08
Actually change `smaller_number_is_higher_priority` for v4

Earlopain:
I removed this config value and noticed that it still falls back to the old behaviour.


bensheldon (approved):
🤦🏻

PR #1408, opened 2024-07-08
Fix a rails 6.1 deprecation warning for the duration attribute

Earlopain:
This is only a thing on Rails 6.1, I noticed this in the CI output for the system tests. It doesn't seem to terribly matter right now if it is returned as a string or interval but I don't think this distinction was intended. 

```
DEPRECATION WARNING: The behavior of the `:interval` type will be changing in Rails 7.0
to return an `ActiveSupport::Duration` object. If you'd like to keep
the old behavior, you can add this line to GoodJob::Execution model:

  attribute :duration, :string

If you'd like the new behavior today, you can add this line:

  attribute :duration, :interval
 (called from block (3 levels) in perform at /home/runner/work/good_job/good_job/app/models/good_job/base_execution.rb:440)
```

### DIFF
--- a/app/models/good_job/discrete_execution.rb
+++ b/app/models/good_job/discrete_execution.rb
@@ -14,6 +14,9 @@ module GoodJob # :nodoc:
 
     alias_attribute :performed_at, :created_at
 
+    # TODO: Remove when support for Rails 6.1 is dropped
+    attribute :duration, :interval
+
     def number
       serialized_params.fetch('executions', 0) + 1
     end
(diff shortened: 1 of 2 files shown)


Earlopain:
I tried making specs fail to verify but I believe this is throwing a wrench into that: https://github.com/bensheldon/good_job/blob/d6d1a8c30100a3a28258c660cdca8bcf9b166157/spec/app/models/good_job/job_spec.rb#L30

The deprecation is gone from the system tests though: https://github.com/bensheldon/good_job/actions/runs/9843564859/job/27175035831 vs https://github.com/bensheldon/good_job/actions/runs/9843820235/job/27175887019 so I believe this is fine

bensheldon:
Thank you 🎉

Issue #1397, opened 2024-07-07
Performance Metrics for individual jobs


Earlopain:
Love the new Performance tab, it gives a nice top-level overview into how jobs are behaving. I'm rather spoiled by how sidekiq does metrics so I'm just going to link to that:

https://github.com/sidekiq/sidekiq/wiki/Metrics

I'm especially fond of the stats for individual jobs:

![image](https://github.com/bensheldon/good_job/assets/14981592/0e347adf-0237-4364-815e-023bbe1b718d)

Looks quite nice I'd say.

I tried adding a few simple stats to the performance table but I'm struggling with bootstrap and its 12 column grid system :shrug:.

bensheldon:
I would love something like that. I'd like to have metrics for jobs, and metrics for queues.

> I'm struggling with bootstrap and its 12 column grid system

YES. Sorry. It probably should be a `table`, but good responsive design is hard regardless ❤

Earlopain:
With the linked PR, I consider this done now.

PR #1411, opened 2024-07-09
Latency charts and histograms for individual job classes

Earlopain:
grafik

A bit for #1397 I guess.

I've moved some things around to something that I think makes sense for multiple charts (the one on the homepage and this one now).
* Created a base chart class to inherit from with common helper methods
* Moved most chart structure into ruby. This makes the most sense I think. Change one chart without changing others, access translations.

----

Translations are machine-made and probably bad. I saw some existing german ones that talk about executions in the killing way and jobs in the employment way.

### DIFF
--- a/app/charts/good_job/base_chart.rb
+++ b/app/charts/good_job/base_chart.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module GoodJob
+  class BaseChart
+    def start_end_binds
+      end_time = Time.current
+      start_time = end_time - 1.day
+
+      [
+        ActiveRecord::Relation::QueryAttribute.new('start_time', start_time, ActiveRecord::Type::DateTime.new),
+        ActiveRecord::Relation::QueryAttribute.new('end_time', end_time, ActiveRecord::Type::DateTime.new),
+      ]
+    end
+
+    def string_to_hsl(string)
+      hash_value = string.sum
+
+      hue = hash_value % 360
+      saturation = (hash_value % 50) + 50
+      lightness = '50'
+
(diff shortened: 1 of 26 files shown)


Earlopain:
I also added a histogram for how long jobs take to execute, it looks like this: 
image

You can go there by following the link of each job on the performance landing page. I used the interval values that sidekiq does, seems good enough.

Earlopain:
One other change: I failed to notice that some of the durations got rounded down to flat seconds. I think it's generally useful to show 1-10 seconds with one decimal place if applicable. For example, the performance index is a bit more useful in that range:
grafik

bensheldon (approved):
@Earlopain this looks good to me 🎉 I made one tweak to not directly reflect the query parameter, though it was probably generally safe before (though I'm somewhat paranoid that the charting library wouldn't escape it either because it gets passed in as a chart label), though the less possible content reflection, the better. 

Also, thinking several steps ahead, I'm imagining that the Performances controller might have both per-queue latency, as well as per-job-class runtime, so eventually the URL might want to become `performance/job_class/:name` and `performances/queue/:name` but we'll get there 😄

Earlopain:
Your change looks reasonable to me, I haven't even considered how the chart would handle this. I would say its fine but didn't check of course. The 404 is also an improvement.

I have some thoughts on how to improve this further as well. Like right now, its just 1 day of data and I can imagine a selection for ranges like last 1 hour, 1 week being useful as well but I didn't want to throw that all into one PR. Somehow making the individual bars of the histogram clickable and getting a list of jobs with their arguments that took that amount of time also sounds useful to me, or something else to surface slow slow jobs.

Haven't looked into that yet, I may make a PR for that in the future but who knows. When the muse strikes.

PR #1425, opened 2024-07-16
Fix nonexistant association error between DiscreteExecution and Execution

### LINKED ISSUE #1424: Could not find the inverse association for execution (:discrete_executions in GoodJob::Execution)
After recent update in my application, `good_job start` prints 
`GoodJob error: ActiveRecord::InverseOfAssociationNotFoundError: Could not find the inverse association for execution (:discrete_executions in GoodJob::Execution)` with a [backtrace](https://gist.github.com/tom-kuca/5c91df6c47cee0e08672f57aca1c7d86) multiple times per second and fails to do anything else.

I checked the sources, the relation `discrete_executions` is defined only for `Job`:

https://github.com/bensheldon/good_job/blob/2dcb518f5e12b14e8cc4d11c3b4edde8dd10a94f/app/models/good_job/job.rb#L23

There are two associations in the project that have `discrete_executions` listed as inverse association, they are in `DiscreteExecution`:

https://github.com/bensheldon/good_job/blob/2dcb518f5e12b14e8cc4d11c3b4edde8dd10a94f/app/models/good_job/discrete_execution.rb#L10-L11

The first one looks suspicious to me. Should it be there? `GoodJob::Execution` doesn't have primary key `active_job_id`, association `discrete_executions` is not defined for Execution.

I noticed the error after I updated good_job from `4.0.2` to `4.0.3`. I updated other gems too.

bensheldon:
Fixes #1424

Remove the `belongs_to :execution` association from the `DiscreteExecution` model to fix the inverse association error.

### DIFF
--- a/app/models/good_job/discrete_execution.rb
+++ b/app/models/good_job/discrete_execution.rb
@@ -7,7 +7,6 @@ module GoodJob # :nodoc:
     self.table_name = 'good_job_executions'
     self.implicit_order_column = 'created_at'
 
-    belongs_to :execution, class_name: 'GoodJob::Execution', foreign_key: 'active_job_id', primary_key: 'active_job_id', inverse_of: :discrete_executions, optional: true
     belongs_to :job, class_name: 'GoodJob::Job', foreign_key: 'active_job_id', primary_key: 'active_job_id', inverse_of: :discrete_executions, optional: true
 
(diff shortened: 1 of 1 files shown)


tom-kuca:
The patch fixes the error in my application, `good_job` runs again with this commit.

PR #1430, opened 2024-07-16
Remove references to and ignore `good_jobs.retried_good_job_id` column

bensheldon:


jgarber623:
Apologies if this is noted somewhere, but:

If I'm reading the changes here correctly, are `retried_good_job_id` (and `is_discrete`, it seems) slated for removal (or able to be removed) from the database schema? Stated differently, is a migration forthcoming that drops them?

Based on the description in #1427, I'm also guessing the changes here are v4 refactor related?

bensheldon:
@jgarber623 no worries! 

> If I'm reading the changes here correctly, are retried_good_job_id (and is_discrete, it seems) slated for removal (or able to be removed) from the database schema? Stated differently, is a migration forthcoming that drops them?

Yes, eventually. I think I added them to `ignored_columns` but apparently not in this PR 🤔 I think I may need to wait until v5 to actually drop the columns because it's not possible to guarantee that the release with the ignored columns has been deployed before the migrations are run. So I'm imagining a major version cycle.

jgarber623:
@bensheldon Thanks for the explanation and quick reply! Much appreciated.

Totally makes sense not to add a migration yet to drop those columns. That'd _definitely_ be worth documenting in a major version upgrade guide (like you have with v3 to v4) since it's (maybe? likely?) a destructive irreversible action.

Thanks again for the clarification. 🙌🏻

PR #1432, opened 2024-07-17
German translation pass

Earlopain:
Mostly changing short translations where machine-translations are missing context. I also changed from formal "Sie" (your boss, someone you don't know) to more informal "du" which is more casual. Overall translations should be more consistent.


bensheldon:
Thank you!

PR #1433, opened 2024-07-17
Refactor inline adapter to enable deferred execution after enqueue to allow batch-callbacks to use transaction-based advisory lock

bensheldon:
I'm not quite sure whether the "juice is worth the squeeze" here, but it feels like the more correct thing to do.

Eventually, GoodJob will move to using a combination of row-level locking (to execute jobs) and transaction-level advisory locks (for control/synchronization). To use transaction-level advisory locks (instead of the current connection-based advisory locks), the locks have to be wrapped in a transaction. I would prefer jobs not execute within a transaction, but this presents a problem for Inline Execution mode in places where it's necessary to take a lock, and enqueue a job. Though this is (right now, maybe always) just a problem for the batch-callback job enqueueing. Which means, this is really a bunch of effort solely so that:

- Batch callbacks can be synchronized with a transaction-based advisory lock
- ...and when running in inline execution mode,
- ...the batch-callback job is not executed inline within a transaction.

### DIFF
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -87,17 +87,24 @@ module GoodJob
         end
 
         if inline_jobs.any?
-          @capsule.tracker.register do
-            until inline_jobs.empty?
-              inline_job = inline_jobs.shift
-              perform_inline(inline_job, notify: false)
+          deferred = InlineBuffer.defer?
+          InlineBuffer.perform_now_or_defer do
+            @capsule.tracker.register do
+              until inline_jobs.empty?
+                inline_job = inline_jobs.shift
+                perform_inline(inline_job, notify: deferred ? send_notify?(inline_job) : false)
+              end
+            ensure
+              inline_jobs.each(&:advisory_unlock)
             end
-          ensure
-            inline_jobs.each(&:advisory_unlock)
           end
         end
 
-        non_inline_jobs = jobs.reject(&:finished_at)
+        non_inline_jobs = if InlineBuffer.defer?
+                            jobs - inline_jobs
+                          else
+                            jobs.reject(&:finished_at)
+                          end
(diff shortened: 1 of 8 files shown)


lib/good_job/adapter.rb
```diff
@@ -87,15 +87,23 @@ def enqueue_all(active_jobs)
...
-              inline_job = inline_jobs.shift
-              perform_inline(inline_job, notify: false)
+          InlineBuffer.now_or_defer do
+            @capsule.tracker.register do
+              until inline_jobs.empty?
+                inline_job = inline_jobs.shift
+                perform_inline(inline_job, notify: InlineBuffer.defer? ? send_notify?(inline_job) : false)
+              end
+            ensure
+              inline_jobs.each(&:advisory_unlock)
```
bensheldon, line 97: This fixes a bug introduced in #1426 that should be extracted regardless.

lib/good_job/adapter.rb
```diff
@@ -87,15 +87,23 @@ def enqueue_all(active_jobs)
...
         if inline_jobs.any?
-          @capsule.tracker.register do
-            until inline_jobs.empty?
-              inline_job = inline_jobs.shift
-              perform_inline(inline_job, notify: false)
+          InlineBuffer.now_or_defer do
+            @capsule.tracker.register do
+              until inline_jobs.empty?
+                inline_job = inline_jobs.shift
+                perform_inline(inline_job, notify: InlineBuffer.defer? ? send_notify?(inline_job) : false)
```
bensheldon, line 94: the `notify` logic isn't quite right here 🤔

lib/good_job/adapter.rb
bensheldon: Fixed in #1438

PR #1441, opened 2024-07-18
Remove duplicate word on batches show page

Earlopain:
Only relevant for the mobile layout. This one got translated but the original was accidentally left around.

image


bensheldon:
Nice catch! Thank you!

Issue #1387, opened 2024-07-04
"Couldn't find GoodJob::BatchRecord" error


jesseduffield:
I'm getting this error on some of my batch callback jobs:

{"_aj_globalid"=>"gid://subble-api/GoodJob::Batch/e4b54d15-a2a4-4778-bf3c-5ce1de3f366f"},
 {"event"=>{"value"=>"finish", 
"_aj_serialized"=>"ActiveJob::Serializers::SymbolSerializer"}, 
"_aj_symbol_keys"=>["event"]}

**Error:** `ActiveJob::DeserializationError: Error while trying to deserialize arguments: Couldn't find GoodJob::BatchRecord with 'id'=e4b54d15-a2a4-4778-bf3c-5ce1de3f366f`
```
[...]

PR #1454, opened 2024-08-01
Redefine `Batch#finished_at` to mean all callback jobs have finished too; add `Batch#jobs_finished_at` to allow not deleting batches until all their callback jobs complete

bensheldon:
Connects to #1387

Add logic to delete batches only after their callback jobs have completed.


app/models/good_job/job.rb
```diff
@@ -761,6 +762,7 @@ def reset_batch_values(&block)
 
     def continue_discard_or_finish_batch
       batch._continue_discard_or_finish(self) if batch.present?
+      callback_batch._continue_discard_or_finish(self) if callback_batch.present?
```
bensheldon, line 765: This shouldn't pass the job. If the job is a callback_job, it shouldn't have the ability to discard the batch.

PR #1456, opened 2024-08-02
Add retry functionality for batches

### LINKED ISSUE #1450: Batch callback job not enqueued after success of retried job
Hi Ben,
Great work on GoodJob V4. We have yet to upgrade, but looking forward to using Ordered Queues and Throttling features.

Current good_job version: 3.17
Rails: 7.0

We've been using Batches for a while, and I observed a strange issue. Here's what the code looks like:
```
def perform(batch, params):
    stage = batch.properties[:stage]
    ...
    if stage.nil?
        GoodJob::Batch.enqueue(on_success: ThisCallbackJob, stage: 1) do
            # perform a bunch of jobs
        end
    elsif stage == 1
        # perform another job
    end
end
```
The Callback job is enqueued with stage: 1 when the bunch of jobs succeed, and it is not enqueued when it fails. But when we fix the issue with our data and retry the failed job, it does not proceed to enqueue the Callback job. Am I misunderstanding what `on_success` is supposed to do?
Moreover, the callback job shows up as succeeded, so we cannot retry that to resume it. How can I turn its state to `:discarded`, so that we could retry the callback job when one of the jobs fail?

bensheldon:
Closes #1450

### DIFF
--- a/app/controllers/good_job/batches_controller.rb
+++ b/app/controllers/good_job/batches_controller.rb
@@ -9,5 +9,11 @@ module GoodJob
     def show
       @batch = GoodJob::BatchRecord.find(params[:id])
     end
+
+    def retry
+      @batch = GoodJob::Batch.find(params[:id])
+      @batch.retry
+      redirect_back(fallback_location: batches_path, notice: t(".notice"))
+    end
   end
 end
--- a/app/models/good_job/batch.rb
+++ b/app/models/good_job/batch.rb
@@ -92,10 +92,12 @@ module GoodJob
       if record.new_record?
         record.save!
       else
-        record.with_advisory_lock(function: "pg_advisory_lock") do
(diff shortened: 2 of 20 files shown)


bavani69:
We pulled this ref, made a minor change, and it works well. Thanks a lot, Ben!

bensheldon:
@bavani69 awesome! What did you change?

bavani69:
We simply added a check `raise SomeErrorClass unless batch.succeeded?` and at the start of our batch job, which enqueues on_finish with `batch.enqueue`. With a discard_on said error class for the batch job.
  So if some job from a stage fails, we are able to discard the batch and retry from the dashboard's batches page. With your commit, it now retries all failed jobs and then re-enqueues the batch job with current stage, which is exactly the behaviour we needed.

PR #1420, opened 2024-07-15
Use rails enum for `error_event` and `lock_type`

Earlopain:
A bit more cleanup now that v4 is released.

Rails 7.2 will deprecate the legacy enum notation but the new one was only introduced in Rails 7.0 so a bit of version introspection is needed here.

I also had to notice that proper validations are only available in rails 7.1. I think that's fine, the test matrix will take care of this but its not as nice as I expected it would be.

There was a duplicate `table_name` in `process.rb`, I've removed that in the process of this.

PR #1459, opened 2024-08-06
Fix a few method redefinition warnings

Earlopain:
2 enum values conflict with already existing scopes: "retried" and "discarded" (oops) (#1420)

"running" and "finished" got moved from the discrete execution cleanup (#1427). This change I'm not so certain about which of these two is the correct one. Tests do pass but you'd better check these two anyways, this just removes the earlier definition:

```rb
# first
scope :running, -> { where.not(performed_at: nil).where(finished_at: nil).joins_advisory_locks.where.not(pg_locks: { locktype: nil }) }
# second
scope :running, -> { where.not(performed_at: nil).where(finished_at: nil) }

# first
scope :finished, -> { where.not(finished_at: nil) }
# second
scope :finished, ->(timestamp = nil) { timestamp ? where(arel_table['finished_at'].lteq(bind_value('finished_at', timestamp, ActiveRecord::Type::DateTime))) : where.not(finished_at: nil) }
```

### DIFF
--- a/app/models/concerns/good_job/error_events.rb
+++ b/app/models/concerns/good_job/error_events.rb
@@ -15,9 +15,9 @@ module GoodJob
         discarded: 5,
       }
       if Gem::Version.new(Rails.version) >= Gem::Version.new('7.1.0.a')
-        enum :error_event, error_event_enum, validate: { allow_nil: true }
+        enum :error_event, error_event_enum, validate: { allow_nil: true }, scopes: false
       else
-        enum error_event: error_event_enum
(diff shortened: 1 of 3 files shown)


bensheldon:
Thank you so much for catching these! I must have something wonky in my test setup that I'm not seeing these warnings.

I just chose the other versions 😁 There is already a `finished_before` method that did the same thing as taking an argument.

Earlopain:
Looks good, thanks. You just have to set `$VERBOSE = true` for these to appear (which I tend to do). I would have done this here already but there are two other minor things to get out of the way first.

Issue #1407, opened 2024-07-08
Release GoodJob v5


bensheldon:
This is a placeholder for tracking deprecations and TODOs for whenever this happens, likely far in the future.

## v4.99

- [ ] Add `good_jobs.is_discrete` column to ignored columns list

## v5.0

- [ ] Add migration to drop `good_jobs.is_discrete`

PR #1463, opened 2024-08-09
v5: Update compatibility matrix, remove compatibility code

Earlopain:
v4 updated what it is compatible with but docs weren't updated. I've based these versions on what is written in the v4 upgrade guide.

Honestly not sure if this is possible to do now. While v4 was supposed to drop older postgres versions, I don't think this is enforced anywhere right now and probably still works with at least postgres 11. The rails and ruby constraint at least is enforced through the gemspec.

v4 has been out for about a month now. I would defer the postgres changes for the next major myself, unless there is a check somewhere I didn't find? Let me know what you think.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
         pg: [16]
         include:
           - ruby: 3.3
-            pg: 11
+            pg: 12
 
     env:
       PGHOST: localhost
--- a/README.md
+++ b/README.md
@@ -164,9 +164,9 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
 
 ## Compatibility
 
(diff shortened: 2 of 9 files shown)


bensheldon:
Ugh 🤦🏻 Yes, I agree we should defer this to GoodJob v5 because it would have breakage.

bensheldon:
I noted this in #1407

bensheldon:
@Earlopain Could you send me an email at bensheldon@gmail.com or tell me a good way to get in touch? I wanted to talk to you about maintainership 🙇🏻

Earlopain:
Sure, I can get in touch in a bit when I got more time. I will pick changes from here that wre good to go now then as well

Earlopain:
Just wondering if you got my mail? I should have written you on the same day.

PR #1465, opened 2024-08-09
Update compatibility matrix, remove compatibility code

Earlopain:
Redo of #1463 without the postgres change.

Readme says postgres 10, CI only tests against postgres 11. CI was changed in #828, while v3 was already long released. Checking if this still even works now, otherwise I'll just change the readme to say postgres 11 instead.


bensheldon:
thank you for catching all of my mistakes!

PR #1467, opened 2024-08-12
Update cron documentation to remove confusion about multiple processes

bensheldon:


bradsimantel (approved):
This definitely reads more clearly to me!

PR #1468, opened 2024-08-16
Require fugit >= 1.11.1

jmettraux:
https://github.com/floraison/fugit/issues/104

Prevent Fugit::Nat.parse choking on large input, peg at 256 chars

```ruby
  spec.add_dependency "fugit", "~> 1.11", ">= 1.11.1"
```

Which requires fugit from 1.11.0 to 2.x not included and at least 1.11.1

### DIFF
--- a/good_job.gemspec
+++ b/good_job.gemspec
@@ -47,7 +47,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "activejob", ">= 6.1.0"
   spec.add_dependency "activerecord", ">= 6.1.0"
   spec.add_dependency "concurrent-ruby", ">= 1.3.1"
-  spec.add_dependency "fugit", ">= 1.11.0"
+  spec.add_dependency "fugit", "~> 1.11", ">= 1.11.1"
   spec.add_dependency "railties", ">= 6.1.0"
   spec.add_dependency "thor", ">= 1.0.0"
 


bensheldon:
I don't think this is a dependent gem's responsibility to pin this in their gemspec.  I'm open if you can share documentation/policies somewhere that would recommend that gem's pin secure versions of dependencies. (for example, Rails frequently releases security patches, but I don't pin those in GoodJob's gemsepc)

Specifically for GoodJob, cron-schedules are intentionally designed not to accept user input, so I don't think GoodJob would be effected by malicious values.

jmettraux:
> Specifically for GoodJob, cron-schedules are intentionally designed not to accept user input, so I don't think GoodJob would be effected by malicious values.

Good for you, your project can't be affected then. I'm closing this request.

Thanks for your time.

bensheldon:
thanks for the heads up though! Are you planning to publish a CVE on it? That'd be my recommendation because then automated tooling will pick it up and report it.

jmettraux:
Thanks for the recommendation, I will try to do it.

bensheldon:
There's some details at the bottom of: https://guides.rubygems.org/security/

(disclosure: I work at GitHub) You can do so pretty easily through your GitHub project:  https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/about-repository-security-advisories#cve-identification-numbers

jmettraux:
Thanks! Very kind of you, I was actually looking at https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/publishing-a-repository-security-advisory

jmettraux:
CVE drafted and requested. That was slick so far. Thanks for your guidance.

PR #1475, opened 2024-08-24
fix strict_loading_by_default in BaseRecord

### LINKED ISSUE #1474: Issue with active_record.strict_loading_by_default
If you have `config.active_record.strict_loading_by_default = true` in you `application.rb`

Jobs show throws an error.

image

emilsosa:
Include executions association to prevent `ActiveRecord::StrictLoadingViolationError`

Fixes #1474

### DIFF
--- a/app/controllers/good_job/jobs_controller.rb
+++ b/app/controllers/good_job/jobs_controller.rb
@@ -61,7 +61,7 @@ module GoodJob
     end
 
     def show
-      @job = Job.includes(:executions).includes_advisory_locks.find(params[:id])
+      @job = Job.includes_advisory_locks.find(params[:id])
     end
 
     def discard
--- a/app/controllers/good_job/jobs_controller.rb
+++ b/app/controllers/good_job/jobs_controller.rb
@@ -61,7 +61,7 @@ module GoodJob
     end
 
     def show
-      @job = Job.includes_advisory_locks.find(params[:id])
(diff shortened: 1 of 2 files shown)


Earlopain:
Thanks for the fix!

If I run tests with this config value, I'm getting a few other violations and I'm wondering if those need to be handled as well. Some are test-only but a few happen in library code.

emilsosa:
@Earlopain I'll take a look and try to fix them 👍🏼

Earlopain:
Looks like you can overwrite this per-class with `self.strict_loading_by_default = false`. Might be a thing worth thinking about if it turns out to complex.

In my eyes, you can leave that for a followup issue/pr if you don't want to bother with that. This one right now fixes what you ran into.

emilsosa:
@Earlopain Just now I disabled strict loading for all GoodJob models. As you you said this is the way to go.

Earlopain:
What timing. Yeah, I think this good like that.

Earlopain:
Maybe you can keep your first commit as well? Seems like a good enough improvement to make regardless

emilsosa:
@Earlopain will this be included in the next version?

bensheldon:
Thank you both 🙇 I've been offline the past few days so I'll get this merged and released in the next version (very soon!)

bensheldon:
This has been released: https://github.com/bensheldon/good_job/releases/tag/v4.2.1

PR #1476, opened 2024-08-28
Fix documentation for 2nd Batch callback parameter: consistently call it `context`

martijnversluis:
`options` and `params` are both being used to refer to the second argument passed to the `#perform` method of a callback job. I picked `params`, but I'm fine with both as long as the naming is consistent.


bensheldon:
Thank you for pointing out this inconsistency; the docs are all over the place.

I think this variable should probably be called "context" because it's about the context in which the callback job was enqueued. Currently it only has 1 key (event).

martijnversluis:
Nice! Didn’t get around to it, thanks for fixing!

PR #1482, opened 2024-09-09
Ensure before actions run on root route

ebiven:
When using the recommended method for injecting a before filter into GoodJob the initial root route ignores that filter. In cases where someone is using that before filter to apply security they may want to return a non-301 response code for users who don't pass the authentication check.

For example someone may want to obfuscate that they're using GoodJob by returning a 404 from all GoodJob routes if the user is not logged in. Before this change users would receive the redirect from the root route, then get the 404, essentially "leaking" that the site is running GoodJob.

### DIFF
--- a/app/controllers/good_job/application_controller.rb
+++ b/app/controllers/good_job/application_controller.rb
@@ -26,8 +26,9 @@ module GoodJob
     end
 
     def index
-      # Redirect to the jobs page
-      redirect_to jobs_path
+      # Redirect to the jobs page, maintaining query parameters. This is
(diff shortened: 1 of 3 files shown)


bensheldon:
That's a fantastic observation about leaking info, and a great simple fix. Thank you!

I made one tweak: adding an index action to the ApplicationController would mean that _every_ controller subclass would have an implicit index action, which would be less good. I moved the action to the JobsController; I'm not proud of the name, but it achieves the outcome 👍🏻

PR #1486, opened 2024-09-12
Don't shadow gem loading errors during autoloading with the inline adapter

### LINKED ISSUE #1485: Undefined method `unhandled_error` for nil
See the following:
CleanShot 2024-09-12 at 10 12 27@2x

This errors is happening when using an inline adapter. I am not sure how result can be `nil`, but I happy to provide more information.

GoodJob version: 4.2.1
Rails version: 7.1.4
Ruby version 3.3.5

Earlopain:
`Gem::LoadError` does not inherit from `StandardError` and as such won't be rescued. Just let it bubble up.

That would already happen but since a different error occurs in `ensure`, it was shadowed. I briefly debated rescuing other types of errors [here](https://github.com/bensheldon/good_job/blob/9a781889eb6c66d713695352e4bfd509f96bc70b/app/models/good_job/job.rb#L624) but quickly abandoned that idea.

Closes #1485

----

There's a bit of churn in `todo.rbi`, I regenerated it with `bin/tapioca todo` (apparently that command is deprecated :shrug:)


bensheldon:
I think this change is good. 

I have been slowly introducing more `ensure` usage into my reportoire (rather than rescue and re-raise), so this is useful to see where it fails. 

Aside: seeing the re-raise with fresh eyes, I imagine that might mess up the previous stacktrace. 

> There's a bit of churn in todo.rbi

ugh. I've been manually adding/managing that file (probably should rename it already or note that).

Earlopain:
> I've been manually adding/managing that file (probably should rename it already or note that).

I ended up adding it manually as well. Generating it automatically missed some cases, probably because it doesn't see things in `stub_const`.

> seeing the re-raise with fresh eyes, I imagine that might mess up the previous stacktrace.

It may be fine? At least in tests the exception seems to come from the correct place.

PR #1488, opened 2024-09-14
Add `cron_graceful_restart_period` to avoid missing recurring jobs that occurred during deployment downtime

### LINKED ISSUE #1484: Cron jobs can be missed if deploys are timed just right
TL;DR 
* we deployed our app almost exactly at noon ET, and 
* our cron scheduled for noon did not execute

just glancing at the code in [CronManager](https://github.com/bensheldon/good_job/blob/9a781889eb6c66d713695352e4bfd509f96bc70b/lib/good_job/cron_manager.rb#L11), this seems possible (albeit unlikely), depending on very specific timing

* old worker is shut down with cron JUST about to be run
* new worker spins up and starts cron JUST after cron schedule was due

I would think in a blue-green deployment world, this would not be possible 🤔 (but apparently our heroku worker deployments are not blue-green)

bensheldon:
Closes #1484

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ Rails.application.configure do
   config.good_job.shutdown_timeout = 25 # seconds
   config.good_job.enable_cron = true
   config.good_job.cron = { example: { cron: '0 * * * *', class: 'ExampleJob'  } }
+  config.good_job.cron_graceful_restart_period = 5.minutes
   config.good_job.dashboard_default_locale = :en
 
   # ...or all at once.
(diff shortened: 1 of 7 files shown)


Earlopain:
If I set `preserve_job_records = false` then this will double-enqueue, right?

bensheldon:
Good question! Cron'ed jobs will ignore the `preserve_job_records` configuration and always be preserved (until they're automatically cleaned up): https://github.com/bensheldon/good_job/pull/767

Earlopain:
Ah right, good to know :+1:. I was looking at that piece of code by searching for `preserve_job_records` but missed the relevant condition to the right

tjefferson08:
very cool, thanks for this feature 👍 all your heroku users thank you!

Issue #1490, opened 2024-09-18
Multiple same values in cron_keys_disabled, cron_keys_enabled in good_job_settings table


falegk:
Hello all,

I'm using the .disable, .enable cron entries from the API and I realized that if I call the .disable multiple times the value array fills with multiple same values.
Of course, it doesn't affect the functionality anywhere from what I tried.

Example:

```
GoodJob::CronEntry.find("heartbeat_schedules").disable
=> good_job_settings table has: cron_keys_enabled, ["heartbeat_schedules"]

# call it again
GoodJob::CronEntry.find("heartbeat_schedules").disable
=> good_job_settings table has: cron_keys_enabled, ["heartbeat_schedules", "heartbeat_schedules"]
```
![Screenshot 2024-09-18 at 14 15 51](https://github.com/user-attachments/assets/b6009c3a-2c25-4f8a-94c2-aa02988ff830)

Earlopain:
There's some basic logic which is supposed to prevent this: https://github.com/bensheldon/good_job/pull/1289. Just making sure, what's your good_job version? Apparently this got changed in [3.27.0](https://github.com/bensheldon/good_job/releases/tag/v3.27.0)

falegk:
Hey @Earlopain 

the version is 4.2.0 (updated from previous major versions.)
I can spend more time trying to find out why this happens (check good_jobs code), maybe on the weekend cause my program is full this week.

Earlopain:
PR welcome if you want to do some investigating, and take your time (:
But no worries if you don't want to, let me know if I should take a look instead.

mohammednasser-32:
Heyy, I found this inactive for a while so I jumped into it :sweat_smile: 
Created a [PR](https://github.com/bensheldon/good_job/pull/1498) with the reason of the issue in the description

PR #1498, opened 2024-10-06
Fix GoodJob::Setting duplicate keys

mohammednasser-32:
fix https://github.com/bensheldon/good_job/issues/1490

The problem was that `key` is received as a symbol, while the settings array store the value in the database as strings, so
```ruby
["heartbeat_schedules"].includes?(:heartbeat_schedules)
false
```
so we need to convert the key to string first

I also refactored `cron_key_enabled` test to be consisten with `cron_key_disabled` test


bensheldon:
@mohammednasser-32 thank you! I just had to fix CI on `main`. I'm going to let CI run through (I feel confident it will be green) and I'll merge this.

mohammednasser-32:
@bensheldon Thank you! :pray:

PR #1500, opened 2024-10-08
fix: unable to use keyword args when including the `Labels` extension

### LINKED ISSUE #1350: Unable to use labels with kwargs in job initializer
Hi,
Just trying out labels, and it appears that if I have kwargs in the initializer of my jobs, I can't use labels.

My job:

```
class TestJob < ApplicationJob
    include GoodJob::ActiveJobExtensions::Labels

  def perform(kwarg1:, kwarg2:)
     good_job_labels << did_the_thing? ? 'Foo' : 'Bar'
  end

```

 I get this in the error logs:

```
wrong number of arguments (given 1, expected 0; required keywords: kwarg1, kwarg2)
```

ElMassimo:
[bug]: https://github.com/bensheldon/good_job/issues/1350

### Description 📖

This pull request fixes a [bug] which is described in:

- #1350

### Background 📜

For backwards compatibility, `activejob` uses `ruby2_keywords` to capture arguments that are passed to `perform_now` and `perform_later`.

That ensures that it can serialize and deserialize keyword arguments differently.

`GoodJob::ActiveJobExtensions::Labels` [overrides the `initialize` method](https://github.com/bensheldon/good_job/blob/b714be6f4ad34a643604c54196d9710b28f58b6f/lib/good_job/active_job_extensions/labels.rb#L9), but it does not use `ruby2_keywords`, so by the time `activejob` receives the arguments, keyword arguments have become a plain `Hash`.

### The Fix 🔨

Passing all arguments as-is, by using `...` instead of `*arguments`, since the module doesn't need to use the parameters.

### Notes ✏️ 

Using `ruby2_keywords(:initialize)` would also work, but since this library no longer supports CRuby 2.6 and 2.7, `...` is the way to go.

### DIFF
--- a/lib/good_job/active_job_extensions/labels.rb
+++ b/lib/good_job/active_job_extensions/labels.rb
@@ -6,7 +6,7 @@ module GoodJob
       extend ActiveSupport::Concern
 
       module Prepends
-        def initialize(*arguments)
+        def initialize(...)
           super
           self.good_job_labels = Array(self.class.good_job_labels.dup)
         end


bensheldon:
Thanks @ElMassimo. That makes sense about `ruby2_keywords` being the piece that allows the `initialize(*arguments)` method signature to handle kwargs. 

I'm happy with this change. You're right that we do finally have `...` support 👍🏻

PR #1504, opened 2024-10-08
Better search job arguments (including integers) by adding `serialized_params->>arguments` explicitly

### LINKED ISSUE #1458: Searching for an integer type argument (on the dashboard) does not yeild any results
I encountered an issue while searching for jobs that have two parameters: an integer and a string. When performing a search using the string parameter, the results are accurate. However, no results are returned when searching using the integer parameter.

Additionally, the current test specifications only cover searches with the string parameter, not the integer parameter.

https://github.com/bensheldon/good_job/blob/b0c9707b487652d69264cb145ec8ec9c2aa74263/spec/app/models/concerns/good_job/filterable_spec.rb#L21-L23

Screenshots:
Accurate results when searching with a string query
Screenshot 2024-08-05 at 6 11 21 PM

Nothing when searching for an integer
Screenshot 2024-08-05 at 6 10 20 PM

bensheldon:
Fixes #1458

### DIFF
--- a/app/models/concerns/good_job/filterable.rb
+++ b/app/models/concerns/good_job/filterable.rb
@@ -35,7 +35,7 @@ module GoodJob
         next if query.blank?
 
         # TODO: turn this into proper bind parameters in Arel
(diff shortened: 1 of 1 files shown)


bensheldon:
@swaathi yes, that would be the right thing to do 😉  I'll push a quick test for this.

bensheldon:
@swaathi fyi: https://github.com/bensheldon/good_job/pull/1510

PR #1505, opened 2024-10-09
Clean up appraisals, add rails 8, use postgres 17

Earlopain:
I don't think specifying the two dependencies explicitly is needed anymore. All versions should support the latest variant. Let's see.


Earlopain:
@bensheldon what do you think about using a more elaborate matrix for this? Many versions run in the same job and properly parallelizing would give a nice speed boost and better insight when something fails

Earlopain:
Yeah, I've been thinking about dropping appraisal for doing this as well. Cool that you're already on board with that (:

I'll do that in a separate PR.

Issue #1280, opened 2024-03-10
Notifier errored: ArgumentError: wrong number of arguments (given 1, expected 0)


wbharding:
Hey @bensheldon, long time supporter of the project 👋  Thanks for all the great work you've done over the last couple years. 

Today I upgraded my Rails site that uses good_job to Ruby 3.2, and I used the upgrading occasion to upgrade my good_job from 2.7.4 to 3.26.1. Unfortunately, now when I run `rails s`, I get an endless loop of failure that prevents me from using my app:

```
   (0.1ms)  SET application_name = 'GoodJob::Notifier'
   (0.1ms)  LISTEN good_job
[GoodJob] Notifier subscribed with LISTEN
   (0.1ms)  UNLISTEN *
[GoodJob] Notifier unsubscribed with UNLISTEN
[GoodJob] Notifier errored: ArgumentError: wrong number of arguments (given 1, expected 0)
[...]

PR #1507, opened 2024-10-10
Better parallelize CI jobs

Earlopain:
Drop appraisal while we're at it. Makes it a bit easier


.github/workflows/test.yml
```diff
@@ -186,13 +174,13 @@ jobs:
         uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: screenshots
+          name: screenshots-${{ matrix.pg }}-${{ matrix.gemfile }}-${{ matrix.ruby }}
```
bensheldon, line 177: you fixed a longstanding annoyance! 🙌🏻

.github/workflows/test.yml
```diff
@@ -86,22 +86,39 @@ jobs:
           path: demo/log
 
   test:
-    name: Test
+    name: Test - Ruby ${{ matrix.ruby }} - Postgres ${{ matrix.pg }} - ${{ matrix.gemfile }}
```
bensheldon, line 89: I think Rails is likely to be the culprit in most matrix-related failures:

```suggestion
    name: Test - ${{ matrix.gemfile }} - Ruby ${{ matrix.ruby }} - PG ${{ matrix.pg }}
```

bensheldon (approved):
💯 🥳 🥇 

This is amazing!!!! THANK YOU!! You cleaned up so many problems and it's so much better now (and TIL `eval_gemfile`).  

I'm a little surprised that there aren't more gem tweaks that are necessary because I felt like I was frequently struggling with Appraisal, so this is great!

bensheldon:
btw, that failure with Rails 6.1 is sorta known: https://github.com/bensheldon/good_job/issues/1280

It's fixed by pulling from the Rails 6.1-stable git branch, though I'm a little surprised it shows up now rather than before.

Earlopain:
> btw, that failure with Rails 6.1 is sorta known:

6.1 previously only ran with Ruby 3.0 which explains why it didn't pop up before. I could add excludes to the matrix to preserve that but just pulling 6-1-stable seems fine for CI. WDYT?

Earlopain:
For `eval_gemfile`, it sorta is/isn't actually endorsed for usage by the bundler maintainers https://github.com/rubygems/rubygems/pull/6292#issuecomment-1752938643 https://github.com/rubygems/bundler-site/pull/1081. It makes things so very simple though and don't believe it'll disappear (and if it does, let's just handle that then)

(I'm hugely surprised that this PR turned out well with just one try. Usually my interactions with github actions require much more force-pushes :upside_down_face:)

bensheldon:
🫶 thank you!

PR #1512, opened 2024-10-10
Remove rack from the gemfile

Earlopain:
It was initially added in https://github.com/bensheldon/good_job/pull/987 but the reasoning is lost on me.

Rails 6.1 and 7.0 only work with Rack 2 which causes dependabot update PRs to break CI when it bumps to Rack 3.

Rails head will also switch to be Rack 3 only at some point. Just let bundler resolve to what works.

GoodJob does have to do some version-specific rack handling but I don't think this really needs to be part of the matrix.


bensheldon:
thank you 💖 Glad this works without the explicit dependency 🙌🏻

PR #1514, opened 2024-10-10
Manually bump tapioca / regenerate rbi

Earlopain:
Dependabot does not do this because it resolves with Ruby 3.0 from the gemspec and tapioca now only supports 3.1

This unblocks prism 1.0 which is required for the ruby-lsp to update to the latest version. Currently it runs into https://github.com/Shopify/ruby-lsp/issues/2630, causing a restart loop.

To make sure this does the right thing on older rubies, I added a step to print the resolved lockfile on CI


bensheldon:
Thanks for noticing that!

Earlopain:
It was quite hard not to, the persistent notifications from my IDE about this were quite noticable. I think this dependabot issue is about cases like this: https://github.com/dependabot/dependabot-core/issues/4987, though the situation is a bit different with the lockfile being used for both library and application (the demo).

Issue #1390, opened 2024-07-06
Fix JRuby builds


bensheldon:
These are very much outside my control or expertise :-( Though they do appear to be in the pipeline:

https://github.com/jruby/jruby/issues/7059#issuecomment-2192456888

createdbypete:
After seeing the flaky test in my recent PR I thought worth spending an evening to understand what was going on. JRuby is not something I have much experience with but here's a diagnosis of that particular issue:

`setup-ruby@v1` is now resolving `ruby: jruby-10` to `jruby-10.1.0.0`, which removed the `org.jruby.RubyHash.<init>(Ruby, int)` constructor. Released `activerecord-jdbc-adapter 72.1` (2025-09-26) still calls it from `RubyJdbcConnection#mapRawRow`, so every spec file errors at load:

```
Java::JavaLang::NoSuchMethodError:
  'void org.jruby.RubyHash.<init>(org.jruby.Ruby, int)'
# arjdbc.jdbc.RubyJdbcConnection.mapRawRow(RubyJdbcConnection.java:3562)
```

Upstream fix was merged in [jruby/activerecord-jdbc-adapter#1210](https://github.com/jruby/activerecord-jdbc-adapter/issues/1210) (2026-04-27) but no released arjdbc version contains it yet (latest on rubygems is still 72.1). Verified locally with a docker-compose (postgres:18 + parameterized `jruby:<tag>-jdk21`, isolated bundle volumes per JRuby tag, identical `arjdbc 72.1-java` resolved both legs):

| JRuby tag | Result |
|---|---|
| `jruby:10.1.0.0-jdk21` | spec load fails with the trace above |
| `jruby:10.0.5.0-jdk21` | full unit-test rspec passes — 740 examples, 0 failures, 1 pending |

A one-line workflow pin on `test.yml:103` works around it until arjdbc cuts a 72.2+ release. Branch on my fork: [createdbypete/good_job@9e325c4](https://github.com/createdbypete/good_job/commit/9e325c433cc09a09eecbcb31c59bb211767fc3fd). The trade-off is that the row stops exercising newer JRuby 10 releases until someone manually bumps the pin. Easy to forget once arjdbc 72.2 ships.

Alternatively, `continue-on-error: true` might be appropriate to keep the matrix row visible but non-blocking, or of course dropping official support for this version combination.

Happy to open a PR for whichever direction you'd prefer (the pin, `continue-on-error`, or removing the matrix row).

[...]

PR #1516, opened 2024-10-13
Fix state update + Add Memory usage 

noma4i:
Hi!

I have found that the code around process refresh was shadowed by the `reload` instruction. So basically state filed was never refreshed in the database. For this case I have added additional line in the spec file.

Initially the idea was to add memory usage to the process state. Like that:
image

This change uncovered the state update issue.

### DIFF
--- a/app/models/good_job/process.rb
+++ b/app/models/good_job/process.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'socket'
+require 'get_process_mem'
 
 module GoodJob # :nodoc:
   # Active Record model that represents a GoodJob capsule/process (either async or CLI).
@@ -83,6 +84,7 @@ module GoodJob # :nodoc:
       {
         hostname: Socket.gethostname,
         pid: ::Process.pid,
+        memory: GetProcessMem.new.mb.to_i,
         proctitle: $PROGRAM_NAME,
         preserve_job_records: GoodJob.preserve_job_records,
         retry_on_unhandled_error: GoodJob.retry_on_unhandled_error,
@@ -98,8 +100,8 @@ module GoodJob # :nodoc:
     end
 
     def refresh
-      self.state = self.class.process_state
       reload # verify the record still exists in the database
+      self.state = self.class.process_state
       update(state: state, updated_at: Time.current)
     rescue ActiveRecord::RecordNotFound
       @new_record = true
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,6 +6,7 @@ PATH
       activerecord (>= 6.1.0)
       concurrent-ruby (>= 1.3.1)
       fugit (>= 1.11.0)
+      get_process_mem (>= 1.0.0)
       railties (>= 6.1.0)
       thor (>= 1.0.0)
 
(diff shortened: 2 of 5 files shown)


Earlopain:
Hi there, thanks for the PR!

I don't love the idea of adding a dependency for this, especially since it is dashboard only and memory usage is usually better tracked through other tools.

That said, how about something like this:
```rb
def process_pss(pid)
data = File.read("/proc/#{pid}/smaps_rollup") pss_line = data.lines.find { |line| line.start_with?("Pss:") } pss_line.split(" ")[1].to_i * 1024 end
```

If that file doesn't exist on macos (you seem to use it, good opportunity to find out), probably need to do some switching based on platform like sidekiq does here (and also to not break on windows): https://github.com/sidekiq/sidekiq/blob/d6354bc74204004fc8b446c96875a333c50d27eb/lib/sidekiq/launcher.rb#L227

It will be unix only but that seems perfectly fine.

noma4i:
Hi @Earlopain!
GoodJob depends pretty much on Ruby on Rails so will have all deps already(activesupport depends on bigdecimal and ruby-vips on ffi). Idea to use `get_process_mem` comes from some comments from @bensheldon on the issues here.

bensheldon:
@noma4i thanks for putting together this PR! 

I'm in agreement with @Earlopain about the reluctance to add dependencies (and unfortunately GoodJob's java CI build [are broken at the moment](https://github.com/bensheldon/good_job/issues/1390) so I'm not able to verify it right now). Also seeing it now, I have a thought of "gosh, I wish that was PSS rather than RSS" (which matters less for the `$ good_job` command, but does when running in Puma workers)

I wonder:

- Can the functionality be vendored? similar to vendoring systemd integration
- Can the functionally be added by the application (e.g. if `get_process_mem` gem is installed by the application, then this adds in that behavior)
- Could it be easier for the application to more easily add to the process state it collects via an override/extension or configuration?

noma4i:
Updated. @bensheldon I have used the sidekiq approach to copy same multiplatform behaviour.

app/models/good_job/process.rb
```diff
@@ -12,6 +12,22 @@ class Process < BaseRecord
     STALE_INTERVAL = 30.seconds
     # Interval until the process record is treated as expired
     EXPIRED_INTERVAL = 5.minutes
+    PROCESS_MEMORY = case RUBY_PLATFORM
+                     when /linux/
+                       ->(pid) {
+                         File.readlines("/proc/#{pid}/status").each do |line|
```
Earlopain, line 18: Can you switch this over to `/proc/#{pid}/smaps_rollup`? The older status file doesn't contain PSS which will be more accurate in certain scenarios: https://www.baeldung.com/linux/process-memory-management#pss-memory. It would be nice if you could check if MacOS `ps` also supports pss output.

Sidekiq doesn't have to worry about such things I believe (except maybe in enterprise which supports forking?). It has already been available for almost a decade, but to be safe lets wrap it in rescue ENOENT. The code should be the same except that you're checking for `Pss:`

noma4i: Fixed. Macos doesn't have `proc` concept while `man ps` doesn't have any Pss mentioned there.

noma4i:
@bensheldon any other change you want to see in this PR?

bensheldon:
@noma4i would you be able to fix the lint errors? (`bin/lint`)

bensheldon:
I fixed it. I'll poke at this a tiny bit and then looks good 👍🏻

PR #1525, opened 2024-10-18
Fix badge color for running jobs

### LINKED ISSUE #1518: Modify error color for Running tab when a job has 1 attempt
## Problem
When a job is running and has 1 attempt, the badge with Attempts has an error color
Screenshot 2024-10-14 at 7 45 38 AM

## Proposed Solution
For this tab when the attempt count is equal to 1, show a neutral gray color like the Succeeded tab

Wittiest:
## Problem
When a job is running and has 1 attempt, the badge with Attempts has an error color
Screenshot 2024-10-14 at 7 45 38 AM

## Solution Implementation
- Resolves #1518 by only using the error badge pill when a job has an attached error (from previous execution)
- Removes usage of `GoodJob::Job#recent_error` from Batches show page and applies the same logic there

### Running Jobs Tab
Screenshot 2024-10-17 at 9 09 15 PM

### Batch Jobs Page
Screenshot 2024-10-17 at 9 12 43 PM

### DIFF
--- a/app/views/good_job/batches/_jobs.erb
+++ b/app/views/good_job/batches/_jobs.erb
@@ -35,13 +35,13 @@
               </div>
               <div class="col-4 col-lg-1 text-lg-end">
                 <div class="d-lg-none small text-muted mt-1"><%= t "good_job.models.job.attempts" %></div>
-                <% if job.executions_count > 0 && job.status != :finished %>
+                <% if job.error && job.status != :finished %>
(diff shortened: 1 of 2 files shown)


bensheldon:
I discovered that `:finished` isn't a valid `Job#status` but I was able to use `finished?`. I think I also was able to cover the logic a little better. I appreciate you raising this!!

Issue #1464, opened 2024-08-09
Proposal - A better way of managing errors through the GoodJob UI


lucasfcunha:
Hi,

We recently started migrating our background jobs from Resque to GoodJob. We are very happy with the change overall, the only thing lacking a bit is the UI for managing errors, it's hard to get a good overall picture of what failed, which error happened, and when.

I would like to know if @bensheldon would be willing to accept a contribution to the GoodJob UI to create a "cleaner" tab on the UI that is similar to what Resque had.

![SCR-20240809-kjhn](https://github.com/user-attachments/assets/1ae06ff6-aada-4c24-b4cb-3effa9cbb064)

The idea would be to create a new tab beside the "Performance" tab at the top called "Cleaner" or something similar that would house a UI that resembles the one from Resque above. A table indicating which queues had errors and separate when those happened with some intervals as well as the same view based on the type of error that happened (exception fired).

To achieve this we would have also to create a new column on the table `good_jobs` called `error_class` to filter the jobs that failed with a specific exception. This could become a specific filter on the discard page as well.

The links on the table would drive the user to the "discarded" tab with that specific timeframe/error/class filtered, so they could just retry the jobs or destroy it if necessary.

I believe this change would have a good impact on people who run a lot of different jobs in a production environment and need to have a better visualization of errors happening.

@bensheldon let me know what you think about this, if you think that this is positive and you would be willing to merge it, we would be happy to contribute on this.

bensheldon:
I'm ok with this generally. I think it could simply be called "Errors" in the top navbar.

I think the design questions that come to mind are:

- Generally, how does one use this view? I think seeing a list of errors, and a relative frequency/pain/impact is helpful, but like what's the human workflow one uses this for? Just curious how you use this page with Resque (I'm guessing "Cleaner" implies a usage I'm not familar with). For context, one of my goals for GoodJob is to reduce the effort of using/maintaining a background job system, so I'm wondering (generally) if there are specific ops/anxiety-led question(s) that would lead to someone checking this view, that we could more directly respond to than providing a table of values to eyeball (not a blocking musing).
- What in the job lifecycle should this show? Should an error only show up here once the job has been discarded, or should this also be errors for each and every execution?
- Could the end-result be achieved instead by added an "error class" dropdown filter on the main jobs page instead, and that would utilize the chart to display similar data.

[...]

PR #1538, opened 2024-11-07
Add "Discard Cleaner" page to dashboard UI

lucasfcunha:
Hi,

This PR closes this issue https://github.com/bensheldon/good_job/issues/1464

Here is the final version of the page:
image

This is to help people manage the discarded jobs and take action on specific failed jobs or specific exceptions.

At the moment, no modifications to the database have been made, but I believe it would be interesting in the future to have the `error_class` and `error_message` in separate columns in the `GoodJob::Job` table. This way we could add indexes to make some filters more performant and even add a new select to the `discarded` table to filter by `error_class` (just like we have for queue and job).

PS: I opened a discussion question about the i18n case, at this moment I added the keys to the other languages with the same value as the English one, let me know what's the best way to handle this.

Thanks!


lucasfcunha:
@bensheldon I saw your comment here:

https://github.com/bensheldon/good_job/discussions/1537

Do you want me to delete the other language files from this PR or do something else? 

Let me know if there is anything that you want me to do here before merging it.

bensheldon:
@lucasfcunha I can take it from here.

Question: I renamed the tab "Discard Cleaner". Do you think "Exception Cleaner" might be more accurate?

Issue #1540, opened 2024-11-13
Dashboard performance issues


jgrau:
Hi @bensheldon 😄 

I've been looking into slow performance of the dashboard in production and I want to share my findings in case you didn't know about the bottlenecks that we're seeing or you can spot some low hanging fruits.

Here's a typical trace of a request to `/good_job/jobs`:

<img width="1624" alt="Screenshot 2024-11-13 at 22 33 02" src="https://github.com/user-attachments/assets/ab71df27-a6aa-467e-98bb-67baf9853e82">

It's about a 20s load time where all the time is spent on database queries. The most expensive is

```sql
SELECT "good_jobs".*, "pg_locks".locktype, "pg_locks"."pid" = pg_backend_pid() AS owns_advisory_lock FROM "good_jobs" LEFT JOIN pg_locks ON pg_locks.locktype = $1 AND pg_locks.objsubid = $2 AND pg_locks.classid = ($3 || substr(md5($4 || $5 || "good_jobs"."id"::text), $6, $7))::bit(32)::int AND pg_locks.objid = (($8 || substr(md5($9 || $10 || "good_jobs"."id"::text), $11, $12))::bit(64) << $13)::bit(32)::int ORDER BY COALESCE(scheduled_at, created_at) DESC, id DESC LIMIT $14
```

An `explain` of that yields this result:

```
Limit  (cost=414045.20..414045.27 rows=25 width=1016)
  ->  Sort  (cost=414045.20..419319.12 rows=2109565 width=1016)
        Sort Key: (COALESCE(good_jobs.scheduled_at, good_jobs.created_at)) DESC, good_jobs.id DESC
        ->  Hash Left Join  (cost=15.02..354514.80 rows=2109565 width=1016)
              Hash Cond: (((((('x'::text || substr(md5(('good_jobs-'::text || (good_jobs.id)::text)), 1, 16)))::bit(32))::integer)::oid = l.classid) AND ((((((('x'::text || substr(md5(('good_jobs-'::text || (good_jobs.id)::text)), 1, 16)))::bit(64) << 32))::bit(32))::integer)::oid = l.objid))
              ->  Seq Scan on good_jobs  (cost=0.00..270116.65 rows=2109565 width=975)
              ->  Hash  (cost=15.00..15.00 rows=1 width=44)
                    ->  Function Scan on pg_lock_status l  (cost=0.00..15.00 rows=1 width=44)
[...]

PR #1541, opened 2024-11-14
Remove unneeded include of pg_locks in query when displaying jobs table

jgrau:
Related to https://github.com/bensheldon/good_job/issues/1540

* Removes an (unused?) join of the pg_locks table which was slow

@bensheldon regarding your comments

> * The pg_locks could be replaced with a locked_at IS NOT NULL query

I'm not sure but I found that the joined table was not being used for constraints but for including the `locktype` which seems to be unused(?). I removed it and browsed around the dashboard and found to errors 🤷 

> * The COALESCE can be replaced solely with good_jobs.scheduled_at because now all jobs should have a scheduled_at regardless of whether they are expect to run immediately or in the future.

Can you elaborate? There's multiple places where we use `coalesce(scheduled_at, created_at)` so if that's not needed in general anymore I feel there's a lot of cleanup that could be done and I can do that in this PR with your guidance :)

### DIFF
--- a/app/filters/good_job/jobs_filter.rb
+++ b/app/filters/good_job/jobs_filter.rb
@@ -39,7 +39,7 @@ module GoodJob
         when 'scheduled'
           query = query.scheduled
         when 'running'
-          query = query.running.select("#{GoodJob::Job.table_name}.*", 'pg_locks.locktype')
+          query = query.running
         when 'queued'
           query = query.queued
         end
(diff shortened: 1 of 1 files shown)


bensheldon:
Maybe let's do the COALESCE removal in a separate PR so I can land this one sooner. I want to poke at it a little, but this looks good 👍

PR #1550, opened 2024-11-25
GoodJob.cleanup_preserved_jobs: add :include_discarded option

jonleighton:
This change allows the caller of `GoodJob.cleanup_preserved_jobs` to specify the `:include_discarded` option, rather than always relying on the value that is set in the configuration.

This would enable us to e.g. set `config.cleanup_preserved_jobs_before_seconds_ago` to a very low value (e.g. 10 seconds), but also set `config.cleanup_discarded_jobs = false` so that we can retain discarded jobs for longer in case we want to inspect them or manually retry then.

We would then set up a separate maintenance job to call `GoodJob.cleanup_preserved_jobs(older_than: some_longer_period, include_discarded: true)`.

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -203,11 +203,11 @@ module GoodJob
   # If you are preserving job records this way, use this method regularly to
   # destroy old records and preserve space in your database.
   # @param older_than [nil,Numeric,ActiveSupport::Duration] Jobs older than this will be destroyed (default: +86400+).
+  # @param include_discarded [Boolean] Whether or not to destroy discarded jobs (default: per +cleanup_discarded_jobs+ config option)
   # @return [Integer] Number of job execution records and batches that were destroyed.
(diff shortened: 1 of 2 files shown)


jonleighton:
Note that it was proving difficult to get the tests running in my local dev environment, so I hoped to push this up and run it on your CI.

jonleighton:
I realised that this project is setup for devcontainers and managed to run and fix the tests locally, so hopefully CI will pass now.

jonleighton:
Looks like the `rails_head` one failed CI, but that looks unrelated to my change.

bensheldon:
This looks good 👍🏻 

I have this vague memory of not wanting to pass through the value as an argument because I was worried that someone might manually delete jobs and cause operational confusion... but I think your reasoning for why it would be useful is compelling.

jonleighton:
Thank you!

PR #1553, opened 2024-11-26
Fix compatibility with `rails-head` when duplicated advisory lockable column

Earlopain:
The primary key and column may be identical. Current rails head will select this column twice, which confuses postgres.

Draft for now, let's first see how https://github.com/rails/rails/issues/53747 goes.

### DIFF
--- a/spec/app/models/concerns/good_job/advisory_lockable_spec.rb
+++ b/spec/app/models/concerns/good_job/advisory_lockable_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe GoodJob::AdvisoryLockable do
           FROM "good_jobs"
           WHERE "good_jobs"."id" IN (
             WITH "rows" AS #{'MATERIALIZED' if model_class.supports_cte_materialization_specifiers?} (
-              SELECT "good_jobs"."id", "good_jobs"."id"
+              SELECT "good_jobs"."id"
               FROM "good_jobs"
               WHERE "good_jobs"."priority" = 99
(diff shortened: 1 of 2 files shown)


spec/app/models/concerns/good_job/advisory_lockable_spec.rb
```diff
@@ -51,7 +51,7 @@
           FROM "good_jobs"
           WHERE "good_jobs"."id" IN (
             WITH "rows" AS #{'MATERIALIZED' if model_class.supports_cte_materialization_specifiers?} (
-              SELECT "good_jobs"."id", "good_jobs"."id"
```
Earlopain, line 54: Seems this expectation was wrong from the start. I don't think this would execute, same as in the issue

bensheldon:
This looks like a great fix! Simple and definitely _better_ now 🙇🏻

bensheldon:
I think we may as well fix this in GoodJob. I'd like to make one tweak to see if I can avoid the array/uniq calculation because I think it would be marginally more efficient to do an `==` check`. I can land this 🫡

bensheldon:
> You may also want to drop the to_sym conversion.

I eventually plan to extract `advisory_lockable.rb` to a separate gem (I'm already using it in a few projects) so I want to be defensive about its inputs. I also wish Active Record was clear about when it wants symbols and when it wants strings; I frequently see a symbol input documented but when I go read the code it's coercing it to a string internally 🤷🏻

PR #1559, opened 2024-12-03
Reintroduce JRuby 9.4 to testing matrix

bensheldon:
Removed in #1391


Earlopain:
Made some tweak since it would definitly fail the way you had it beforehand. JRuby only supports Ruby 3.1 syntax so anything after Rails 7.2 would 100% not work. Up to you if you want to run different Rails versions for JRuby, I have it only on 7.2 now. I think that is fine? But maybe you think otherwise.

PR #1561, opened 2024-12-12
Expose good job labels in dashboard

BClark88:
Does the following:

* Exposes labels in the dashboard

image


app/views/good_job/shared/_filter.erb
```diff
@@ -71,5 +82,29 @@
         });
       })
     })
+
+    document.addEventListener("DOMContentLoaded", () => {
```
BClark88, line 86: I don't write a lot of javascript, so I wouldn't be surprised if there's a better way to do this

app/views/good_job/shared/_filter.erb
```diff
@@ -31,12 +31,23 @@
...
+          </select>
+        </div>
+
         <div class="me-md-2 mb-2 mb-md-0 flex-fill">
           <%= label_tag "query", t(".search"), class: "visually-hidden" %>
           <%= search_field_tag "query", params[:query], class: "form-control form-control-sm", placeholder: t(".placeholder"), autocomplete: "off" %>
         </div>
 
-        <div class="d-flex flex-col align-items-end">
+        <div class="d-flex flex-col align-items-start">
```
BClark88, line 50: I did this so that the `Search` and `Clear` buttons were positioned nicely. Otherwise it would look like this:
image

PR #1564, opened 2024-12-18
Remove `rubygems: latest` from CI workflows because latest Rubygems is not compatible with Ruby 3.0

bensheldon:


Earlopain:
I think this is not needed anymore because of https://github.com/ruby/setup-ruby/pull/676

PR #1565, opened 2024-12-27
Add Ruby 3.4 to CI

Earlopain:
'twas ruby release day :christmas_tree:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,11 +92,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


Earlopain:
CI is :green_circle:, only two tests failed because Ruby changed how hashes are printed. 7.0 now uses 7-0-stable because there is no released version that declares bigdecimal as a dependency.

`gem-releaser` complains about `ostruct` but I can't seem to silence those warnings :shrug: https://github.com/svenfuchs/gem-release/pull/112

bensheldon:
Thank you!!

PR #1568, opened 2025-01-02
Allow filtering by label

BClark88:


app/views/good_job/shared/_filter.erb
```diff
@@ -71,5 +82,26 @@
         });
       })
     })
+
+    const labelFilter = document.getElementById('job_label_filter');
+    labelFilter.addEventListener('change', () => {
+      const selectedValues = Array.from(labelFilter.selectedOptions).map(option => option.value);
+      const allLabelsOption = labelFilter.querySelector('option[value="_all"]');
+
+      if (selectedValues.includes('_all')) {
```
BClark88, line 91: There's probably a better way to do this; I don't write a lot of javascript

app/filters/good_job/base_filter.rb
```diff
@@ -31,6 +31,15 @@ def queues
                 .to_h
     end
 
+    def labels
+      filtered_query(params.slice(:queue_name))
+        .joins("CROSS JOIN unnest(labels) AS label")
+        .group("label")
+        .order("label")
+        .count
+        .to_h
```
BClark88, line 40: As you mentioned, this probably needs to be moved into something that can be done asynchronously

Issue #1574, opened 2025-01-17
Ability to pause queues


salmonsteak1:
Hey there, is the functionality of pausing queues on the roadmap anytime soon? (Something like what Solid Queue has, where we can pause individual job queues). Thank you!

bensheldon:
Just got this released 🎉 https://github.com/bensheldon/good_job/releases/tag/v4.8.0

Issue #919, opened 2023-04-06
Pause or cancel a batch?


ajselvig:
I'm using the new batches functionality and it's working well. However, I haven't been able to find a way to pause or cancel the execution of a particular batch. I tried setting the `finished_at` value but that didn't have any effect. 

I could do a bulk update on all jobs in the batch, but even then it's not clear what I should be setting. Any guidance would be appreciated. 

Thanks!

bensheldon:
Great feature request. Something like that doesn't currently exist.

Currently, it's possible to manually Discard each of the jobs within a batch, but that would still result in the Batch Callback(s) being triggered.

There aren't any manual Batch operations right now that are analogous to a Job being able to be manually retried/rescheduled/discarded.

Thinking aloud here on designing something new:
- A batch could be "Discarded" which would discard all of the jobs that have not yet begun executing. Manually discarding the batch would not trigger any batch callbacks.
- Discarding a batch would Discard all of the jobs that haven't yet begun executing.
- Un-discarding a batch would Retry all of the jobs in the batch that had been Discarded.
- Would you want to track Manually Discarded batches on their own? I'm thinking we'd want to introduce a new column like `batch_discarded_at` to differentiate between Batches that are "Discarded" because a job _within the batch_ was discarded (and thus would trigger callbacks), and Batches that are Manually Discarded as a whole (and thus not trigger callbacks). Maybe that deserves a new term like "Cancel"?

bensheldon:
...and now that I'm looking through the manual job actions `GoodJob::Job#discard_job` and `GoodJob::Job#retry_job` I'm noticing that they aren't batch aware and don't trigger batch callbacks. That could be a bug, or just YAGNI until someone says they want that. I'm imagining that they _should_ trigger callbacks though.

ajselvig:
Wow, thanks for the extremely quick and thorough response!

> Would you want to track Manually Discarded batches on their own?

Definitely seems like a different concept. For our use case, at least, the idea that a batch gets "discarded" just because one of it's jobs is discarded isn't super meaningful. If we queue up 10,000 invoices to be sent and 20 of them fail for some reason, the batch itself was still a success. That's fundamentally different than a user explicitly cancelling the batch. 

> If you simply want a workaround, you could discard all of the jobs this way:

I'm fine with workarounds but this seems quite heavyweight for a large number of jobs. Seems like this would work fine as well?

```ruby
[...]

PR #1575, opened 2025-01-18
Add ability to pause jobs by queue, job class, or label (experimental)

bensheldon:
### Technical Description

Paused values are stored in the GoodJob::Setting table. During the fetch-and-lock query, this record is loaded as sub=selects that add additional scopes to the query that exclude them. Paused jobs will appear as "queued" (a visible "paused" status in the Dashboard is tbd).

I'm unsure of the performance impact of the query scopes, so I've made this feature "opt-in". 

Query right now looks like this:

```sql
WHERE "good_jobs"."queue_name" NOT IN (SELECT jsonb_array_elements_text(value->'queues') FROM "good_job_settings" WHERE "good_job_settings"."key" = 'pauses') 
  AND "good_jobs"."job_class" NOT IN (SELECT jsonb_array_elements_text(value->'job_classes') FROM "good_job_settings" WHERE "good_job_settings"."key" = 'pauses') 
  AND NOT ("good_jobs"."labels" && ARRAY((SELECT jsonb_array_elements_text(value->'labels') FROM "good_job_settings" WHERE "good_job_settings"."key" = 'pauses')))
```
---

Connects to #919, #1574 

I put the pause/unpause buttons on the Performance page. I think that could be problematic, though I'm not sure if it will be more or less so than how it affects the fetch-and-lock query to add new conditions. 

Future thoughts:

- [x] Should probably also allow pausing based on label
- [x] I think Pausing may deserve it's own page. The Performance page already has performance problems (the irony!) so this should probably simply be a "type in the thing you want to pause" 
- [ ] The views should be updated to show a "Paused" status if the individual job matches the conditions. Paused jobs will still show up on "Queued" page.


app/models/good_job/job.rb
```diff
@@ -96,6 +96,21 @@ class Job < BaseRecord
...
+    # @return [ActiveRecord::Relation]
+    scope :exclude_paused, lambda {
+      return all unless GoodJob.configuration.enable_pause
+
+      paused_queues_query = GoodJob::Setting.where(key: GoodJob::Setting::PAUSED_QUEUES)
+                                            .select('jsonb_array_elements_text(value)')
+      paused_job_classes_query = GoodJob::Setting.where(key: GoodJob::Setting::PAUSED_JOB_CLASSES)
+                                                 .select('jsonb_array_elements_text(value)')
+      where.not(queue_name: paused_queues_query).where.not(job_class: paused_job_classes_query)
+    }
```
bensheldon, line 112: This is a new scope in the hot path of job fetch-and-lock.

jnunemaker:
Exciting!

bensheldon:
It's happening!

app/models/good_job/job.rb
```diff
@@ -96,6 +96,33 @@ class Job < BaseRecord
...
+               Arel::Nodes::NamedFunction.new(
+                 "COALESCE", [
+                   Arel::Nodes::InfixOperation.new('&&', arel_table['labels'], Arel::Nodes::NamedFunction.new('ARRAY', [paused_labels_query.arel])),
+                   Arel::Nodes::SqlLiteral.new('FALSE'),
+                 ]
+               )
+             )
+           )
+    }
+
```
bensheldon, line 125: THis is the hot query, now with labels, which makes it slightly more complicated.

README.md
```diff
@@ -329,6 +330,8 @@ Available configuration options are:
     config.good_job.probe_handler = 'webrick'
     ```
 
+- `enable_pause` (boolean) whether job processing can be paused. Defaults to `false`. You can also set this with the environment variable `GOOD_JOB_ENABLE_PAUSE`.
```
bensheldon, line 333: Update

app/controllers/good_job/performance_controller.rb
```diff
@@ -3,17 +3,25 @@
...
+
+      @queue_performances = GoodJob::Execution.group(:queue_name).select("
+        queue_name,
+        COUNT(*) AS executions_count,
+        AVG(duration) AS avg_duration,
+        MIN(duration) AS min_duration,
+        MAX(duration) AS max_duration
+      ").order(:queue_name)
+
+      @paused = GoodJob::Setting.paused
```
bensheldon, line 22: Remove

config/routes.rb
```diff
@@ -34,7 +34,19 @@
   end
 
   resources :processes, only: %i[index]
-  resources :performance, only: %i[index show]
+  resources :performance, only: %i[index show] do
+    collection do
+      get :pause, to: redirect(path: 'performance')
+      post :pause
+      delete :unpause
```
bensheldon, line 41: Remove

lib/good_job/configuration.rb
```diff
@@ -369,6 +371,16 @@ def enqueue_after_transaction_commit
       DEFAULT_ENQUEUE_AFTER_TRANSACTION_COMMIT
     end
 
+    # Whether the job processing can be paused.
+    # @return [Boolean]
+    def enable_pauses
+      return options[:enable_pauses] unless options[:enable_pauses].nil?
+      return rails_config[:enable_pauses] unless rails_config[:enable_pauses].nil?
+      return ActiveModel::Type::Boolean.new.cast(env['GOOD_JOB_ENABLE_PAUSE']) unless env['GOOD_JOB_ENABLE_PAUSE'].nil?
```
bensheldon, line 379: Pluralize

PR #1586, opened 2025-01-26
[WIP] Add row-level "FOR (NO KEY) UPDATE SKIP LOCKED" option 

bensheldon:


spec/app/models/concerns/good_job/row_lockable_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  around do |example|
+    RSpec.configure do |config|
+      config.expect_with :rspec do |c|
+        original_max_formatted_output_length = c.instance_variable_get(:@max_formatted_output_length)
+
+        c.max_formatted_output_length = 1000000
+        example.run
+
+        c.max_formatted_output_length = original_max_formatted_output_length
```
francois, line 27: Don't we need to `ensure` so that the value is reset?

PR #1587, opened 2025-01-26
Fix passing a block to `advisory_lock` in tests

Earlopain:
This method does not take a block. As such, the expectation is never executed


bensheldon:
Thanks for catching that!

PR #1588, opened 2025-01-26
Ensure that executions are properly ordered in the dashboard

Earlopain:
They are not ordered, so postgres may return them in any order. I encounter this on postgres 17, checked the demo app to link to but didn't find any. For reference, locally:

image


bensheldon:
Nice! I don't think I've upgraded the Demo app yet to PG 17 😓

PR #1589, opened 2025-01-26
Sync demo schema

Earlopain:
After https://github.com/rails/rails/pull/52313, the schema name is now included


Earlopain:
Huh, looks like this doesn't play well with older rails versions, I guess they don't support extension schemas to begin wtih. Not that important, just a bit annoying having an untracked change

Issue #1591, opened 2025-01-30
Use custom ActiveRecord connection parameters in GoodJob external process


rdallasgray:
Hi there, migrating from Sidekiq, where we would use the `configure_server` block to set a longer `statement_timeout` for all ActiveRecord queries inside the Sidekiq server process (where, in the 'real time' Rails server, we would use a timeout commensurate with the timeout for requests). Is there a way to do something equivalent in GoodJob? It seems like `configure_active_record` would only apply to models inheriting from `GoodJob::BaseRecord`, not to all connections.

sam-ejma081:
Hello, we tried to solve the issue.

This is what we did:

Add a new configure_server method to the GoodJob module to allow setting custom configurations for the GoodJob server process, including the ability to set a longer statement_timeout for ActiveRecord queries.

You can review changes in this commit: https://github.com/sam-ejma081/bensheldon-good_job-1591/commit/f5c69f9604386adf794944e301193b354786b3dd.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

bensheldon:
If these are dynamic values, I think a simple `around_perform` in your application's `ApplicationJob` would be a good place.

I think that setting statement timeout though is done via `database.yml`, right? Any setup that happens during Rails initialization you'd need to set up directly in Rails (via an ENV variable probably). 

Another way to do it would be by having multiple named database configurations. Here's how I do it in one of my apps:

```
default: &default
  adapter: postgresql
  encoding: unicode
  # For details on connection pooling, see rails configuration guide
  # http://guides.rubyonrails.org/configuring.html#database-pooling
  pool: 20
  connect_timeout: 5
  checkout_timeout: 5

development:
  <<: *default
  database: dayoftheshirt_development

test:
  <<: *default
  database: dayoftheshirt_test

staging:
  primary:
    <<: *default
    database: dayoftheshirt_staging
    pool: 20
  unpooled:
    <<: *default
    url: <%= ENV["DATABASE_UNPOOLED_URL"] || ENV["DATABASE_URL"] %>
    pool: 20

production:
  primary:
    <<: *default
[...]

PR #1592, opened 2025-01-30
Add `GoodJob.cli?` to check if the current process was started by the `good_job` executable

bensheldon:
Connects to #1591


rdallasgray:
@bensheldon thanks so much for this, works perfectly for our use case. We're using the gem from main at the moment in order to try out the functionality -- do you have a timeline for the next release?

PR #1604, opened 2025-02-14
Order Dashboard jobs in more "natural" order

### LINKED ISSUE #1580: [dashboard] Scheduled tasks are shown "backwards"
It shows tasks that are scheduled in 3 hours before tasks that are scheduled in 1 minute. That makes the tab much less useful... I can't see the next upcoming tasks, only the things that are in the far future. I have to "Older Job" multiple times to get to something useful.

#### Scheduled at 22:14:19
Image

#### Scheduled at 22:10:15, but lower in the list
Image

francois:
"Next-job-first" makes more sense for the scheduled tab.

Closes #1580.

### DIFF
--- a/app/filters/good_job/base_filter.rb
+++ b/app/filters/good_job/base_filter.rb
@@ -13,13 +13,15 @@ module GoodJob
     end
 
     def records
-      after_scheduled_at = params[:after_scheduled_at].present? ? Time.zone.parse(params[:after_scheduled_at]) : nil
+      after_at = params[:after_at].present? ? Time.zone.parse(params[:after_at]) : nil
+      after_id = params[:after_id] if after_at
+      limit = params.fetch(:limit, DEFAULT_LIMIT)
 
       query_for_records.display_all(
-        state: params[:state],
-        after_scheduled_at: after_scheduled_at,
-        after_id: params[:after_id]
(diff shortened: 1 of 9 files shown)


francois:
Incidentally, I have not been able to find where the default sort order is applied? @bensheldon, feel free to move elsewhere if it makes more sense.

francois:
Hold on! This is insufficient. I hadn't tried with multiple pages.

francois:
All right, I have a solution! I had to pass `state` to the `display_all` scope. `BatchRecord` redefined the `display_all` scope so I had to introduce an unused parameter to respect the contract.

End of page 1:
Capture d’écran, le 2025-02-14 à 17 06 08

Start of page 2:
Capture d’écran, le 2025-02-14 à 17 06 14

Notice that job 01d... is present on both pages. Not sure if that is acceptable.

bensheldon:
I added ordering for all of the states. The overall principle I went with was: most recent relevant change first. e.g

- All: most recently created
- Scheduled / Retried: next scheduled job
- Queued: most recent job to become queued
- Running: most recent job to start running
- Succeeded / Discarded: most recent job to finish

I think for Queued and Running, there is an alternatively important question that's relevant which is: what is the oldest or longest running job? I will defer that until someone wants to add a feature to "reorder".

Issue #1603, opened 2025-02-13
Poor performance when enqueuing many throttled jobs because of unindexed queries


Intrepidd:
When the concurrency extension looks up for existing queued jobs within the throttle period, it performs the following query : 

https://github.com/bensheldon/good_job/blob/b557525a9003a1fc1a434742dae58053b947e708/lib/good_job/active_job_extensions/concurrency.rb#L93-L95

In my case, this query will run in 200-300ms for each enqueue attempt, so it adds up pretty quickly when trying to enqueue hundreds of jobs.

The index on `concurrency_key` has a `where finished_at is NULL`, so it can't be used here.

Would it make sense to have an index solely on `concurrency_key` ? 

Thanks

bensheldon:
Good catch on that. I think you're correct that the current index is insufficient for the throttle query 😓

I think the correct index here would be `[concurrency_key, :created_at]`. That would also cover the execution throttling (as it can use  justthe first column of the compound index):

https://github.com/bensheldon/good_job/blob/b557525a9003a1fc1a434742dae58053b947e708/lib/good_job/active_job_extensions/concurrency.rb#L162C1-L164C165

Intrepidd:
Thanks ! I will make a PR shortly if that's ok

Intrepidd:
Just opened #1605 to try and fix this

Intrepidd:
Closing as it's been merged, thanks !

Intrepidd:
Just wanted to report that it worked very well, thanks again !

![Image](https://github.com/user-attachments/assets/86cfba28-31d5-48ee-b17f-9edf06ac5487)

PR #1605, opened 2025-02-16
Add index on good_jobs: [:concurrency_key, :created_at] to improve performance of throttling (#1603)

Intrepidd:
As discussed in #1603 

Did I miss anything ?

### DIFF
--- a/lib/generators/good_job/templates/update/migrations/03_add_index_good_jobs_concurrency_key_created_at.rb.erb
+++ b/lib/generators/good_job/templates/update/migrations/03_add_index_good_jobs_concurrency_key_created_at.rb.erb
@@ -4,6 +4,14 @@ class AddIndexGoodJobsConcurrencyKeyCreatedAt < ActiveRecord::Migration<%= migra
   disable_ddl_transaction!
 
   def change
+    reversible do |dir|
+      dir.up do
+        # Ensure this incremental update migration is idempotent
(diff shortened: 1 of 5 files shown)


Intrepidd:
Looks like sorbet doesn't like the rails 8.0 migration, I'm not familiar with it,  how should I go about fixing it ?

bensheldon:
Thank you! 🙇🏻 I can fix this up and get it merged 👍🏻

lib/generators/good_job/templates/update/migrations/03_add_index_good_jobs_concurrency_key_created_at.rb.erb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIndexGoodJobsConcurrencyKeyCreatedAt < ActiveRecord::Migration<%= migration_version %>
+  def change
+    add_index :good_jobs, [:concurrency_key, :created_at]
```
bensheldon, line 5: Always make sure this is done concurrently

Intrepidd: Sure thing, done in 8b69fd99fa388b7e50967ab6d735cd6a7762ab39

PR #1610, opened 2025-02-21
Update `form_with` calls to be compatible with Rails 8

sallyhall:
[Rails 8 removed deprecated support to passing `nil` to the `model:` argument of `form_with`.](https://github.com/rails/rails/blob/8-0-stable/actionview/CHANGELOG.md#rails-800rc1-october-19-2024)

To make the good job dashboard views compatible with Rails 8, this adds `model: false` to forms in both  `app/views/good_job/jobs/_table.erb` and `app/views/good_job/shared/_filter.erb`.

Without these changes, loading the dashboard in a Rails 8 application causes `ArgumentError (Passed nil to the :model argument, expect an object or false)`.

### DIFF
--- a/app/views/good_job/jobs/_table.erb
+++ b/app/views/good_job/jobs/_table.erb
@@ -1,4 +1,4 @@
-<%= form_with(url: mass_update_jobs_path(filter.to_params), method: :put, local: true, data: { "checkbox-toggle": "job_ids" }) do |form| %>
+<%= form_with(model: false, url: mass_update_jobs_path(filter.to_params), method: :put, local: true, data: { "checkbox-toggle": "job_ids" }) do |form| %>
   <div class="my-3 card" data-gj-poll-replace id="jobs-table">
     <div class="list-group list-group-flush text-nowrap table-jobs" role="table">
       <header class="list-group-item bg-body-tertiary">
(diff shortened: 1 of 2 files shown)


bensheldon:
@sallyhall thank you! No objection to this for explicitness, but I wonder if your application has monkeypatched `form_with`. I suggest this because Rail's param default is `model: false` already:

https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/actionview/lib/action_view/helpers/form_helper.rb#L756

sallyhall:
> @sallyhall thank you! No objection to this for explicitness, but I wonder if your application has monkeypatched `form_with`. I suggest this because Rail's param default is `model: false` already:
> 
> https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/actionview/lib/action_view/helpers/form_helper.rb#L756

@bensheldon Thank you for pointing that out! It turns out that several gems dependencies down, we are using https://github.com/amatsuda/html5_validators, which explicitly passes `model: nil` to `form_with`. I'll close this PR and make a pr there instead.

bensheldon:
No worries. I imagine others might run into this too, so I'll merge your change. Can never be too defensive 😇

PR #1626, opened 2025-04-03
Fix enqueue_after_transaction_commit config

santib:
Hey Ben 👋 first of all, thanks for this great gem, we chose it as the default for our base app https://github.com/rootstrap/rails_api_base.

I was trying to configure `enqueue_after_transaction_commit = true` by default so that we don't get different behaviors when using the same DB for jobs and the app, or multiple DBs.

I did some testing but couldn't make the config work, I guess it's because I added like:
```ruby

```
  # Don't depend on transactions
  config.good_job.enqueue_after_transaction_commit = true
```

and that lives in `rails_config` not in `options`. Is that right or did I miss something?


bensheldon:
Whoops! Thank you for catching that!

bensheldon (approved):

PR #1628, opened 2025-04-16
Fix Batches dashboard deserealization error

JulianPasquale:
Hey @bensheldon, amazing work with this gem!! I've been using it in a prod app for some time and it works really well.

I found an issue with the Batch jobs dashboard.
image

When a batch record is created and an ActiveRecord object is passed as a property, an error is raised in the batches UI if the record cannot be deserialized.  Following the example in the README file:
```ruby
GoodJob::Batch.enqueue(on_finish: MyBatchCallbackJob, user: current_user) do
  ...
end
```
If the user that is being passed as an argument here is later removed from the database, the batches UI will start to break.

I understand capturing the DeserializationError exception is not a good approach because this "UI issue" should not impact the job tool itself: If I run `batch.properties[:user]` and a hash is returned instead of an ActiveRecord instance, it doesn't sound right.
Also know there is a comment about potencial deserialization issues on the [README](https://github.com/bensheldon/good_job?tab=readme-ov-file#other-batch-details), but I still think we should fix the UI somehow. 

I know the data is not being rendered in a "pretty" way anymore, so let me know if you think there is a better approach and I can adjust the PR accordingly 🙂. Also wondering if we should add a test for this scenario somewhere.


app/views/good_job/batches/_table.erb
```diff
@@ -76,7 +76,7 @@
             </div>
           </div>
           <%= tag.div id: dom_id(batch, "properties"), class: "batch-properties list-group-item collapse small bg-dark text-light" do %>
-            <%= tag.pre JSON.pretty_generate(batch.properties) %>
+            <%= tag.pre JSON.pretty_generate(JSON.parse(batch.read_attribute_before_type_cast(:serialized_properties))) %>
```
bensheldon, line 79: Could this more simply be `serialized_properties` directly because it will be typecast as JSON?

JulianPasquale: hmm I think it doesn't work, the issue happens when the model is deserialized, so most of the methods actually raise an error.

Reproduction steps:
```ruby
record = GoodJob::Job.create # Or any ActiveRecord instance
batch = GoodJob::Batch.enqueue(record: record)
record.destroy!

batch.reload => # An error occurred when inspecting the object: #<ActiveJob::DeserializationError: Error while trying to deserialize arguments: Couldn't find GoodJob::Job with 'id'=20ea1655-c5ca-4cbc-988e-aae7a6b1dad8>

batch._record.id # => "some-uuid"
batch._record.serialized_properties # Error while trying to deserialize arguments: Couldn't find GoodJob::Job with 'id'=20ea1655-c5ca-4cbc-988e-aae7a6b1dad8 (ActiveJob::DeserializationError)
```

JulianPasquale: My initial approach was to actually catch the deserialization error (see [first commit](https://github.com/bensheldon/good_job/pull/1628/commits/60c41f6ddab63972df5e027a5eedae3440c9e0c0)). As I mentioned in the description, I don't think it's the right way. It fixes the batch record though

bensheldon: Can this mirror how Job's serialized_parameters are handled?

JulianPasquale: I added a new `display_serialized_properties` method that handles the deserialization errors, and also updated the `display_attributes` with the same logic.

The Job model always keeps the `serialized_params` data unserialized, and it's actually serialized only when the instance is casted into an active_job. If that's what you meant, we should get ride of the serialization logic for `serialized_properties` and only keep it for the `properties` methods since that is the "public" interface. Can give it a try if you agree with that.

bensheldon (approved):

bensheldon:
Thanks for working on this! I made a few tweaks and I'll merge momentarily.

Looking at this now, I do think that I shouldn't use Active Record `serialize` and instead manage that directly. Of what I remember, I think I used `serialize` because I was struggling with dirty tracking. That should be revisited.

PR #1637, opened 2025-05-16
Do not heartbeat when transactional tests / shared connections are in use

bensheldon:
This is somewhat speculative. I'm experiencing a deadlock in:

- A test
- That's using transactional fixtures
- And GoodJob is running inline with `GoodJob.perform_inline_jobs`


bensheldon:
Oh! It's not _actually_ the heartbeat that's causing this issue. It's the Turbo debouncer which enqueued a job in a background thread and that's causing AR to be confused. Wow.

PR #1638, opened 2025-05-21
Fix the path for the extended dashboard views in README.md

zarembas:
I think the correct path does not include the `jobs` folder.


bensheldon:
Thanks for catching that!

PR #1639, opened 2025-05-21
Add customisable job index details partial

zarembas:
Adds a new extension point to the GoodJob dashboard that allows applications to customise the job index page. The new `_custom_job_index_details.html.erb` partial is rendered in each job row providing the same customisation capabilities as the existing job and execution details partials.

An extension of https://github.com/bensheldon/good_job/pull/1200


zarembas:
Thanks for your review. I played around with it and ended up adding it at the bottom as a row. It works well for my specific use case, as there was limited space in the first column I used initially. What do you think?

zarembas:
I'm not sure adding more partials is the right approach. I'll close this for now.

Issue #1632, opened 2025-04-24
The probe server doesn't handle SIGINT/SIGTERM


TAGraves:
If I'm running with `--probe-port`, sending SIGINT to the good_job process logs this to stderr on MacOS:
```
Server encountered an error: Bad file descriptor
```

On Linux in our production deployments, it logs:
```
Server encountered an error: closed stream
```

This isn't particularly problematic for us, but it is noisy in our error logs since we see it any time we deploy. The probe server should gracefully handle termination signals.

bensheldon:
That's weird! And not good. I bet it's somewhere in here:

https://github.com/bensheldon/good_job/blob/49befb9831d9c809ba80fb0d84818cc1b7e08041/lib/good_job/probe_server/simple_handler.rb#L48C11-L48C29

I'll see if I can reproduce and track down.

bensheldon:
Fixed by #1644, I hope.

PR #1644, opened 2025-05-29
Probe Server rescues additional IO errors

bensheldon:
Connects to #1632.


bensheldon:
I think this solution is rather sus because these exceptions are not actually retryable. Some analysis:

### 1. `IO::WaitReadable`
**Truly retryable**: Yes. This is specifically designed for non-blocking I/O operations and indicates that an operation would block because no data is available yet. Retrying after some time (which happens naturally after the `IO.select` call) is the correct approach. 

### 2. `Errno::EINTR`
**Truly retryable**: Yes. This indicates a system call was interrupted by a signal. It's a classic example of a transient error that should be retried. The interrupted system call is expected to succeed if attempted again.

### 3. `Errno::EPIPE`
**Not always retryable**: This indicates a broken pipe - typically when the remote end has closed the connection. Retrying the _same_ operation on the _same_ socket will not succeed. However, in this server loop context, retrying means going back to the beginning of the loop to handle a new connection, which is reasonable. But it would be clearer to handle this separately rather than retrying immediately.

### 4. `Errno::EBADF`
**Rarely retryable**: This indicates a bad file descriptor - the socket or file handle is invalid. This is usually a permanent error that won't be fixed by retrying. If `@server` becomes invalid, retrying won't help. This might indicate a deeper issue like another thread closing the server socket. 

### 5. `IOError`
**Generally not retryable**: This is a broad category of I/O errors, many of which are permanent. For instance, operations on closed I/O objects will consistently fail.

PR #1647, opened 2025-06-03
Make the processes page more responsive by extracting queues and max_threads

luizkowalski:
Hey 👋🏻 

I noticed that on my Macbook 14, the processes page looks off

image

I played around a bit with the view (though I'm not really versed in Bootstrap) and managed to get to this (this is around 1300px but looks like this, even under 1000px)

image

because I truncated the schedulers, I had to add `title` to it so we can see what's going on when we leave the cursor on top of it


bensheldon:
@luizkowalski thanks for proposing this. Any chance you could improve it by just showing what's inside the `()`-parenthesis. The class name really isn't necessary, it's the configuration that's meaningful.

luizkowalski:
hey ben! good idea. I updated to a regex to capture what's inside `()`

Issue #1571, opened 2025-01-09
[GoodJob] Notifier errored: NameError: uninitialized constant GoodJob::Job in develpment


tdegrunt:
Thanks for such a great tool, really loving the gem!

I'm seeing the following in development mode, these errors are intermittent. 
After restart of good_job they disappear, but they always come back after some time.
Not sure what's the cause of this, do you have any idea? Could it be the rails reloader?

This is what I'm seeing in my (overmind) logs:

```
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/good_job-4.4.2/lib/good_job/notifier.rb:257:in `block in with_connection'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/activesupport-7.1.5.1/lib/active_support/execution_wrapper.rb:92:in `wrap'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/good_job-4.4.2/lib/good_job/notifier.rb:256:in `with_connection'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/good_job-4.4.2/lib/good_job/notifier.rb:204:in `block in create_listen_task'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/concurrent-ruby-1.3.4/lib/concurrent-ruby/concurrent/executor/safe_task_executor.rb:24:in `block in execute'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/concurrent-ruby-1.3.4/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `block in synchronize'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/concurrent-ruby-1.3.4/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `synchronize'", 
"/Users/tdegrunt/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/concurrent-ruby-1.3.4/lib/concurrent-ruby/concurrent/synchronization/mutex_lockable_object.rb:48:in `synchronize'", 
[...]

PR #1653, opened 2025-06-30
Add `require rails` to good_job/engine.rb to allow `bundle console` usage

bensheldon:
Problem noted in https://github.com/bensheldon/good_job/issues/1571#issuecomment-3018612173

This seems like something other gems/engines have experienced: https://github.com/pat/combustion/issues/27

My fear is that this will make GoodJob require `gem "rails"` rather than allow for enumerating a subset of gems, but it does seem like Rails packages do this too... though maybe they don't know why either 🤷🏻 https://github.com/rails/rails/commit/a1a363a8aa0f0717ef329eba4a5fe57705fb6b2f


Earlopain:
The `railties` gem actually ships with the file for `require "rails"` (`rails` is just a metapackage) so you don't depend on all the individual subgems this way, but you do expect `railties` to be present. `solid_queue` does declare its dependency on `railties` which seems more correct but in practise it doesn't seem to make much a difference, since you can't have a functioning rails app without it.

As for why rails does it this way: It sets up the autloads properly. If you just require "rails/engine", all sort of other things you expect to be avaialble won't be.

Issue #1065, opened 2023-09-02
Allow priority to be globally disabled/ignored


bensheldon:
It would be nice if GoodJob could be configured to disable/ignore priority as a job feature. e.g. `config.good_job.enable_active_job_priority = false`. Why?

- For performance reasons: it's not possible to create an efficient database index for time-based (e.g. scheduled and FIFO) _and_ ordered by priority. That requires either:
  - Using an index ordered on `priority, created_at` and then doing a condition check on scheduled_at (this is what I find most query planners choose)
  - Using an index ordered on `scheduled_at` that then does a reorder on priority
- I believe priority only makes sense when either queues are not well-designed, or there is insufficient capacity. Either way, there is a better action to take than introducing priority (redesigning queues based on SLOs; adding capacity) when the primary success metric is reducing queuing latency (as I believe it should be)
- GoodJob provides ample tools for executing different named queues (separately, concurrently, combined). 

I do observe though that developers like to name queues based on the constrained resource (e.g. `email_backend`) and then use priority to hint at the SLO. I imagine people will want to preserve that organization style, so I'll probably need to introduce something like job "labels" or "tags" (which maybe connects to #753)

Wittiest:
Hey @bensheldon,

We're looking to boost our performance with good_job and _I think_ this is the lowest effort change I've seen in the discussions/issues on the repository.

I noticed this message from you in [another comment](https://github.com/bensheldon/good_job/issues/985#issuecomment-1674928336):
> I think the most delicate part of providing different job querying patterns is database indexing: GoodJob would need to have migrations that create indexes for both patterns (over-indexing is its own performance problem), or describe that it's not as simple as only setting a configuration option. I'm approaching the "Maybe I should write a book about GoodJob"-stage 😁

Would you be able to let me know your latest thoughts around the best implementation path? With some high-level guidance, I'd be happy to take this on

pgvsalamander:
> I believe priority only makes sense when either queues are not well-designed, or there is insufficient capacity

I noticed this referenced on #753, and wanted to provide a use case in support of job priorities - efficiently updating tree/graph structures.

In one of my projects I'm working with a massive graph structure (+100,000 nodes) where the state of a node is deterministic and based on the state of the nodes it's attached to.  While not a true tree, the graph has a rough top and bottom and a node's state mainly depends on the nodes above it in the graph.  Any node in the graph may be updated at any time, including multiple nodes can be modified simultaneously.

[...]

PR #1645, opened 2025-06-03
order by scheduled_at instead of created_at when dequeueing job

lsylvester:
In our production environment, we have a largish number of jobs (approx 70,000) that are scheduled to run in the future.

When performing the query to select a job as part of `perform_with_advisory_lock` in our production environment, we have noticed that the query can be quite slow (p50 is around 50ms). 

The query is using the `index_good_job_jobs_for_candidate_lookup` index on priority and created_at even though the vast majority of the records are not eligible to be returned due to the condition on the scheduled_at column. 

This PR changes the order that is used when finding a job to use the `scheduled_at` instead of the `created_at` and adds an index for this query. This allows the same index to be used for ordering the records as well as finding ones that are eligible to execute and is now performing this query much faster (p50 is 2ms).

It is not clear to me if there is any reason to prefer created_at over scheduled_at when ordering jobs to perform, so this PR just changes the order, but if there is a case where created_at is strongly preferred this could be changed to a configuration option instead.

There may be indexes that are no longer required once this index and query change is in place, but I think that they would needy to be removed seperately in a later version so that the indexes aren't removed while the queries that use them could still be running.

PR #1666, opened 2025-07-31
Add configuration for disabling priority

nickstanish:
Closes https://github.com/bensheldon/good_job/issues/1065
Relates to https://github.com/bensheldon/good_job/pull/1645

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,48 +20,55 @@ For more of the story of GoodJob, read the [introductory blog post](https://isla
 <details markdown="1">
 <summary><strong>📊 Comparison of GoodJob with other job queue backends (click to expand)</strong></summary>
 
-|                 | Queues, priority, retries | Database                              | Concurrency       | Reliability/Integrity  | Latency                  |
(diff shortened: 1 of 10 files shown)


bensheldon:
@nickstanish just wanted to quickly thank you for this. I will find the time to review it. 

> Wondering if this should be paired with ordering by scheduled_at since both would be impacting new indexes

We don't need to jam too much scope in, but I imagine both will be _very_ similar to implement. So I would say do them as separate PRs, but we can do one right after the other (and let's do the simplest one first).

nickstanish:
@bensheldon I did more benchmarking with ordering by the `scheduled_at` too and I'm seeing 16-76% faster processing 
I added the results here https://github.com/nickstanish/good_job/pull/1

bensheldon:
> Would it be easier to merge both configuration changes without any migrations? I'm unsure how we should handle it since the configuration values could be changed at any time

Good thought. We can make the judgement that if someone is applying optional configuration, they're also caught up on migrations.

I need to carve out some time in my schedule to review this. But no overall reservations. Just want to feel confident we've got all the pieces in place to safely roll out.

PR #1668, opened 2025-08-07
Reorganize Dashboard header, ensure favicon svg is encoded

bensheldon:
I'm not entirely sure if the favicon is to blame, but I've had a few weird experiences of the CSS or SVG spritesheet not being loaded during local development.


itsalongstory:
> Follow-up to #1664

The code was copied from https://emojitofavicon.com

PR #1672, opened 2025-08-27
Optimize job_class scope performance by using column instead of JSON extraction

### LINKED ISSUE #1671: params_job_class JSON extraction causing performance issues - why not use job_class column?
Disclaimer: This issue's text is very verbose due to Claude's assistance but although the main gist is simple I think the verbosity still helps comprehend the issue quickly.

We're experiencing performance issues and statement timeouts (set at 10s) in the GoodJob admin dashboard due to expensive JSON extraction queries. This is related to https://github.com/bensheldon/good_job/issues/876.

  **Production Impact**

  - Database size: 500k+ job records
  - Timeout errors: 10+ second queries causing 500 errors in admin interface
  - Affected pages: Job filtering dropdowns in on dashboard load

  **Root Cause Analysis**

  The issue stems from GoodJob::Job.params_job_class using JSON extraction instead of the dedicated job_class column:

  **Current Implementation (GoodJob 4.11.2)**
```ruby
  def params_job_class
    json_string(arel_table['serialized_params'], 'job_class')
  end
```
  **Generated SQL (causing timeouts)**

  -- This query times out with 500K records
```sql
  SELECT COUNT(*) AS "count_all",
         ("good_jobs"."serialized_params" ->> 'job_class') AS "good_jobs_serialized_params_job_class"
  FROM "good_jobs"
  GROUP BY ("good_jobs"."serialized_params" ->> 'job_class')
```
  Execution time: 13.0ms+ per request (2.5+ seconds on larger datasets)

  **Why Existing Indexes Don't Help (according to Claude)**

  Adding an index for serialized params doesn't help:
 ` t.index "((serialized_params ->> 'job_class'::text))", name: "index_good_jobs_on_job_class_expression"`

  While this expression index can speed up WHERE clauses filtering by job_class, it doesn't solve the fundamental performance problem with GROUP BY +
  COUNT operations:

  1. JSON extraction overhead: Every row (500k+) still requires JSON parsing and key extraction during the GROUP BY operation, even with the index
  2. Aggregation cost: PostgreSQL must process all extracted values to perform grouping and counting - the index helps locate rows but doesn't
  eliminate the computational cost of JSON operations
  3. Index scan vs computation: The bottleneck isn't finding the data, it's the repeated JSON extraction across hundreds of thousands of rows during
  aggregation

  **Proposed Solution**

  Since modern GoodJob versions have a dedicated job_class column, we can eliminate JSON extraction entirely:

  **Proposed Implementation**
```ruby
  def params_job_class
    arel_table[:job_class]
  end
```
  **Generated SQL (with patch)**
```sql
  -- This query is fast with dedicated column + index
  SELECT COUNT(*) AS "count_all",
         "good_jobs"."job_class" AS "good_jobs_job_class"
  FROM "good_jobs"
  GROUP BY "good_jobs"."job_class"
```
  Execution time: 1.1ms per request

  **Performance Results**

  We implemented this as a monkey patch and tested on production data:

  | Approach          | SQL                                 | Execution Time | Improvement  |
  |-------------------|-------------------------------------|----------------|--------------|
  | Current (JSON)    | (serialized_params ->> 'job_class') | 13.0ms         | Baseline     |
  | Proposed (Column) | good_jobs.job_class                 | 1.1ms          | 11.8x faster |

  **Complete fix with index**

  Adding a standard index on the job_class column provides optimal performance:
  CREATE INDEX CONCURRENTLY index_good_jobs_on_job_class ON good_jobs (job_class);

  **Our Temporary Fix**

  We've implemented this as a monkey patch while awaiting upstream resolution:
```ruby
  # config/initializers/good_job_performance_patch.rb
  module GoodJob::JobExtensions
    extend ActiveSupport::Concern

    class_methods do
      def params_job_class
        arel_table[:job_class]
      end
    end
  end

  Rails.application.reloader.to_prepare do
    GoodJob::Job.prepend(GoodJob::JobExtensions)
  end
```
**Questions for @bensheldon** 

  1. Is there any reason NOT to use the dedicated job_class column instead of JSON extraction?
  2. Migration strategy: Would you prefer this change to be:
    - Direct replacement of params_job_class method
    - Gradual migration approach with deprecations?
  3. Index strategy: Should GoodJob migrations automatically create an index on job_class column?

  **Environment**

  - GoodJob version: 4.11.2
  - Rails version: 8.0
  - PostgreSQL version: 15+
  - Production scale: 523K+ job records

ScotterC:
- Replace JSON extraction in params_job_class with direct column access
- Add concurrent index on job_class column for improved query performance
- Update migration tracking to include job_class index check
- Add comprehensive tests verifying column-based filtering

  Fixes #1671

### DIFF
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -290,8 +290,7 @@ module GoodJob
   # For use in tests/CI to validate GoodJob is up-to-date.
   # @return [Boolean]
   def self.migrated?
-    GoodJob::Job.concurrency_key_created_at_index_migrated? &&
-      GoodJob::Job.job_class_index_migrated?
+    GoodJob::Job.job_class_index_migrated?
   end
 
   # Pause job execution for a given queue or job class.
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -290,7 +290,8 @@ module GoodJob
   # For use in tests/CI to validate GoodJob is up-to-date.
   # @return [Boolean]
   def self.migrated?
-    GoodJob::Job.concurrency_key_created_at_index_migrated?
+    GoodJob::Job.concurrency_key_created_at_index_migrated? &&
+      GoodJob::Job.job_class_index_migrated?
   end
 
   # Pause job execution for a given queue or job class.
--- a/spec/app/models/good_job/job_spec.rb
+++ b/spec/app/models/good_job/job_spec.rb
@@ -1185,5 +1185,24 @@ RSpec.describe GoodJob::Job do
         end
       end
     end
+
+    describe '.job_class' do
+      it 'filters jobs by job_class column directly, not JSON extraction' do
+        job = described_class.create!(job_class: "TestJob", serialized_params: { job_class: "SerializedJob" })
+
+        # Should find jobs by the column value, not the JSON value
+        expect(described_class.job_class("TestJob")).to contain_exactly(job)
+        expect(described_class.job_class("SerializedJob")).to be_empty
+      end
+
+      it 'uses the job_class column for performance' do
+        # Verify that the scope uses the column, not JSON extraction in WHERE clause
+        scope_sql = described_class.job_class("TestJob").to_sql
+        expect(scope_sql).to include('"good_jobs"."job_class" =')
+        # Should not use JSON extraction operators in WHERE clause
+        expect(scope_sql).not_to include('serialized_params->')
+        expect(scope_sql).not_to include('->>')
+      end
+    end
   end
 end
--- a/spec/app/models/good_job/job_spec.rb
+++ b/spec/app/models/good_job/job_spec.rb
@@ -1185,24 +1185,5 @@ RSpec.describe GoodJob::Job do
         end
       end
     end
-
-    describe '.job_class' do
-      it 'filters jobs by job_class column directly, not JSON extraction' do
-        job = described_class.create!(job_class: "TestJob", serialized_params: { job_class: "SerializedJob" })
-
-        # Should find jobs by the column value, not the JSON value
-        expect(described_class.job_class("TestJob")).to contain_exactly(job)
-        expect(described_class.job_class("SerializedJob")).to be_empty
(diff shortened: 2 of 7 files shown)


spec/app/models/good_job/job_spec.rb
```diff
@@ -1185,5 +1185,24 @@ def job_params
...
+    describe '.job_class' do
+      it 'filters jobs by job_class column directly, not JSON extraction' do
+        job = described_class.create!(job_class: "TestJob", serialized_params: { job_class: "SerializedJob" })
+
+        # Should find jobs by the column value, not the JSON value
+        expect(described_class.job_class("TestJob")).to contain_exactly(job)
+        expect(described_class.job_class("SerializedJob")).to be_empty
+      end
+
+      it 'uses the job_class column for performance' do
```
Earlopain, line 1198: I don't think we need this test

lib/good_job.rb
```diff
@@ -290,7 +290,8 @@ def self.deprecator
   # For use in tests/CI to validate GoodJob is up-to-date.
   # @return [Boolean]
   def self.migrated?
-    GoodJob::Job.concurrency_key_created_at_index_migrated?
+    GoodJob::Job.concurrency_key_created_at_index_migrated? &&
+      GoodJob::Job.job_class_index_migrated?
```
Earlopain, line 294: You can just check for the latest migration, no need to check for any of the past ones

```suggestion
    GoodJob::Job.job_class_index_migrated?
```

Earlopain (changes requested):
The column was added in v3 so this is good :+1: https://github.com/bensheldon/good_job/commit/e5fbe917c09c8f26affcdb158db22cdc355ef218

I see you already fixed the demo migration to use 7.1, thanks! 8.0 should be fine, just some outdated rbi I think

ScotterC:
Removed the spec and redundant migration

Earlopain (approved):

PR #1676, opened 2025-09-18
Remove obsolete property from tests

RDIL:
It was removed in #1453

Side note: please bring this option back!! I'm currently monkeypatching it in, but I don't like doing so.


bensheldon:
Thanks for catching that. 

> Side note: please bring this option back!! I'm currently monkeypatching it in, but I don't like doing so.

What are you monkeypatching? "smaller number is higher priority" is now GoodJob's default and aligns GoodJob's behavior with Active Job's expectation (#883)

bensheldon:
I agree with you conceptually, which is why GoodJob did it the other way initially. But GoodJob is _only_ an Active Job adapter, so GoodJob should do it the Active Job way.

PR #1682, opened 2025-09-23
Add custom partial for adding to the head tag

RDIL:
This PR adds a custom partial that allows the user to add new stuff into the `head` tag of the dashboard pages - the goal here is to allow easier injection of user-scripts/custom styles/theming.


bensheldon:
Thanks! I rename it to "custom_head" but otherwise looks great!

PR #1683, opened 2025-09-26
deserialize ids using AR column type after insert all

stathis-alexander:
Hello! 

We use a monkey patched `Oid::Uuid` column type to standardize on non-canonical (no dashes) form uuids, and version 7 uuids. Without going into the detail, this is for historical reasons and we're in the process of trying to correct it. PG handles this automatically under the hood since it stores uuids as 16 bytes (two ascii per byte), but it, by default returns the canonical dashed forms. 

This interferes with the `GoodJob::Adapter`'s use of `insert_all`, since it does not apply the type deserialization methods and thus the logic to identify inline execution fails since the `id`'s don't match.

### DIFF
--- a/lib/good_job/adapter.rb
+++ b/lib/good_job/adapter.rb
@@ -78,6 +78,11 @@ module GoodJob
             job_id_to_provider_job_id[aj_id] = id
           end
 
+          active_jobs.each do |active_job|
+            active_job.provider_job_id = job_id_to_provider_job_id[active_job.job_id]
+            active_job.successfully_enqueued = active_job.provider_job_id.present? if active_job.respond_to?(:successfully_enqueued=)
+          end
+
           jobs.each do |job|
             job.instance_variable_set(:@new_record, false) if job_id_to_provider_job_id[job.active_job_id]
           end


stathis-alexander:
Hi @bensheldon ,

This change is unfortunately incomplete for our use case because of the way the advisory lockable concern works. I cannot see a way of generalizing that logic in a way that works for all consumers and is not specific to our system, so I may close this unless you see something I don't. 

Here's the issue: 
1. `AdvisoryLockable` composes SQL to run queries on the database. In particular, the [lock key](https://github.com/bensheldon/good_job/blob/main/app/models/concerns/good_job/advisory_lockable.rb#L202) is composed from bound variables passed in from the app level. 
2. The app level version of our keys that get passed in as bound variables are non-canonical dashless form. 
3. The DB level, however, is converted to text which is _always_ canonical form.

One option is to modify
```SQL
substr(md5($1::text), 1, 16))
```
to be 
```
substr(md5($1::uuid::text), 1, 16))
```
but that makes this incredibly opinionated about the lock column, which appears to be configurable, and the type of data it stores. 

We are casting our `uuid` values into a custom `Uuid` class which handles comparison against strings and other things for us, so our solution looks something like:
```ruby
Rails.application.config.after_initialize do
  module GoodJobAdvisoryLockableKeyPatch
    def lockable_column_key(column: self.class._advisory_lockable_column)
      unique_key = self[column]
      unique_key = unique_key.canonical if unique_key.is_a?(Uuid)

      "#{self.class.table_name}-#{unique_key}"
    end
  end
  GoodJob::Job.include(GoodJobAdvisoryLockableKeyPatch)
  GoodJob::Process.include(GoodJobAdvisoryLockableKeyPatch)
  GoodJob::BatchRecord.include(GoodJobAdvisoryLockableKeyPatch)
end
```

Issue #1090, opened 2023-09-26
Recommended method to set up job execution timeout


francois-ferrandis:
Hello there! :wave: 

First of all, thank you for this great gem, it solves many problems that we have and it is wonderfully documented. :heart_eyes: 

This issue is kind of a followup to #19.

My team and I were wondering whether it is dangerous to use the `Timeout.timeout` method as demonstrated in [README.md](https://github.com/bensheldon/good_job/blob/98d022f0ee291163643bfaa731764dfd76b4cbd4/README.md#timeouts). We have come across several articles explaining that using this method puts you at risk:
- https://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/
- http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html

So we got curious as to why there is no mention of the risks in the README even though you seem to know about them (#19)! Is there something that would make you consider the risks here are acceptable? We would love to have your input on this :ear: 

I just took a peek at other librairies that offer a timeout feature:
- Delayed Job [uses `Timeout.timeout`](https://github.com/collectiveidea/delayed_job/blob/92fb435131d05b8a8715ac52acaacd9cde89619c/lib/delayed/worker.rb#L231)
- Resque seems to save a [timeout value](https://github.com/search?q=repo%3Anesquena%2Fbackburner+%22respond_timeout%22&type=code) to DB but I could not find how it is used
- Sneakers is event [more mysterious](https://github.com/search?q=repo%3Ajondot%2Fsneakers%20timeout_job_after&type=code) about how it handles timeouts

Most librairies don't offer that feature, for good reasons I presume.

Thank you so much for your time! :pray:

bensheldon:
@francois-ferrandis great question! 

Yes, it is dangerous to use `Timeout.timeout` in GoodJob. That's because GoodJob, like Puma, is multithreaded and `Thread.raise/kill` can leave shared resources (like Active Record database connection pool) in uncertain states.

> we got curious as to why there is no mention of the risks in the README

Simply overlooked on my part. I think I wrote that in the very early days of GoodJob and that section is inadequate today. I think my thinking there was that if _your_ code is safe to be wrapped in a `Timeout.timeout` then it's safe for GoodJob too... which is true, but doesn't mention that it's very, very, very likely that your code isn't safe to do that.

I do think the Readme should mention the dangers.

Issue #19, opened 2020-03-09
Add a job timeout configuration to time out jobs that have run too long


bensheldon:
Timeouts in Ruby are rather dangerous so this should be given some thought as it's probably unavoidable to entirely avoid the risk of corruption of shared state.

Concurrent Ruby has `Cancellations` but it seems impossible to safely cancel a uncontrolled code (e.g. active job): http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Cancellation.html

I think the biggest danger around an ActiveJob is corrupting the ActiveRecord database pool. So adding a timeout solely around the ActiveJob code, but within the executor wrap might work. Even better would be running it within an ActiveJob `around` hook so that it could potentially be caught by ActiveJob's `rescue_from`/`retry_on` handlers.

bensheldon:
Killing threads is not something actually supported by Concurrent Ruby because of the difficulty in doing it safely with user code. Added instructions for configuring timeouts in user code: https://github.com/bensheldon/good_job/pull/39

PR #1684, opened 2025-10-02
Add warning and alternative to Ruby Timeout example

seanpdoyle:
Follow-up to [#19][]
Closes [#1090][]

Avoid using Ruby's built-in [Timeout](https://github.com/ruby/timeout) mechanism
([1](https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/), [2](https://blog.headius.com/2008/02/rubys-threadraise-threadkill-timeoutrb.html)). Instead, declare either of Active Job's [discard_on][] or [retry_on][] to handle the underlying mechanism's timeout exceptions (when available).

For example, rescue from `Net::OpenTimeout` or `Net::ReadTimeout` and discard the job:

```ruby
class MyJob < ApplicationJob
  discard_on Net::OpenTimeout, Net::ReadTimeout

  def perform(uri)
    Net::HTTP.start(uri.host, uri.port, open_timeout: 3, read_timeout: 3) do |http|
      http.request(...)
    end
  end
end
```

[#19]: https://github.com/bensheldon/good_job/issues/19
[#1090]: https://github.com/bensheldon/good_job/issues/1090
[discard_on]: https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-discard_on
[retry_on]: https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on

### DIFF
--- a/README.md
+++ b/README.md
@@ -1060,7 +1060,7 @@ end
 Avoid using Ruby's built-in [Timeout](https://github.com/ruby/timeout) mechanism
 ([1](https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/),
 [2](https://blog.headius.com/2008/02/rubys-threadraise-threadkill-timeoutrb.html)).
-Instead, declare either of Active Job's [discard_on][] or [retry_on][] to handle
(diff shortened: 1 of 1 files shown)


ghiculescu:
Regarding the `around_perform` example: I _think_ that the error should subclass `Exception` instead of `StandardError`. Otherwise you open yourself up to this issue: https://github.com/Betterment/delayed/issues/64

PR #1686, opened 2025-10-16
Add CI for Rails 8.1 (rc) and use postgres 18

Earlopain:
Rails 8.1 is not to far of from being released. Postgres 18 is also available now (still testing against pg10 in the matrix)


bensheldon:
Thank you! 🙏🏻

PR #1688, opened 2025-10-17
Add environment configuration option and update dashboard display

amkisko:
This addresses the need for customizable environment display in the GoodJob dashboard, providing more flexibility for different deployment scenarios and branding requirements.

- Introduced a new configuration option `environment` to specify the environment displayed in the dashboard, defaulting to `Rails.env`.
- Updated the navbar to reflect the configured environment instead of directly using `Rails.env`.
- Added tests to ensure the correct behavior of the new environment configuration option.

## Usage Examples

```ruby
# Via direct configuration
GoodJob.configure do |config|
  config.environment = "Production Cluster A"
end

# Via Rails configuration
  config.good_job.environment = "Staging Environment"
end

# Via environment variable
ENV['GOOD_JOB_ENVIRONMENT'] = "Custom Environment"
```

### DIFF
--- a/app/views/good_job/shared/_navbar.erb
+++ b/app/views/good_job/shared/_navbar.erb
@@ -5,7 +5,7 @@
     <div class="d-lg-none d-flex flex-fill justify-content-end me-2">
       <span class="navbar-text">
         <%= render_icon "globe", class: "align-text-bottom" %>
-        <small><%= GoodJob.configuration.environment.capitalize %></small>
+        <small><%= GoodJob.configuration.environment %></small>
       </span>
     </div>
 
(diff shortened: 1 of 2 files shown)


amkisko:
@bensheldon Hey! The motivation here is when environment name is different, but Rails.env stays production always — basically proper titles for tabs or visual difference between tabs.

I think amount of code will stay almost same if I move it to helper method `display_environment`, to my mind this is fully configuration thing and should be set once during the load, no need to recalculate it.

amkisko:
And sure there is always an option to fork, I keep this as last resort if there will be a clear need to go that way (no fun with git sourced ruby gems).

bensheldon:
Sorry, I have been thinking a lot about this PR. I don't like adding to GoodJob configuration things that aren't related to job execution. The Dashboard is a very large surface area and I worry that there will be lots of things people will want to slightly change.

There are custom partials for the dashboard that allow injecting HTML. It would be possible to inject a script tag and override these values using javascript. ...and even easier if they had some HTML classes on them to target.

Issue #1371, opened 2024-06-11
Have total_limit and perform_throttle use separate keys


drale2k:
I named the issue based on what i believe the cause is. 

I have a job to fetch news stastics on a schedule, from a rate limited API. I want to ensure the following for my job:

1. Run max of 15 jobs per minute (use `perform_throttle`)
2. Run a maximum of 1 job for a given news at the same time (use `total_limit: 1` + `news.id` as the key)

```ruby
class News::FetchStatsJob < ApplicationJob
  queue_as :later

  include GoodJob::ActiveJobExtensions::Concurrency
  good_job_control_concurrency_with(
    total_limit: 1,
    key: -> { "#{self.class.name}-#{arguments.first.id}" },
    perform_throttle: [10, 1.minute]
  )

  def perform(news)
   ...
  end
end
```

The issue i am running into is that with the above settings, i can run a lot more than 10 jobs per minute. If i only use `perform_throttle` without the `key` and `total_limit` then it will properly stop at after 10 and retry the remaining jobs - but it will allow to queue and execute multiple jobs for the same `news.id`. 

I believe the issue is that by setting the key, `total_limit` and `perform_throttle` are now both scoped to the key? So in my case, it means throttle at max 10/min and limit to 1 job max only for jobs with the same `news.id`? 

Is there a way i can achive what i described above? I would like to:

1. Run max of 15 jobs per minute of the class `News::FetchStatsJob`
2. But only 1 job at the time for a given `news.id`. So stats for a single news will not be fetched multiple times at once.

bensheldon:
Unfortunately Concurrency Controls only allow one set of rules with one key. I'd like it to do more, but it hasn't risen to the top of my todo list.

I've [imagined](https://github.com/bensheldon/good_job/issues/1084#issuecomment-1777404960) that concurrency-controls could use labels instead of the single key. That would decouple the concurrency rules from the jobs more cleanly, I think.

I'm imagining something like:

```
good_job_concurrency_rule(
  total_limit: 1,
  label: -> { "#{self.class.name} },
)
good_job_concurrency_rule(
  perform_throttle: [10, 1.minute]
  label: -> { "#{self.class.name}-#{arguments.first.id}" }
)
```

I'm unsure whether adding those rules should specifically add the labels, or whether they should be set separately, but that seems like a minor detail once the concurrency rules were implemented.

drale2k:
Understood, that would really useful. Any suggestions how i could approach my issue?

[...]

Issue #1084, opened 2023-09-22
Unique job


gagalago:
I have lot of duplicate job enqueued that take some time to perform. that is a waste of resources to run them all as one them is enough.

I solved it by adding this callback:

```ruby
  # unique job with same arguments, keep the oldest
  #   call the job only if there is no other job with same arguments in the queue
  before_enqueue do |job|
    other_jobs_arguments = GoodJob::Job
      .where(finished_at: nil)
      .where(job_class: job.class.name, concurrency_key: job.good_job_concurrency_key)
      .map do |other_job|
        ActiveJob::Arguments.deserialize(other_job.serialized_params["arguments"])
      end

    throw(:abort) if other_jobs_arguments.any?(job.arguments)
  end
```

I was wondering what you think about my solution? Do you think that we generalize it in a way that we can integrate it into good_job? 

ps: some more general alternatives that use redis to work are https://github.com/mhenrixon/sidekiq-unique-jobs or https://github.com/veeqo/activejob-uniqueness . From my understanding my implementation is similar to their `until_executed`.

sasharevzin:
I think this issue was answered in https://github.com/bensheldon/good_job/issues/531

bensheldon:
💖 Thank you @sasharevzin for that link! I think that's right; the feature to use is Concurrency Controls: https://github.com/bensheldon/good_job#concurrency-controls

And sorry @gagalago for overlooking this when you originally posted 😞

gagalago:
you are right but I didn't explain enough my use-case :see_no_evil: 

something that I want to be able to do is the following:
```ruby
class ExampleJob < ApplicationJob
  include GoodJob::ActiveJobExtensions::Concurrency


  good_job_control_concurrency_with(
    # only one job at the time for the same record, prevent me to do record.lock! in the job
    perform_limit: 1,
    perform_key: -> { "ExampleJob-#{arguments.first}" },
    # only one job with exactly same arguments enqueued at any time
    enqueue_limit: 1,
    enqueue_key: -> { "ExampleJob-#{arguments.first}-#{arguments.second}" }
  )

  def perform(record, attributes)
    # do something
  end
end
```

due to the fact that I need two different key for enqueue and perform. I had to do the before enqueue myself.

now that I understand better how is working good_job, what do you think to have the same behavior as of the limit but for the key as I presented in my example?

bensheldon:
[...]

PR #1700, opened 2025-12-17
Allow multiple concurrency rules per job via labels

bscofield:
This isn't complete, but I wanted to open it for discussion, at least.

We've got a job that hits an external service provider. There's an overall rate limit on the number of requests we can make to them (say, 60/hour), but we also want to throttle jobs that operate against a single object -- which means we need concurrency controls with different keys (job name vs. argument-specific). #1084 and #1371 are related issues here, and both mention the idea of using labels on a job as concurrency keys. The latter issue also sketches Ben's idea for the ability to set distinct rules around concurrency, each with their own key.

This PR takes a swing at implementing that. There are some differences from the version Ben originally described, and some things I already know I'd like feedback on:

- Rules are linked to a stage (enqueue, perform, or total); you can't mix `perform_throttle` and `total_limit` and `enqueue_limit` in this structure, for instance -- those would be three distinct rules.
- I've updated the legacy configuration to create new-style rules behind the scenes, if labels are available; those rules would then be run before enqueue and perform, instead of the key-based checks. Maybe we'd prefer to leave that whole codepath unchanged, though?
- Related: we should raise or warn if someone tries to configure rules in a job that doesn't support labels -- or maybe include the Labels module automatically in that case? 
- I haven't updated the README given that the answers to some of the above might change what'd be needed.


bscofield:
@bensheldon OK, hopefully this is a little closer to the mark. To your feedback specifically:

- Labels are no longer set automatically; you've got to `JobClass.set(good_job_labels: ["..."]).perform_later(*args)` now to get 'em in place.
- The Labels module is included automatically when you pull in the Concurrency module.
- Labels are used to find jobs that might be concurrency-violating; keys are used for the advisory lock.

Other tweaks:

- I went back to the state-specific limits and throttles (so `enqueue_limit`, `perform_throttle`, etc., instead of scoping each rule to enqueue or perform) -- it makes complex rules a little easier.
- I added a Rule class to encapsulate a lot of the logic around resolving limits, throttles, keys, and labels.
- I added `Rule.evaluate(job, stage)`, which runs `check_enqueue` or `check_perform` in the job context as appropriate.
- I moved limit / throttle resolution for the legacy config to an ephemeral Rule, but left the concurrency checks for that config in the Job itself (`_check_enqueue` and `_check_perform`) -- there's duplication, but I think it makes Rule easier to grasp.

## Examples of use

```ruby
class TestJob
  
  # Limits total running concurrency for `TestJob` jobs to 1
  good_job_concurrency_rule(total_limit: 1)
  
  # Only one job can be enqueued with the following call at a time; it will use `TestJob` as the advisory lock key
  # TestJob.set(good_job_labels: ["Alice"]).perform_later(name: "Alice")
  good_job_concurrency_rule(label: -> { arguments.first[:name] }, enqueue_limit: 1)
  
  # Use `TestJob::Alice` as the advisory lock key -- but will only run if _no_ other jobs of any class are running (because there's no label to scope that query)
  # TestJob.perform_later(name: "Alice")
  good_job_concurrency_rule(key: -> { "#{self.class.name}:#{arguments.first[:name]}", perform_limit: 1)

  # Up to 10 jobs can be enqueued with the "static" label, but only one job can be performed per minute.
  # The advisory key is `TestJob`
  # TestJob.set(good_job_labels: ["static"]).perform_later(name: "Alice")
  good_job_concurrency_rule(label: "static", enqueue_limit: 10, perform_throttle: [1, 1.minute])
end
```

lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -90,12 +120,12 @@ def deserialize(job_data)
...
-                enqueued_within_period = GoodJob::Job.where(concurrency_key: key)
-                                                     .where(GoodJob::Job.arel_table[:created_at].gt(throttle_period.ago))
-                                                     .count
+                enqueued_within_period = query_scope
+                                         .where(GoodJob::Job.arel_table[:created_at].gt(throttle_period.ago))
+                                         .count
 
                 if (enqueued_within_period + 1) > throttle_limit
-                  logger.info "Aborted enqueue of #{job.class.name} (Job ID: #{job.job_id}) because the concurrency key '#{key}' has reached its throttle limit of #{limit} #{'job'.pluralize(limit)}"
+                  job.logger.info "Aborted enqueue of #{self.class.name} (Job ID: #{job.id}) because the concurrency key '#{key}' has reached its throttle limit of #{throttle_limit} #{'job'.pluralize(throttle_limit)}"
```
bensheldon: **Bug: Two errors in this log message — `self.class.name` and `job.id`**

1. **`self.class.name`** — inside the `Rule` class, `self` is the Rule instance, so this logs `"GoodJob::ActiveJobExtensions::Concurrency::Rule"` instead of the job class name. The limit branch above (line 114) correctly uses `job.class.name`.

2. **`job.id`** — `ActiveJob::Base` does not define an `#id` method. In production this will raise `NoMethodError`. It should be `job.job_id`. (The tests stub `def id; job_id; end` on their test job classes, which masks this.)

PR #1705, opened 2026-01-01
Add Ruby 4.0 to CI

Earlopain:
Merry (late) christmas and also :fireworks:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,13 +113,14 @@ jobs:
             gemfile: rails_8.1
           - ruby: 3.1
             gemfile: rails_8.1
-          # Rails head is >= 3.2
+          # Rails head is >= 3.3
           - ruby: "3.0"
             gemfile: rails_head
           - ruby: 3.1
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -27,12 +27,16 @@ rails_versions = {
...
+# Ruby 4.0 has moved this gem to a bundled gem. Rails 6.1 doesn't declare it.
+gem "benchmark" if ENV["RAILS_VERSION"] == "6.1"
+
 platforms :ruby do
   gem "bootsnap"
   gem "dotenv-rails"
   gem "foreman"
   gem "gem-release"
   gem "github_changelog_generator", require: false
+  gem "rdoc", require: false
```
Earlopain: The Rakefile requires `rdoc/task`. Newer rails versions depend on irb which in turn depends on rdoc. Since Ruby 4.0 rdoc has become a bundled gem, so it needs to be specified explicitly for those older versions.

bensheldon:
Thank you! 💪🏻 🎄 🎆

Issue #1623, opened 2025-03-27
Adapt `AdvisoryLockable` to use more secure algorithm instead of MD5


edmunteanu:
I've stumbled upon an interesting issue while setting up a containerised Rails project on Microsoft Azure. The database we used for our infrastructure is called "Azure Database for PostgreSQL flexible server". When we created this resource for our production environment a few months ago, Microsoft was enforcing [Federal Information Processing Standards](https://en.wikipedia.org/wiki/Federal_Information_Processing_Standards) (FIPS for short) for this type of resource - this has in the meantime been rolled back, but it offered some valuable insight into what might come in the future.

One such standard was the deactivation of insecure cryptographic algorithms, such as **MD5**. Naturally, since this algorithm was not supported, trying to use it was causing an error, which looked something like the following:

```
PG::InternalError: ERROR:  could not compute MD5 hash: disabled for FIPS
```

<details>
<summary>Stack Trace</summary>
<img src="https://github.com/user-attachments/assets/671571c6-933b-41be-9bf0-0a72e016466f">
</details>

This could indeed be confirmed by connecting to the DB directly via `psql` and running `SELECT md5('test') AS md5_hash;`, which would return an error similar to the one above.

Upon a closer look, it became clear that the cause for this error is the [`AdvisoryLockable`](https://github.com/bensheldon/good_job/blob/main/app/models/concerns/good_job/advisory_lockable.rb) module, which is to my understanding using MD5 hashed keys to query advisory locks.

Some further research around this topic led me to the following [PR](https://github.com/sEpt0r/pgcopydb/pull/1), which seems to be tackling the very same issue by changing the hashing algorithm to SHA256.

Considering all of this information, I believe it's a good idea to adapt the cryptographic algorithm used in the `AdvisoryLockable` module to something that respects also higher security standards. Please let me know what you think as well - I'll gladly open a PR to tackle this change while also accounting for your thoughts.

bensheldon:
There isn't any security implication because the hashing doesn't touch any user-provided data. Though I can understand it's unusable if the database provider has disabled md5. 

I'd accept a PR if you wanted to add a configuration option to changing the algorithm to sha256. It would need to be optional because that would be a breaking change to modify the lock value. 

You'll also need to cut the output in half because the goal is to map the value to a bigint tuple.

edmunteanu:
[...]

PR #1722, opened 2026-03-16
Introduce advisory lock key customization support methods

amkisko:
Hey!

This pull request introduces advisory lock key customizability both by quick configuration and module patching.

- It covers core PostgreSQL functions and in addition pg_crypto with uuid_v5.
- Default kept md5 function.
- It changes bitwise operations to substring trimming by char count, I am okay with both options, substring reads a bit easier, anyways easy to get back to bitwise operations as in original implementation.

I am still thinking if there should be any mention of `hashtextextended` and `hashtext` (considering also complex implementation), as they are not recommended for usage by PostgreSQL team and intentionally left out from official documentation, although there is quite much usage for table partitioning (even more sensitive area).

Also observed bug/feature in method `def owns_advisory_lock?(key: lockable_key)` which had duplicate code fully identical to class method.

Thank you!

References:
- https://github.com/bensheldon/good_job/issues/1623

### DIFF
--- a/README.md
+++ b/README.md
@@ -1319,7 +1319,9 @@ To explain where this value is used, here is the pseudo-query that GoodJob uses 
   )
 ```
 
-By default, advisory lock keys are derived with `md5`. If you need a different hash strategy, set it globally:
+By default, advisory lock keys are derived with `md5`. The string key (e.g. `good_jobs-<active_job_id>`) is hashed and truncated to a 64-bit bigint, which is the key space PostgreSQL advisory locks operate in. `md5` is used for its wide availability — no PostgreSQL extensions required — and good bit distribution, not for any cryptographic property.
+
+If you need a different hash strategy, set it globally:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


app/models/concerns/good_job/advisory_lockable.rb
```diff
@@ -289,6 +297,29 @@ def _advisory_lockable_column
         advisory_lockable_column || primary_key
       end
 
+      def advisory_lock_bigint_sql(value_sql)
+        case advisory_lockable_hash_function.to_s.downcase
+        when "hashtextextended"
+          "hashtextextended((#{value_sql})::text, 0)"
+        when "hashtext"
+          "((hashtext((#{value_sql})::text)::bigint << 32) + (hashtext(('good_job-' || (#{value_sql})::text))::bigint & 4294967295::bigint))"
```
bensheldon: Also explain the values here? I don't want to have "good_job-" prefix in AdvisoryLockable.

amkisko:
Nice, thank you for moving this forward and polishing!

PR #1761, opened 2026-04-20
Add update migration to drop redundant active_job_id index

### LINKED ISSUE #1661: Drop Duplicate index
We use `good_job 4.10.0`. The `good_job` table in the database has 19 indexes. While indexes can improve `SELECTs`, many indexes can slow down data modifications. 
Since `good_job` table is used for a queuing purposes, it undergoes `UPDATEs` and `DELETEs` in order of magnitude of hundreds of thousands per day.
We have identified a duplicate index on the table that can be safely dropped.
```
"index_good_jobs_on_active_job_id" btree (active_job_id) 
is covered by
"index_good_jobs_on_active_job_id_and_created_at" btree (active_job_id, created_at)
```
Opening an issue to consider implementing this suggestion

MahmoudBakr23:
Closes #1661

## What

Adds update migration `15_remove_duplicate_active_job_id_index` to drop `index_good_jobs_on_active_job_id` from the `good_jobs` table, and updates `GoodJob.migrated?` to reflect the latest migration.

## Why

`index_good_jobs_on_active_job_id` is fully redundant. PostgreSQL can satisfy single-column lookups on `active_job_id` using the composite `index_good_jobs_on_active_job_id_and_created_at` as a prefix scan — the single-column index adds no selectivity and is pure write overhead.

The `good_jobs` table undergoes `INSERT`, `UPDATE`, and `DELETE` at high volume, so every redundant index adds measurable write amplification and lock contention.

## History

This was originally fixed in PR #1181 (December 2023) as migration `10_remove_good_job_active_id_index`. That migration was intentionally removed during the v4 major release (July 2024), which reset the numbered update migration series. The install template has been correct since v3 (no single-column index), but existing pre-v4 installs that never ran migration #10 still carry the redundant index. This re-lands the fix as migration `#15` for the post-v4 world.

## Changes

- `lib/generators/good_job/templates/update/migrations/15_remove_duplicate_active_job_id_index.rb.erb` — new update migration; idempotent guard via `index_name_exists?`, uses `algorithm: :concurrently`, fully reversible
- `lib/good_job.rb` — `GoodJob.migrated?` updated to check absence of the redundant index (replaces the previous #14 check per the established pattern)

## Testing

- New installs (current template never creates the single-column index): migration is a no-op, `migrated?` returns `true` ✓
- Old installs that haven't run this migration yet: index still present, `migrated?` returns `false` ✓
- Old installs after running this migration: index dropped, `migrated?` returns `true` ✓
- `spec/generators/good_job/update_generator_spec.rb` "has files with unique number prefixes" passes ✓
- `spec/integration/` (62 examples) and `spec/lib/good_job_spec.rb` (18 examples) all pass ✓

### DIFF
--- a/lib/generators/good_job/templates/update/migrations/15_remove_duplicate_active_job_id_index.rb.erb
+++ b/lib/generators/good_job/templates/update/migrations/15_remove_duplicate_active_job_id_index.rb.erb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class RemoveDuplicateActiveJobIdIndex < ActiveRecord::Migration<%= migration_version %>
+  disable_ddl_transaction!
+
+  def change
+    reversible do |dir|
+      dir.up do
+        return unless connection.index_name_exists?(:good_jobs, :index_good_jobs_on_active_job_id)
+      end
+    end
+
+    remove_index :good_jobs, :active_job_id, name: :index_good_jobs_on_active_job_id, algorithm: :concurrently
+  end
+end
--- a/lib/good_job.rb
+++ b/lib/good_job.rb
@@ -315,7 +315,7 @@ module GoodJob
   # @return [Boolean]
   def self.migrated?
     GoodJob::Job.lock_type_migrated? &&
-      GoodJob::Job.connection.index_name_exists?(:good_jobs, "index_good_jobs_on_unfinished_or_errored")
+      !GoodJob::Job.connection.index_name_exists?(:good_jobs, "index_good_jobs_on_active_job_id")
   end
 
   # Pause job execution for a given queue or job class.


bensheldon:
@MahmoudBakr23 Thank you! you're the best! 🥇

bensheldon:
Ohh, sorry, I misunderstood. This is the key thing:

> The install template has been correct since v3 (no single-column index), but existing pre-v4 installs that never ran migration https://github.com/bensheldon/good_job/pull/10 still carry the redundant index.

It sounds like your app didn't fully run all the v3 migrations, not that GoodJob failed to include it. This was done in #1181.

MahmoudBakr23:
@bensheldon Thanks for clarifying! You're right that #1181 covered both the install template fix and update migration #_10, so GoodJob did provide the path to clean this up in v3.

The gap I had in mind is apps that ran `good_job:install` before #1181 landed, then jumped to v4 without running migration #_10 first. Since v4 reset the update migration series, `good_job:update` no longer offers a path to drop that index. The migration here is a no-op for anyone already clean, so the blast radius is pretty small.

Please correct me if I'm wrong! Unless the expectation is that users should have run all v3 update migrations before upgrading to v4, then it's all good.

bensheldon:
> the expectation is that users should have run all v3 update migrations before upgrading to v4

That's the correct expectation. Devs should always go through the x.99 release and apply migrations. 

You can pin GoodJob v3.99 in local dev for a minute, and then run good_job:update to grab the migrations if you skip it. 

For context, pruning migrations has pluses and minuses. I don't want someone who is setting up GoodJob for the very first time to be handed a pile of historical migrations.

MahmoudBakr23:
That makes total sense, thanks for explaining the reasoning. The v3.99 pin workaround is a clean solution for anyone who finds themselves in that situation. Looking forward to tackling another open issue soon!

PR #1762, opened 2026-04-21
Set default queue_select_limit to 1000

### LINKED ISSUE #1596: Set a default queue_select_limit
I've seen folks get bitten by this when scaling up significantly. 

Maybe 1k is a reasonable default that wouldn't be necessary to do this as part of a major release cycle and do it as a minor release.

MahmoudBakr23:
Closes #1596

## What

Sets `DEFAULT_QUEUE_SELECT_LIMIT = 1_000` as the fallback when no explicit `queue_select_limit` is configured via options, Rails initializer, or environment variable.

## Why

Without a default, `queue_select_limit` returns `nil`, which skips the `LIMIT` clause on the materialized CTE used to find candidate jobs. On large queues (100k+ eligible jobs), this forces PostgreSQL to fully materialize every row before acquiring an advisory lock — an expensive scan that grows unbounded with queue depth.

1000 is already the value recommended in the README and covers virtually all real-world deployments: it exceeds the database connection limits of most PaaS platforms, while still providing the intended performance benefit.

## Changes

- `lib/good_job/configuration.rb` — adds `DEFAULT_QUEUE_SELECT_LIMIT = 1_000` constant; updates `queue_select_limit` to use the inline-default pattern consistent with other numeric settings; fixes `@return` tag from `[Integer, nil]` to `[Integer]`
- `lib/good_job/cli.rb` — updates the `--queue-select-limit` description from `default: nil` to `default: 1000`
- `spec/lib/good_job/configuration_spec.rb` — adds `#queue_select_limit` spec covering the default and all three override sources (options, Rails config, environment variable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1292,7 +1292,7 @@ The recommended way to monitor the queue in production is:
 
 GoodJob’s advisory locking strategy uses a materialized CTE (Common Table Expression). This strategy can be non-performant when querying a very large queue of executable jobs (100,000+) because the database query must materialize all executable jobs before acquiring an advisory lock.
 
-GoodJob offers an optional optimization to limit the number of jobs that are queried: Queue Select Limit.
+GoodJob's Queue Select Limit restricts how many jobs are materialized per advisory lock cycle. **It defaults to `1000` — no configuration needed.** Override it if your deployment requires a different value:
 
 ```none
 # CLI option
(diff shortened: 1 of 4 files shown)


bensheldon:
@MahmoudBakr23 thanks for picking this issue up 🙇🏻

bensheldon:
@MahmoudBakr23 could you please also update the Readme too?

MahmoudBakr23:
@bensheldon With pleasure! Coming right up.

PR #1765, opened 2026-04-25
Make execution state completely Fiber-safe via Rails isolated execution state

ollym:
This updates GoodJob’s execution-local state handling for Fiber-aware Rails environments such as Falcon/Async.

Changes include:

- Replace Thread.current[:good_job_scheduler] with ActiveSupport::IsolatedExecutionState[:good_job_scheduler] for current scheduler status.
- Keep existing thread_mattr_accessor / thread_cattr_accessor usage, relying on Rails’ fiber-aware implementation when config.active_support.isolation_level = :fiber.
- Add specs proving GoodJob::CurrentThread and scheduler status are isolated across Fibers.
- Drop Rails 6.1 support and require Rails 7.0+ so ActiveSupport::IsolatedExecutionState is available.
- Require Ruby 3.2+ for modern Fiber storage support.
- Remove Rails 6.1-only compatibility code and CI/appraisal coverage.


bensheldon (commented):
Thanks for starting this! I'm skeptical that the problem here is that one usage of Thread.current which is used in a thread that GoodJob manages, but otherwise no objection to handling it differently.

It looks like you're doing agentic stuff, so speaking to your agent, it needs to change:

- This must contain only a single change related to fiber safety.
- You cannot remove compatibility with anything
- Make a compatibility object like `GoodJob::SafeContext` that does a rails version check and only then uses IsolatedExecutionContext or else continues using Thread.current
- You can make an ENV variable for isolation level (defaults to thread) in the demo app and then create a test run that sets it to fiber as a smoketest/confidence check.

PR #1766, opened 2026-04-28
Truncate long labels in dashboard badges

### LINKED ISSUE #1674: Handle long labels more gracefully
Currently labels that are decently long will be formatted weirdly in the UI:

Image

I feel like there should be a nicer way to handle this gracefully, e.g. with truncation.

MahmoudBakr23:
Fixes #1674.

## What

Long label strings caused badge overflow that broke the table row layout — the badge would expand to full label width and spill into adjacent columns.

**Before:**

Before: long labels break the table layout

**After:**

Screenshot 2026-04-28 at 23 26 47

## How

- `jobs/_table.erb` and `batches/_jobs.erb`: use Rails' `truncate` helper (length: 15) to shorten the displayed label. Only labels that are actually truncated get a `title` attribute and `data-bs-toggle="tooltip"` so the full text remains accessible on hover.
- `bootstrap_init.js`: added `setupTooltips()` (mirrors the existing `setupPopovers()`) so all `[data-bs-toggle="tooltip"]` elements are initialized on `turbo:load`.

## Tests

Added a system spec that creates a job with a long label, visits the jobs index, and asserts:
- the badge shows the truncated text (`"a_very_long_..."`)
- `data-bs-original-title` (set by Bootstrap after tooltip init) holds the full label

### DIFF
--- a/app/frontend/good_job/modules/bootstrap_init.js
+++ b/app/frontend/good_job/modules/bootstrap_init.js
@@ -15,7 +15,14 @@ function setupPopovers() {
   })
 }
 
+function setupTooltips() {
+  document.querySelectorAll('[data-bs-toggle="tooltip"]').forEach((el) => {
+    new bootstrap.Tooltip(el)
+  })
+}
+
 window.document.addEventListener("turbo:load", function() {
   showToasts();
   setupPopovers();
+  setupTooltips();
 });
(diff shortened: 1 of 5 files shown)


MahmoudBakr23:
**Styling-alignment recommendation:** The `PRIORITY`, `LABELS`, and `ATTEMPTS` sections' values are not displayed text-centered as the `QUEUE` section!

PR #1768, opened 2026-05-01
Fix PG::AmbiguousColumn in skiplocked/hybrid claim with ordered queues + concurrency rules

### LINKED ISSUE #1767: Raises `PG::AmbiguousColumn` under hybrid lock strategy on ordered queues with throttle
Thanks for adding skip locked to GoodJob. I rolled attempted to roll out `:hybrid` yesterday and hit a bug I've tried to describe as best possible below. Ultimately I had to revert back to advisory but filing it in case it saves anyone else the same surprise in the interim.

This could be a quirk of how this application I'm working on have developed but we encountered  `PG::AmbiguousColumn: ERROR: column reference "queue_name" is ambiguous` errors immediately on every job and it seems related to when these three conditions hold simultaneously:

1. `lock_strategy: :hybrid` (or `:skiplocked`)
2. A worker pool uses `+`-prefixed (ordered) queues, e.g. `GOOD_JOB_QUEUES="+a,b:1"`
3. The job uses `GoodJob::ActiveJobExtensions::Concurrency` with `perform_throttle`

`:advisory` is unaffected. 

## Production error extracts

```
ActiveRecord::StatementInvalid: PG::AmbiguousColumn: ERROR:
  column reference "queue_name" is ambiguous
LINE 1: ...ob_executions"."error" != $7) ORDER BY (CASE WHEN queue_name...
                                                             ^
```

Key stack frames:

```
good_job-4.18.2/app/models/good_job/job/lockable.rb:106:in `with_hybrid_lock_claim'
  yield(record)
good_job-4.18.2/app/models/good_job/job.rb:453:in `block in perform_with_hybrid_lock'
  job.perform(lock_id: lock_id, already_claimed: true)
...
good_job-4.18.2/lib/good_job/active_job_extensions/concurrency.rb:147:in `check_perform'
  GoodJob::Job.transaction(requires_new: true, joinable: false) do
good_job-4.18.2/lib/good_job/active_job_extensions/concurrency.rb:172:in `pluck'
  Execution.joins(:job).merge(query_scope).pluck(:active_job_id)
```

## Reproduction

This snippet reproduces the failure mode without needing to spin up a scheduler I think. It mirrors the relation that's open during a `:hybrid` or `:skiplocked` perform block, plus the join the throttle callback issues:

```ruby
GoodJob::Job.queue_ordered(%w[a b]).scoping do
  GoodJob::Execution
    .joins(:job)
    .merge(GoodJob::Job.where(concurrency_key: "test"))
    .to_a
end
# ActiveRecord::StatementInvalid:
#   PG::AmbiguousColumn: ERROR:  column reference "queue_name" is ambiguous
```

Advisory appears to be immune as it runs in an unscoped block.

## Environment

- good_job 4.18.2
- PostgreSQL 16, Rails 8.1.2.1, Ruby 3.4.7

Let me know if it would be helpful to have further information.

createdbypete:
Fixes #1767.

Under `lock_strategy: :hybrid` or `:skiplocked`, workers using `+`-prefixed ordered queues raised `PG::AmbiguousColumn: column reference "queue_name" is ambiguous` on every perform of any job using `GoodJob::ActiveJobExtensions::Concurrency`. Introduced by #1731.

The skiplocked/hybrid claim methods in `Job::Lockable` yielded the locked record while `GoodJob::Job` was still scoped to the dequeue relation, so `GoodJob::Job` queries inside the perform block (e.g. the `query_scope` built by `Concurrency#check_perform`) inherited that scope, including `queue_ordered`'s unqualified `queue_name` ORDER BY, plus `unfinished`, `where(locked_by_id: nil)`, and `limit(1)`. The advisory path already wraps its yield in `unscoped` (see `AdvisoryLockable#advisory_lock`), which is why `:advisory` was unaffected. Beyond the SQL ambiguity, the same scope leak would also silently miscount throttle limits under `:hybrid`/`:skiplocked`.

The fix wraps the yield in `unscoped { }` in `with_skip_locked_claim` and `with_hybrid_lock_claim`, matching the advisory path. It also defensively qualifies `queue_name` in the `queue_ordered` scope so it remains safe to merge into joined queries on its own. A behavioral spec parameterized over all three strategies covers the original symptom; a separate join-merge test on `queue_ordered` covers the column qualification.

### DIFF
--- a/app/models/good_job/job.rb
+++ b/app/models/good_job/job.rb
@@ -188,7 +188,7 @@ module GoodJob
     # @return [ActiveRecord::Relation]
     scope :queue_ordered, (lambda do |queues|
       clauses = queues.map.with_index do |queue_name, index|
-        sanitize_sql_array(["WHEN queue_name = ? THEN ?", queue_name, index])
+        sanitize_sql_array(["WHEN #{quoted_table_name}.queue_name = ? THEN ?", queue_name, index])
       end
       order(Arel.sql("(CASE #{clauses.join(' ')} ELSE #{queues.size} END)"))
     end)
--- a/app/models/good_job/job.rb
+++ b/app/models/good_job/job.rb
@@ -187,8 +187,9 @@ module GoodJob
     # @param queues [Array<string] ordered names of queues
     # @return [ActiveRecord::Relation]
     scope :queue_ordered, (lambda do |queues|
+      qualified_queue_name = "#{quoted_table_name}.#{adapter_class.quote_column_name('queue_name')}"
       clauses = queues.map.with_index do |queue_name, index|
-        sanitize_sql_array(["WHEN #{quoted_table_name}.queue_name = ? THEN ?", queue_name, index])
+        sanitize_sql_array(["WHEN #{qualified_queue_name} = ? THEN ?", queue_name, index])
       end
       order(Arel.sql("(CASE #{clauses.join(' ')} ELSE #{queues.size} END)"))
     end)
--- a/app/models/good_job/job/lockable.rb
+++ b/app/models/good_job/job/lockable.rb
@@ -59,7 +59,7 @@ module GoodJob
           record = unscoped.find_by_sql([sql, locked_by_id, locked_at, lock_types[lock_type.to_s]]).first
 
           begin
-            yield(record)
+            unscoped { yield(record) }
           ensure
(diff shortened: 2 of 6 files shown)


bensheldon:
Thanks! This looks great!

bensheldon:
The JRuby failures is a flake, but the linter failure is sorbet saying the new test constants need to be added to todo.rbi

createdbypete:
Cleaned up the commits but brakeman is warning of possible SQL injection risk so I'll see about resolving that now

PR #1769, opened 2026-05-07
Fix `PG::ProgramLimitExceeded` in jobs index search for large error payloads

createdbypete:
Filtering the jobs index with `?query=...` can raise `PG::ProgramLimitExceeded: string is too long for tsvector` whenever any row has a column whose tokenised form exceeds PostgreSQL's ~1 MB tsvector limit. `Job.format_error` stores `"#{exception.class}: #{exception.message}"` without truncation, so HTTP clients echoing response bodies, `ActiveRecord::StatementInvalid` echoing bulk-insert SQL, and similar exception messages can blow past the limit and produce 500s on the dashboard search page.

The fix wraps `error` and `serialized_params->>'arguments'` with `LEFT(...)` so individual oversized rows degrade gracefully instead of breaking the page. The full `error` column is still stored and rendered on the job detail page; only the search-side tokenisation is bounded.

`LEFT(..., 262_144)` (256 KB per bounded column) leaves comfortable headroom under PG's ~1 MB tsvector ceiling even when both bounded columns are at the cap. A regression spec covers an oversized `error` row and asserts both that the search query no longer raises and that a token in the truncated prefix is still findable.per

### DIFF
--- a/app/models/concerns/good_job/filterable.rb
+++ b/app/models/concerns/good_job/filterable.rb
@@ -5,6 +5,11 @@ module GoodJob
   module Filterable
     extend ActiveSupport::Concern
 
+    # Bounds each to_tsvector() input so the combined tsvector stays under
+    # PG's ~1 MB limit (PG::ProgramLimitExceeded: "string is too long for
+    # tsvector").
+    MAX_SEARCH_COLUMN_CHARS = 262_144
+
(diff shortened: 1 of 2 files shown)


bensheldon:
TIL. Thank you! 🙇🏻

PR #1770, opened 2026-05-08
Use explicit keyword arguments for concurrency controls

bdewater-thatch:
Use explicit keyword arguments for the GoodJob concurrency DSL methods so Sorbet can model the accepted options. Preserve omitted-vs-nil behavior with an internal sentinel.

Unfortunately we can't do `T.proc.bind` for procs passed as an argument, only for the distinguished `&blk` parameter ([docs](https://sorbet.org/docs/procs#casting-the-self-type-with-tbind))

### DIFF
--- a/lib/good_job/active_job_extensions/concurrency.rb
+++ b/lib/good_job/active_job_extensions/concurrency.rb
@@ -269,16 +269,48 @@ module GoodJob
       end
 
       class_methods do
-        def good_job_control_concurrency_with(config)
-          self.good_job_concurrency_config = config
+        def good_job_control_concurrency_with(
+          total_limit: NONE,
+          enqueue_limit: NONE,
+          perform_limit: NONE,
+          enqueue_throttle: NONE,
+          perform_throttle: NONE,
+          key: NONE
+        )
+          self.good_job_concurrency_config = {
+            total_limit: total_limit,
+            enqueue_limit: enqueue_limit,
+            perform_limit: perform_limit,
+            enqueue_throttle: enqueue_throttle,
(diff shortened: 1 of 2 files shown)


lib/good_job/active_job_extensions/concurrency.rb
```diff
@@ -6,6 +6,7 @@ module Concurrency
       extend ActiveSupport::Concern
 
       VALID_TYPES = [String, Symbol, Numeric, Date, Time, TrueClass, FalseClass, NilClass].freeze
+      UNSET = Object.new.freeze
```
bensheldon: We already have GoodJob::NONE for this.

rbi/good_job.rbi
```diff
@@ -0,0 +1,104 @@
+# typed: true
+
+module GoodJob
```
bensheldon: I don't want to ship an RBI. Remove this.

bensheldon (commented):
Thanks! I left some comnents. I assumed you're using an LLM so sorry for their terseness.

rbi/good_job.rbi
bdewater-thatch: Done. Curious as to why :)

[Preserved here](https://gist.github.com/bdewater-thatch/d8fe3f72c3598f8165357c81154c08f3) in case it's useful for others to adopt in their app.

bensheldon: >  Curious as to why :)

I just haven't done it at all, so mainly want to do it with strong intention rather than shipping it in a small sorta-unrelated change.

PR #1772, opened 2026-05-18
Fix incorrect ENV variable guard for GOOD_JOB_ENABLE_PAUSES

### LINKED ISSUE #1771: enable_pauses ENV var check has singular/plural mismatch
Looks like a typo on the [ENV variable name check and read](https://github.com/bensheldon/good_job/blob/6637ecc228305cc02f0cb070a0057df1b570b55c/lib/good_job/configuration.rb#L391) making it unusable:

```ruby
return ActiveModel::Type::Boolean.new.cast(env['GOOD_JOB_ENABLE_PAUSES']) unless env['GOOD_JOB_ENABLE_PAUSE'].nil?
                                                                     ^                                     ^
```

The guard checks `GOOD_JOB_ENABLE_PAUSE` (singular) but the value is read from `GOOD_JOB_ENABLE_PAUSES` (plural). Neither env var works on its own:

```bash
$ GOOD_JOB_ENABLE_PAUSES=true bin/rails runner "puts GoodJob.configuration.enable_pauses.inspect"
false

$ GOOD_JOB_ENABLE_PAUSE=true bin/rails runner "puts GoodJob.configuration.enable_pauses.inspect"
nil
```

That said, the config object works just fine as a work-around.

jqr:
Fixes #1771 by making the guard match the cast value.

It looks to me like the plural form was the intended one as the README and actual setting are plural.

AFTER:

```bash
GOOD_JOB_ENABLE_PAUSES=true bin/rails runner "puts GoodJob.configuration.enable_pauses.inspect" 
true
```

### DIFF
--- a/lib/good_job/configuration.rb
+++ b/lib/good_job/configuration.rb
@@ -388,7 +388,7 @@ module GoodJob
     def enable_pauses
       return options[:enable_pauses] unless options[:enable_pauses].nil?
       return rails_config[:enable_pauses] unless rails_config[:enable_pauses].nil?
-      return ActiveModel::Type::Boolean.new.cast(env['GOOD_JOB_ENABLE_PAUSES']) unless env['GOOD_JOB_ENABLE_PAUSE'].nil?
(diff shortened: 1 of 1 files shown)


bensheldon:
Whoops! Thanks for catching this 🫶

Issue #1775, opened 2026-05-26
feature: ability to cancel a running job


mildred:
It would be great to have a way to cancel running jobs. If the job is enqueued, it should be removed from the queue.

It is not possible to stop the job execution, but the job can itself query at regular intervals if it has been cancelled. If it is cancelled, it could stop using an exception like a CancelError. In which case, the exception will not be seen as an error but as a successful cancellation.

I'm doing this with Sidekiq (by hand) and would like to have something similar with GoodJob. I can store the cancellation status separately, but it would be great if GoodJob could store it in its own tables.

bensheldon:
Through the dashboard, jobs can be discarded while enqueued. And during running there is an option in the dashboard to "Force Discard", which doesn't interrupt/preempt the running job, but will prevent the job from running _again_. The use-case for Force Discard is a job stuck in a retry loop (#1073)

The simplest thing in your loop would be to check something like `GoodJob::Job.find_by(active_job_id: job_id).finished_at.present?` 

I could imagine turning that into something slightly more ergonomic...

...can you tell me more about your use case?

jamie:
I'm coincidentally looking for a similar feature.

Business background: We're in the B2B accounts receivable space, our customers (aka suppliers) issue invoices to their customers, who pay them through our site. Some of our customers have very loose definitions of those terms, leading to highly variable data shapes. Normally a supplier has thousands of customers, and issues a handful of invoices to each per month. But sometimes instead of one invoice with 50 line items, they'll issue 50 invoices. Or they'll put all their 3rd-party amazon sales behind one customer, who gets 50,000 invoices. The net result is that we don't exhaustively test data at scale prior to deployment, and discover degenerate application or db query behaviours at runtime.

Technical background: we're coming from DelayedJob, where one job == 1 process, which allows us very granular control over execution and misbehaving workers. Should we hit degenerate behaviour that impacts the rest of the app (typically hitting row locks, or triggering query timeouts _elsewhere_ while waiting on a busy table), our current process is for our ops folks to pull up the offending query, track it to the DJ worker process, and kill it.

[...]

PR #1778, opened 2026-05-30
Allow Continuation jobs to be paused

bdewater-thatch:
Implements the Good Job portion of https://github.com/rails/rails/pull/57472

Related to (but doesn't exactly fix) https://github.com/bensheldon/good_job/issues/1775

Couple notes:
- I wasn't expecting `CurrentThread.job` 😅 LMK if you prefer the current conditional or want to rely on it solely
- I didn't want this to hammer the database constantly for so figured the cache was a good idea

### DIFF
--- a/app/models/good_job/setting.rb
+++ b/app/models/good_job/setting.rb
@@ -101,10 +101,22 @@ module GoodJob
       setting.save!
     end
 
-    def self.paused?(queue: nil, job_class: nil, label: nil)
-      raise ArgumentError, "Must provide at most one of queue, job_class, or label" if [queue, job_class, label].many?(&:present?)
+    def self.paused?(active_job: nil, queue: nil, job_class: nil, label: nil)
+      raise ArgumentError, "Must provide at most one of active_job, queue, job_class, or label" if [active_job, queue, job_class, label].many?(&:present?)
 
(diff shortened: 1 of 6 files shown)


bensheldon:
> I wasn't expecting CurrentThread.job 😅 LMK if you prefer the current conditional or want to rely on it solely

If we're relying on the Active Job continuation interface, I think we should only check the passed-in job and not use `CurrentThread.job`.

I'd like to have `paused?` take an optional param of `active_job:` and embed the checking behavior in there instead of in `stopping?` e.g.

```ruby
def stopping?(active_job = nil)
  return :stopping if GoodJob.current_thread_shutting_down?
  return :stopping if GoodJob::Setting.paused?(active_job: active_job)
  return :stopping if CurrentThread.job&.finished_at # job force-discarded 
  false
end
```

I'm kinda meh on the caching. Hammering the database should be fine, or at least, not a problem to speculatively solve right now. Unless you feel strongly, could you remove that? Otherwise, I'd prefer to to make a first class GoodJob::Cache object that can be generically reused.

bdewater-thatch:
Thanks for the comments, I'll pass through the job to `paused?` and remove the caching.

PR #1779, opened 2026-06-08
Emit ActiveSupport::Notifications for concurrency-aborted enqueues

bdewater-thatch:
I came across a few hardcoded logger calls and the readme promises "GoodJob is fully instrumented with ActiveSupport::Notifications", this makes sure we do :)


bensheldon:
Nice find. Thank you!

PR #1783, opened 2026-06-25
Support Rails Edge version of Arel::Table

luizkowalski:
I use Rails from the main branch on my app and I noticed that [this commit](https://github.com/rails/rails/commit/191069ea89a53c9b4d9c498c940bf146efbd967d) has introduced a breaking change to the `Arel::Table` signature. 

I added a wrapper to handle both versions here. I tried to add a test, but could only think of mocking the `Arel::Table` method signature, which I think is kinda pointless: you create a mock, then test the mock. I guess a better test is to make sure that when running the tests with  `rails_head` gemfile, it passes

I tested locally on my project and it worked (where's my "Works On My Machine™ pin 😄)

### DIFF
--- a/app/models/concerns/good_job/advisory_lockable.rb
+++ b/app/models/concerns/good_job/advisory_lockable.rb
@@ -357,7 +357,7 @@ module GoodJob
           primary_key_for_select = primary_key.to_sym
           column_for_select = column.to_sym
 
-          cte_table = Arel::Table.new(:rows)
+          cte_table = arel_table_for(:rows)
           cte_query = original_query.except(:limit)
           cte_query = if primary_key_for_select == column_for_select
                         cte_query.select(primary_key_for_select)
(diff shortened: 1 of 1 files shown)


app/models/concerns/good_job/advisory_lockable.rb
```diff
@@ -485,14 +485,8 @@ def advisory_unlock_key!(key, function: advisory_unlockable_function, connection
...
 
-      # Arel::Table.new has different parameters depending on Rails version.
-      # This method creates a new Arel::Table instance with the correct parameters.
-      # New version expects keyword arguments for the name while the old version expects a positional argument.
       def arel_table_for(name)
-        arel_table = Arel::Table
-        parameters = arel_table.instance_method(:initialize).parameters
-
-        parameters.first.any?(:req) ? arel_table.new(name) : arel_table.new(name: name)
+        ActiveRecord.version > "8.1.3" ? Arel::Table.new(name: name) : Arel::Table.new(name)
```
bensheldon: I think we need to do something like this:

```ruby
AREL_TABLE_NEW_KWARGS = Arel::Table.instance_method(:initialize).parameters.any? { |type, name| type == :key && name == :name }

def arel_table_for(name)
  AREL_TABLE_NEW_KWARGS ? Arel::Table.new(name: name) : Arel::Table.new(name)
end
```

luizkowalski: done!

bensheldon:
Beautiful. Thank you for tracking the problem down! It's been released: https://github.com/bensheldon/good_job/releases/tag/v4.19.1

luizkowalski:
oh man, already? haha thank you! and thanks for fixing my leftover, went over my head

PR #1789, opened 2026-07-01
Preserve created_at when a job is retried

### LINKED ISSUE #1788: created_at is reset every time a job is retried
When a job is retried, GoodJob reuses the same `good_jobs` record (its `id` is the `active_job_id`), but its `created_at` is overwritten with the current time on every retry, so it ends up reflecting the most recent attempt rather than when the job was first enqueued.

`GoodJob::Job.enqueue_args` always sets `created_at: Time.current`, and the retry branch of `.enqueue` assigns those attributes onto the existing record:

```ruby
retried = current_job && current_job.active_job_id == active_job.job_id
if retried
  job = current_job
  job.assign_attributes(enqueue_args(active_job, scheduled_at: scheduled_at))
  # ...
```

So after each retry the head row's `created_at` moves forward. Once a job has retried there is no longer any column that records the original enqueue time: `scheduled_at` also advances with each backoff, and the `good_job_executions` rows are per-attempt.

### Steps to reproduce

1. Enqueue a job with a `retry_on` handler that fails and retries.
2. Compare `good_jobs.created_at` before and after a retry — it changes to the retry time.

### Expected

`created_at` stays at the time the job was first enqueued, across retries, so it can be used as a stable "originally enqueued at" timestamp.

Version: good_job 4.19.1

ollym:
Fixes #1788.

`GoodJob::Job.enqueue_args` always sets `created_at: Time.current`, and the retry path in `.enqueue` assigns those attributes onto the existing job record, so `created_at` moved forward to the latest attempt on every retry.

When reenqueuing the current job, keep the record's existing `created_at` so it continues to reflect when the job was first enqueued. `scheduled_at` and the per-attempt `good_job_executions` rows still track retry timing, so nothing that depends on retry times is affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,9 @@ Available configuration options are:
     ```
 
 - `enable_pauses` (boolean) whether job processing can be paused. Defaults to `false`. You can also set this with the environment variable `GOOD_JOB_ENABLE_PAUSES`.
+- `dequeue_query_sort` (symbol) sets the sort order used when querying for the next job(s) to execute. Defaults to `:created_at`. It can be any one of:
+    - `:created_at` sorts by priority, then job creation order. This is the default, but no longer best matches GoodJob's contemporary data model.
+    - `:scheduled_at` sorts by priority, then `scheduled_at`, then `id`, which better matches GoodJob's contemporary data model. This will become the default/only behavior in a future major release.
 
 By default, GoodJob configures the following execution modes per environment:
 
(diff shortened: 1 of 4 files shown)


bensheldon:
Thanks for finding this!

And ugh, I so regret not doing `scheduled_at`-ordering correctly during the last major release cycle. 

I can get this merged. The short answer is that by default GoodJob dequeues jobs by `created_at`, so this is a load-bearing bug.  There's a undocumented (TIL) configuration option that is:

```ruby
config.good_job.dequeue_query_sort = :scheduled_at
```

But I'll update this PR merge 🫡

ollym:
@bensheldon so your recommendation is that we add `dequeue_query_sort = :scheduled_at` once this is merged?

PR #1807, opened 2026-08-19
Add spec coverage for Configuration#max_threads

olitreadwell:
## What changed

Add a `#max_threads` describe block to `spec/lib/good_job/configuration_spec.rb`.

Tests only.

No source change.

## Why

`GoodJob::Configuration#max_threads` had no direct spec.

Its precedence chain is: option, then Rails config, then `GOOD_JOB_MAX_THREADS`, then `RAILS_MAX_THREADS`, then the default of 5.

The `RAILS_MAX_THREADS` fallback had no coverage anywhere in the suite (`grep -rn RAILS_MAX_THREADS spec/` returned nothing).

That fallback is a documented Rails integration point, so a refactor could silently drop it.

The gap was verified by mutation: deleting the `env['RAILS_MAX_THREADS']` line from `max_threads` makes the new "falls back to RAILS_MAX_THREADS" example fail (expected 8, got the default 5).

So the test has teeth.

## How to verify

```
git fetch https://github.com/olitreadwell/good_job.git test-configuration-max-threads
git checkout FETCH_HEAD
bundle exec rspec spec/lib/good_job/configuration_spec.rb
bundle exec rubocop spec/lib/good_job/configuration_spec.rb
```

## Coverage

Before: `#max_threads` had 0 examples / After: 6 examples.

File suite: 48 -> 54 examples, 0 failures.

## Checks run locally

- [x] Tests: PASS (54 passed, 0 failed in configuration_spec.rb)
- [x] Lint: PASS (rubocop, no offenses)
- [ ] Typecheck: N/A (Ruby; no separate typecheck for this file)
- [ ] Build: N/A (library gem, no build step)

Env: Ruby 3.4.9, Postgres on 5432, run against the file's own suite.

The full Appraisal matrix (multiple Rails versions) was not run locally, but the fork CI runs it on this PR.

This change is spec-only, one file.

### DIFF
--- a/spec/lib/good_job/configuration_spec.rb
+++ b/spec/lib/good_job/configuration_spec.rb
@@ -57,6 +57,53 @@ RSpec.describe GoodJob::Configuration do
     end
   end
 
+  describe '#max_threads' do
+    it 'defaults to 5' do
+      stub_const 'ENV', ENV.to_hash.except('GOOD_JOB_MAX_THREADS', 'RAILS_MAX_THREADS')
+      configuration = described_class.new({})
+      expect(configuration.max_threads).to eq 5
+    end
+
+    context 'when option is given' do
+      it 'uses option value' do
+        configuration = described_class.new({ max_threads: 4 })
+        expect(configuration.max_threads).to eq 4
+      end
+    end
+
+    context 'when rails config is set' do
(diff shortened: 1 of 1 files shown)


bensheldon:
But why?

PR #1808, opened 2026-08-31
Deprecate the fallback when `probe_handler` can't be used

shouichi:
Problem:

In ruby 3.0, webrick was removed from the standard libraries. After we upgraded, `probe_handler: :webrick` stopped working. GoodJob logged a warning and used its own server instead, and we didn't notice it for months.

There is one more case. If the value is not `:webrick` or `nil`, GoodJob uses its own server without any warnings.

Solution:

Use `GoodJob.deprecator` for both cases. The behavior doesn't change, but the warning now says the fallback will raise in the next release, and an application can set `GoodJob.deprecator.behavior = :raise` to find the problem early.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1716,7 +1716,7 @@ Note that GoodJob doesn't include WEBrick as a dependency, so you'll need to add
 gem 'webrick'
 ```
 
-If WEBrick is configured to be used, but the dependency is not found, GoodJob will log a warning and fallback to the default probe server.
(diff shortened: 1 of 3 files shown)


bensheldon:
I would consider this a breaking change because it would lead to GoodJob not booting. 

This should probably be a config-driven deprecation so that I'll remember to change it to raise at the next major release.

shouichi:
Thank you. That makes sense. I switched to GoodJob.deprecator.

bensheldon:
Thank you for the slight improvement ahead of a better improvement at the next release 🙏🏻

PR #1810, opened 2026-09-02
Add a queue-scoped index for the dequeue candidate query

### LINKED ISSUE #1790: Dequeue candidate indexes don't cover queue-filtered workers — one queue's backlog degrades every other queue's claim query
## Setup

- good_job 4.19.1, PostgreSQL 18.4
- Multiple worker processes, each pinned to its own queue via `GOOD_JOB_QUEUES` (no overlap)
- Default `dequeue_query_sort` (`:created_at`)
- One queue has accumulated a backlog of ~100k jobs; the other queues are near-empty

## Problem

The claim query (hybrid lock strategy shown; the skiplocked and advisory paths build the same candidate scan) is:

```sql
WITH candidate AS MATERIALIZED (
  SELECT "good_jobs"."id"
  FROM "good_jobs"
  WHERE "good_jobs"."queue_name" = $1
    AND "good_jobs"."finished_at" IS NULL
    AND "good_jobs"."locked_by_id" IS NULL
    AND "good_jobs"."scheduled_at" <= $2
  ORDER BY priority ASC NULLS LAST, "good_jobs"."created_at" ASC
  LIMIT $3
  FOR NO KEY UPDATE SKIP LOCKED
)
UPDATE "good_jobs"
SET locked_by_id = $4, locked_at = $5, lock_type = $6
FROM (
  SELECT id FROM candidate
  WHERE pg_try_advisory_lock(($7 || substr(md5($8 || $9 || id::text), $10, $11))::bit(64)::bigint)
) AS locked
WHERE "good_jobs".id = locked.id
RETURNING "good_jobs".*
```

The expensive part is the `candidate` CTE. The shipped indexes come in two families, and neither covers it:

- **Dequeue-ordered, no `queue_name`**: `index_good_job_jobs_for_candidate_lookup` (`priority, created_at`), `index_good_jobs_for_candidate_dequeue_unlocked` (`priority, scheduled_at, id`)
- **`queue_name`-leading, not dequeue-ordered**: `index_good_jobs_on_queue_name_and_scheduled_at` and friends (`queue_name, scheduled_at`)

The planner picks a dequeue-ordered index and applies `queue_name` as a filter, so the scan walks every unfinished job from *other* queues that sorts earlier before reaching the first row of the worker's own queue. Since jobs commonly have `NULL` priority (sorted last, then `created_at ASC`), a large old backlog in any one queue sits at the front of the scan for **all** queues' workers.

Here's an actual plan captured in production for one claim by a worker on a *quiet* queue while a different queue held the ~100k backlog:

```
Limit  (cost=0.42..5.62) (actual time=29.018..29.019 rows=1)
  ->  LockRows  (cost=0.42..371388.14 rows=71459) (actual time=29.018..29.018 rows=1)
        ->  Index Scan Forward using index_good_job_jobs_for_candidate_lookup on good_jobs
              (cost=0.42..370673.55 rows=71459) (actual time=28.981..28.983 rows=2)
              Rows Removed by Filter: 101014
              Buffers: shared hit=34666
```

**~29 ms and ~34,700 buffer hits (~270 MB of page accesses) to claim a single job**, with 101,014 index entries scanned and discarded by the `queue_name` filter. Workers repeat this on every claim, so the cost scales as O(other queues' backlog) per job executed, multiplied across every worker — the busiest queue taxes the latency-sensitive ones.

## Question

Would you be open to shipping a `queue_name`-leading candidate index? Something like:

```ruby
add_index :good_jobs, [:queue_name, :priority, :created_at],
  order: { priority: "ASC NULLS LAST", created_at: :asc },
  where: "finished_at IS NULL",
  name: :index_good_jobs_dequeue_by_queue
```

(Predicate deliberately without `locked_by_id IS NULL` so it remains usable by the advisory strategy's dequeue, which doesn't filter on that column; a `scheduled_at` variant would be needed for `dequeue_query_sort = :scheduled_at`.)

This would benefit any deployment that splits workers across queues, which seems like the common shape for larger installations.

a-abdellatif98:
Fixes #1790.

Workers pinned to a queue via `GOOD_JOB_QUEUES` run a claim query filtered by `queue_name` and ordered by `priority` and `created_at`. No index leads with `queue_name`, so the planner has two bad choices: walk a priority-ordered index and filter out every other queue's rows, or use `index_good_jobs_on_queue_name` and sort the result. Which one it picks depends on table statistics, so a backlog building up on one queue degrades the claim query for every *other* queue.

### Measured

200k jobs backed up on one queue, 50 on another, claiming from the small one:

**Before**
```
Limit (actual time=19.598..19.599 rows=1)
  Buffers: shared hit=3244
  ->  Index Scan using index_good_job_jobs_for_candidate_lookup on good_jobs
        Filter: ((finished_at IS NULL) AND (locked_by_id IS NULL) AND (queue_name = 'urgent') AND (scheduled_at <= now()))
        Rows Removed by Filter: 199953
```

**After**
```
Limit (actual time=0.050..0.050 rows=1)
  Buffers: shared hit=3 read=3
  ->  Index Scan using index_good_jobs_dequeue_by_queue on good_jobs
        Index Cond: (queue_name = 'urgent')
        Filter: ((finished_at IS NULL) AND (locked_by_id IS NULL) AND (scheduled_at <= now()))
```

19.6ms to 0.05ms, 3,244 buffers to 6, and `Rows Removed by Filter` goes away. This matches what @eidam measured in the issue, where the same scan was discarding 5,982 rows.

### Two choices worth a look

**`created_at` rather than `scheduled_at`.** You noted in the issue that `(queue_name, priority, scheduled_at, id)` is what you would expect to be ideal, but that scheduled_at-ordering is not the default. `Job.dequeueing_ordered` sorts by `priority, created_at` unless `dequeue_query_sort` is `:scheduled_at`, and the default in `Configuration#dequeue_query_sort` is `:created_at`, so this index matches what most installs actually run. If `:scheduled_at` becomes the default, this index wants to change with it, and @ollym's question in the issue about whether that is planned is still open.

**Partial predicate matches index 08.** @eidam's version used `where: "finished_at IS NULL"`. I used `finished_at IS NULL AND locked_by_id IS NULL` to match `index_good_jobs_for_candidate_dequeue_unlocked`, since the claim query always includes both and the narrower predicate keeps the index smaller. Happy to widen it if you would rather it also serve queries that do not filter on `locked_by_id`.

### Changes

- New update migration template, `16_add_index_good_jobs_dequeue_by_queue.rb.erb`, following the `if_not_exists` and `algorithm: :concurrently` pattern of the existing index migrations. `UpdateGenerator` picks it up automatically from `Dir.children`.
- Matching demo migration and `demo/db/schema.rb`.

The schema line carries no `order:` key on purpose: I created the index on a real database and ran `SchemaDumper` against it, and Rails does not round-trip `ASC NULLS LAST`, exactly as with `index_good_jobs_for_candidate_dequeue_unlocked` today.

I have not touched `GoodJob.migrated?`, since that would make the new index mandatory for existing installs rather than optional.

### DIFF
--- a/demo/db/migrate/20260902000000_add_index_good_jobs_dequeue_by_queue.rb
+++ b/demo/db/migrate/20260902000000_add_index_good_jobs_dequeue_by_queue.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddIndexGoodJobsDequeueByQueue < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :good_jobs, [:queue_name, :priority, :created_at],
+      name: :index_good_jobs_dequeue_by_queue,
+      order: { priority: "ASC NULLS LAST", created_at: :asc },
+      where: "finished_at IS NULL AND locked_by_id IS NULL",
+      algorithm: :concurrently,
+      if_not_exists: true
+  end
+end
--- a/demo/db/schema.rb
+++ b/demo/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2026_07_17_000000) do
+ActiveRecord::Schema.define(version: 2026_09_02_000000) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "pg_stat_statements"
@@ -110,6 +110,7 @@ ActiveRecord::Schema.define(version: 2026_07_17_000000) do
     t.index ["priority", "scheduled_at", "id"], name: "index_good_jobs_for_candidate_dequeue_unlocked", where: "((finished_at IS NULL) AND (locked_by_id IS NULL))"
     t.index ["priority", "scheduled_at", "id"], name: "index_good_jobs_on_priority_scheduled_at_unfinished", where: "(finished_at IS NULL)"
     t.index ["priority", "scheduled_at"], name: "index_good_jobs_on_priority_scheduled_at_unfinished_unlocked", where: "((finished_at IS NULL) AND (locked_by_id IS NULL))"
+    t.index ["queue_name", "priority", "created_at"], name: "index_good_jobs_dequeue_by_queue", where: "((finished_at IS NULL) AND (locked_by_id IS NULL))"
     t.index ["queue_name"], name: "index_good_jobs_on_queue_name"
     t.index ["queue_name", "scheduled_at", "id"], name: "index_good_jobs_on_queue_name_priority_scheduled_at_unfinished", where: "(finished_at IS NULL)"
     t.index ["queue_name", "scheduled_at"], name: "index_good_jobs_on_queue_name_and_scheduled_at", where: "(finished_at IS NULL)"
--- a/lib/generators/good_job/templates/update/migrations/16_add_index_good_jobs_dequeue_by_queue.rb.erb
+++ b/lib/generators/good_job/templates/update/migrations/16_add_index_good_jobs_dequeue_by_queue.rb.erb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddIndexGoodJobsDequeueByQueue < ActiveRecord::Migration<%= migration_version %>
+  disable_ddl_transaction!
+
+  def change
+    add_index :good_jobs, [:queue_name, :priority, :created_at],
+      name: :index_good_jobs_dequeue_by_queue,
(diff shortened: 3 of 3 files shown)


a-abdellatif98:
Heads up that CI has not run on this branch: both `Test` and `CodeQL` are sitting at `action_required`, which is the first-time-contributor approval gate, so the PR shows as having no checks rather than passing ones. It needs an "Approve and run workflows" click whenever you get to it.

For what it is worth locally: the full suite passes, and the index is measured in the PR description against a 200k-row backlog, 19.6ms and 3,244 buffers down to 0.05ms and 6.

No rush, and happy to rebase if `main` has moved.

Issue #494, opened 2022-01-24
Consider new Ruby 3.0 Fiber Scheduler


ollym:
Now that https://github.com/ged/ruby-pg/pull/397 is merged, would be good to investigate adding full compatibility to Fiber.scheduler for Good Job also instead of using threads.

Obviously would require Ruby 3.0, and I don't claim to fully understand how it all works but I got a feeling Good Job performance could benefit from it for Ruby 3+

bensheldon:
@ollym Thanks for flagging this for me. GoodJob is dependent on Concurrent-Ruby for concurrency primitives, so when/if Concurrent Ruby decides to implement Fibers, GoodJob can stand on its shoulders.

PR #1811, opened 2026-09-08
Add opt-in fiber execution for I/O-bound jobs

giacope:
Hi @bensheldon, thanks for all your work on GoodJob! :)

I've been exploring fiber execution for jobs that spend most of their time waiting on I/O, and would love to get your thoughts on this approach. This follows up on #494, with inspiration from Solid Queue’s [fiber worker execution mode](https://github.com/rails/solid_queue/pull/728).

This PR adds an optional Async-backed executor that runs concurrent jobs as fibers on one reactor thread per scheduler. It can be enabled through `config.good_job.fibers`, `GOOD_JOB_FIBERS`, or `--fibers`, with existing queue counts controlling concurrency per pool. Threads remain the default, and Async is an optional dependency.

A few details:

- Fiber execution requires CRuby 3.2+, Rails 7+, fiber-scoped Rails isolation, and disabled code reloading.
- Invalid fiber configuration raises for external workers; in-process execution logs the problem and falls back to threads.
- Startup logs and the process dashboard report fiber capacity.
- The README covers setup, cooperative I/O, shutdown behavior, and database connection sizing. Advisory locks still hold a connection per running job; `:skiplocked` can allow connection sharing on Rails 7.2+, depending on what the job does.

I've added specs covering concurrency, isolation, retries, shutdown, crash recovery, and forking, along with Async compatibility CI coverage and a benchmark script for comparing thread and fiber execution across different workloads.

## Benchmark smoke results

As a quick starting point, I ran 250 jobs per scenario with three measured runs (warmups excluded). Values below are median jobs/s, with the three-run min–max in brackets. These are synthetic local results, not a production-performance claim.

| Workload (`:skiplocked`) | Threads, pool 30 | Fibers, pool 30 | Fibers, pool 5 |
|---|---:|---:|---:|
| `sleep` | 383.8 [355.9–397.5] | 448.8 (+16.9%) [446.6–455.8] | 457.9 (+19.3%) [456.6–459.6] |
| Local `Net::HTTP` | 386.1 [370.2–386.3] | 405.7 (+5.1%) [403.2–415.0] | 414.0 (+7.2%) [408.2–414.7] |
| Ruby CPU | 268.3 [267.6–271.4] | 279.3 (+4.1%) [275.2–279.4] | 276.3 (+3.0%) [273.3–280.2] |
| OpenSSL PBKDF2 | 232.1 [228.2–235.4] | 237.0 (+2.1%) [236.7–237.3] | 238.2 (+2.6%) [237.2–238.6] |

- With advisory locks on `sleep`, the medians were 344.5 jobs/s for threads/pool 30, 377.9 for fibers/pool 30, and 142.8 for fibers/pool 5—consistent with advisory locks pinning connections.
- The transaction-held-across-`sleep` case was 161.7 jobs/s for fibers/pool 5.
- Fiber workers peaked at 6 Ruby threads, compared with 30 for thread workers. Fiber workers with a pool of 5 peaked at 6 PostgreSQL connections, versus 27 with pool 30.

Environment: macOS 15.6.1 on arm64 with 8 logical CPUs; Ruby 4.0.2, Rails 8.1.3, Async 2.38.1, `pg` 1.6.3, and PostgreSQL 17.8.

I'd be very happy to rework this, narrow the scope, or fully rewrite it around a different approach if there's a better fit for GoodJob. In particular, I'd love your thoughts on the executor integration and configuration behavior.

Thanks in advance! :)


lib/good_job/cli.rb
```diff
@@ -70,6 +71,10 @@ def exit_on_failure?
                   type: :numeric,
                   banner: 'COUNT',
                   desc: "Default number of threads per pool to use for working jobs. (env var: GOOD_JOB_MAX_THREADS, default: 5)"
+    method_option :fibers,
+                  type: :string,
+                  banner: 'COUNT',
+                  desc: "Execute jobs as COUNT fibers on one reactor thread per scheduler. 'true' means 25; 'false' or 0 disables. Queue counts override COUNT, and COUNT overrides max_threads. Requires CRuby 3.2+, async >= 2.24 (>= 2.25 on Ruby 4), Rails 7.1+ with isolation_level = :fiber, and reloading disabled. (env var: GOOD_JOB_FIBERS, default: nil)"
```
bensheldon: This should simply be default: 0, otherwise any positive number enables fibers.

lib/good_job/configuration.rb
```diff
@@ -89,6 +92,11 @@ def self.total_estimated_threads(warn: false)
         GoodJob.logger.warn message
       end
 
+      if warn && scheduler_stats.any? { |stats| stats[:max_fibers] }
+        potential_checkouts = utility_threads + puma_threads + scheduler_stats.sum { |stats| stats.fetch(:max_fibers, stats[:max_threads]) }
+        GoodJob.logger.warn "GoodJob and Puma may use up to #{potential_checkouts} concurrent database checkouts, including job fibers. Pool needs depend on connection use: advisory locks, transactions, and permanent leases hold connections; jobs using :skiplocked may share them between operations."
```
bensheldon: We don't need any of these warnings anywhere.

lib/good_job/cli.rb
bensheldon: This setting is just like "default number of fibers per scheduler if not specified" right?

bensheldon (commented):
@giacope thank you for tackling this! I just wanted to give some high level commentary to see if you had any reaction to that; and after that I'm happy to get into the details.

**High Level:** This is great and I'm definitely open to including the functionality.

If I understand what you currently have:

- You make a custom Fiber-based Executor that has a compatible interface to the existing Concurrent Ruby-based thread executor.
- When passing the fiber-flag, GoodJob replaces the Thread-based executor with the Fiber-based executor. Otherwise everything else is untouched (e.g. future-scheduled jobs still go into the cache) 
- Otherwise, there's not any dependency on whether this is used with Advisory or Row/Skip locks, etc. 

Otherwise, individual comments are just things I noticed that are confusing for me. I'm not yet at the stage of deeply reading the code.

giacope:
Thanks @bensheldon! Yes, that’s the architecture: FiberPoolExecutor implements Concurrent::ExecutorService and replaces ThreadPoolExecutor in Scheduler#create_executor. The existing TimerSet cache, cache warming, cleanup, and JobPerformer paths are preserved, with small adaptations for fiber capacity and cancellation. Fiber execution works independently of the locking strategy.

Pushed a new commit where `fibers` is now a plain integer with the same shape as `max_threads`. It defaults to 0, a positive value is the default count per scheduler, and a queue count overrides it. The new fiber warnings are gone.

One question: when fiber requirements aren't met, `good_job start` raises, while in-process mode logs an error and falls back to threads. The fallback lets web and worker processes share one config even when their Rails settings differ. Keep that, or raise in both cases?

lib/good_job/cli.rb
giacope: Yes!

PR #1813, opened 2026-09-08
Add Azerbaijani locale

jamalkamaladdin:
This PR adds an Azerbaijani (az) translation.

New file `config/locales/az.yml`: 242 keys, all filled in, same key order and indentation as `config/locales/en.yml`.

`queue` is translated as `Növbə` and `batch` as `toplu iş`, the same way `tr.yml` and `ru.yml` translate them. `cron`, `GoodJob` and `ActiveJob` are left as they are.

### DIFF
--- a/config/locales/az.yml
+++ b/config/locales/az.yml
@@ -1,0 +1,321 @@
+---
+az:
+  good_job:
+    actions:
+      destroy: Sil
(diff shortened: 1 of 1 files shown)


bensheldon:
THANK YOU! 🙇🏻

PR #1817, opened 2026-09-21
Pin json < 3 as a temporary workaround

gap777:
ActiveSupport::JSON::Encoding::JSONGemEncoder#stringify calls JSON.generate(..., quirks_mode: true, max_nesting: false, script_safe: false). 

json 3.0.0 removed the quirks_mode option, so the call now raises.

Pin json < 3 until Rails fixes upstream.


gap777:
@bensheldon Here ya go!

